PR #1, opened 2016-12-08
Add repo cloning script

rlucioni:
Python 3 script for cloning Git repos housing edX services. These repos are mounted as data volumes into their corresponding Docker containers to facilitate development. Repo clone URLs can be found in settings.yml. Repos are cloned to the directory above the one housing this file.

ECOM-6562

@clintonb @jmbowman @davec-edx


clone.py
```diff
@@ -0,0 +1,77 @@
+"""
```
jmbowman, line 2: Add a shebang to the start of the script so it can be run standalone? `#!/usr/bin/env python3`

clone.py
```diff
@@ -0,0 +1,77 @@
...
+    },
+    'loggers': {
+        '': {
+            'handlers': ['console'],
+            'level': 'DEBUG',
+            'propagate': False
+        },
+    },
+})
+logger = logging.getLogger()
```
jmbowman: Any reason for not going with `getLogger(__name__)`?

settings.yml
```diff
@@ -0,0 +1,5 @@
+repos:
+  - https://github.com/edx/course-discovery.git
```
jmbowman, line 2: If I remember correctly, working with HTTPS github URLs with 2FA enabled involves making a personal access token, etc. Any chance of giving the option of doing SSH-style clones instead?

clintonb (approved):

clone.py
rlucioni: Sure, done.

clone.py
rlucioni: Nope, just forgot. Added.

settings.yml
rlucioni: HTTPS clone URLs are currently GitHub's [recommended remote URL](https://help.github.com/articles/which-remote-url-should-i-use/#cloning-with-https-urls-recommended). Cloning with these and 2FA enabled does require a personal access token, but you only need to provide it once, at which point it can be cached by your credential helper of choice (e.g., osxkeychain).

What's the benefit of using SSH URLs at this point? We can add it, but I'd rather not overcomplicate things if we don't need to.

jmbowman: Not a strong requirement, I've just always found SSH checkouts to be easier and more reliable (although that's largely due to experiences with subrepos a few years back)

jmbowman (approved):

settings.yml
rlucioni: Okay. I'll leave it out for now, but if it becomes an issue adding it will be straightforward.

PR #2, opened 2016-12-08
Added docker-compose file

clintonb:
FYI @rlucioni @davec-edx @jmbowman

### DIFF
--- a/.dev/volumes/.keep
+++ b/.dev/volumes/.keep
@@ -1,0 +1,1 @@
+This directory contains the data for devstack. The contents of this directory are ignored by Git.
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,19 @@
+.PHONY: devstack.reset devstack.start devstack.stop help
+
+help: ## Display this help message
+	@echo "Please use \`make <target>' where <target> is one of"
+	@perl -nle'print $& if m{^[\.a-zA-Z_-]+:.*?## .*$$}' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m  %-25s\033[0m %s\n", $$1, $$2}'
+
+# TODO Print out help for this target. Even better if we can iterate over the services in docker-compose.yml, and
+# print the actual service names.
+devstack.open.%:    ## Open a shell into the specified service container
+	docker exec -it edx.devstack.$* env TERM=$(TERM) /edx/app/$*/devstack.sh open
+
+devstack.reset:     ## Remove all service containers
+	docker-compose down
+
+devstack.start:     ## Start all services
+	docker-compose up
+
+devstack.stop:      ## Stop all services
+	docker-compose stop
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,95 @@
+# This file contains all of the services for an edX installation. See https://docs.docker.com/compose/compose-file/
+# for the appropriate syntax and definitions.
+#
+# Housekeeping Rules:
+# - Group third-party and edX services separately
+# - Alphabetize services in the groups
+# - Alphabetize individual configuration options for each service
+# - Every service's container name should be prefixed with "edx.devstack." to avoid conflicts with other containers
+#   that might be running for the same service.
+
+version: "2.1"
+
+services:
+  ## Third-party services
+  elasticsearch:
+    container_name: edx.devstack.elasticsearch
+    image: elasticsearch:1.5.2
+    ports:
+      - "9200:9200"
+      - "9300:9300"
+    volumes:
+      - ./.dev/volumes/elasticsearch/data:/usr/share/elasticsearch/data
+      - ./.dev/volumes/elasticsearch/logs:/usr/share/elasticsearch/logs
+      - ./.dev/volumes/elasticsearch/plugins:/usr/share/elasticsearch/plugins
+
+  memcached:
+    container_name: edx.devstack.memcached
+    image: memcached:1.4.24
+    ports:
+      - "11211:11211"
+
+  mysql:
+    command: mysqld --character-set-server=utf8 --collation-server=utf8_general_ci
+    container_name: edx.devstack.mysql
+    environment:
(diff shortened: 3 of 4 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,96 @@
...
+    container_name: edx.devstack.credentials
+    depends_on:
+      - mysql
+      - memcached
+      - discovery
+    environment:
+      ENABLE_DJANGO_TOOLBAR: 1
+    image: edxops/credentials:latest
+    ports:
+      - "18150:18150"
```
davec-edx, line 56: Credentials IDA currently has 18150 and 8150 exposed, is the current one incorrect or is 8150 missing from here?

clintonb: 8150 is the `gunicorn` port. 18150 is the `nginx` port. I can't think of any good reason to prefer gunicorn over nginx given that nginx also gives you static assets.

docker-compose.yml
```diff
@@ -0,0 +1,96 @@
...
+##
+## Housekeeping Rules:
+## - Group third-party and edX services separately
+## - Alphabetize services in the groups
+## - Alphabetize individual configuration options for each service
+## - Every service's container name should be prefixed with "edx.devstack." to avoid conflicts with other containers
+##   that might be running for the same service.
+##
+
+version: "2"
```
jmbowman: Would it be worth using [2.1](https://docs.docker.com/compose/compose-file/#version-21) instead?

Makefile
```diff
@@ -0,0 +1,13 @@
+.PHONY: devstack.reset devstack.start devstack.stop
+
+devstack.open.%:
```
jmbowman: A help target describing each option would be nice.  I like the way that was done [here](https://github.com/edx/cookiecutter-django-app/blob/master/Makefile#L5), doesn't require any changes as new targets are added

docker-compose.yml
```diff
@@ -0,0 +1,96 @@
...
+  elasticsearch:
+    container_name: edx.devstack.elasticsearch
+    image: elasticsearch:1.5.2
+    ports:
+      - "9200:9200"
+      - "9300:9300"
+    volumes:
+      - ./.dev/volumes/es/data:/usr/share/elasticsearch/data
+      - ./.dev/volumes/es/logs:/usr/share/elasticsearch/logs
+      - ./.dev/volumes/es/plugins:/usr/share/elasticsearch/plugins
```
jmbowman: Would it be worth breaking out the volume definitions for clarity?  May not help that much unless we go with the Compose format 2.1 to get [volume labels](https://docs.docker.com/compose/compose-file/#volume-configuration-reference#labels-1)

.dev/volumes/.keep
```diff
@@ -0,0 +1 @@
+This directory contains the data for devstack. The contents of this directory are ignored by Git.
```
rlucioni, line 1: Is this actually necessary? I read [Docker's reference](https://docs.docker.com/engine/tutorials/dockervolumes/#data-volumes) on data volumes, and it sounds like mounting host directories isn't necessary to persist volumes even if the container is deleted. For our purposes, mounting of host directories only seems to be necessary for mounting source code inside containers.

Makefile
rlucioni: What's up with the percentage symbol here?

Thoughts on using [invoke](http://www.pyinvoke.org/) instead of Make to avoid these questions altogether?

docker-compose.yml
```diff
@@ -0,0 +1,96 @@
+## This file contains all of the services for an edX installation. See https://docs.docker.com/compose/compose-file/
```
rlucioni: Why the double hash marks? Single should serve the same purpose.

docker-compose.yml
```diff
@@ -0,0 +1,96 @@
...
+      - mysql
+      - memcached
+      - discovery
+    environment:
+      ENABLE_DJANGO_TOOLBAR: 1
+    image: edxops/credentials:latest
+    ports:
+      - "18150:18150"
+
+  discovery:
```
rlucioni, line 58: Don't we need a host volume mounted for our services' source code? You should be able to mount
```
../<repo name>:/path/to/container/source
```

docker-compose.yml
clintonb: We aren't using any of those features; but, I see no harm in staying current.

docker-compose.yml
clintonb: We can add that later, once we have things working. I want to avoid introducing sources of complexity too early.

docker-compose.yml
clintonb: Updated.

.dev/volumes/.keep
clintonb: I have yet to get data volumes to work correctly. For now I'd like to avoid that bit of complexity (given that I can't even get time to make one image!). Once we have a working implementation we can explore areas of improvement.

rlucioni: Okay. I'm still researching this myself, but from the looks of it using the [`VOLUME`](https://docs.docker.com/engine/reference/builder/#/volume) instruction in the Dockerfiles used to build these images may actually simplify things on this end.

Makefile
clintonb: > While fully usable, Invoke is still pre-1.0 software and has no backwards compatibility guarantees until the 1.0 release occurs! Please make sure to read the changelog carefully anytime you upgrade!

Besides this, I'm not a big fan of abstracting things too much when it comes to running commands. The farther you get from bash, the more difficult it becomes to debug when things go wrong. The "magic" of edx-platform's Paver setup comes to mind.

.dev/volumes/.keep
clintonb: This has been added to my list of things to explore after we get something up and running.

docker-compose.yml
clintonb: The volume labels do not appear to correspond to directories on the host. I'll await Renzo's findings on data volumes before making this change.

PR #5, opened 2016-12-09
Added Travis support

clintonb:
The docker-compose file is now validated with very push.

ECOM-6561

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+sudo: required
+
+language: python
+
+python:
+  - "3.5"
+
+services:
+  - docker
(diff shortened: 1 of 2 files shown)


rlucioni:
It looks like the version of docker-compose running on Travis may not support docker-compose version 2.1. If https://github.com/edx/devstack/pull/6 doesn't fix it, we may need to do something like https://docs.travis-ci.com/user/docker/#Using-Docker-Compose.

clintonb:
28s. It's all uphill from here!

rlucioni:
🎉

PR #6, opened 2016-12-09
Add docker-compose to requirements

rlucioni:
@clintonb this is probably a good thing to install. :)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,2 @@
-PyYAML==3.12
+docker-compose>=1.9.0,<2.0.0
+PyYAML>=3.12,<4.0
--- a/Makefile
+++ b/Makefile
@@ -17,3 +17,6 @@ devstack.start:     ## Start all services
 
 devstack.stop:      ## Stop all services
 	docker-compose stop
+
+requirements:       ## Install requirements
+	pip install -r requirements.txt


requirements.txt
```diff
@@ -1 +1,2 @@
+docker-compose==1.9.0
```
clintonb: `docker-compose>=1.9.0,<2.0.0`

We should use requirements ranges to avoid having to update this too frequently. Same for other dependencies.

rlucioni: Allowing dependencies to change underneath me like that makes me a little uncomfortable. Isn't there value in being aware of when dependencies change, and in knowing what's going to be installed when this is run?

rlucioni: I guess you could argue that's what `pip freeze` is for?

clintonb: I don't necessarily care about specific minor/patch versions. I can only recall one time where ranges were an issue, and that was due to a bug in `astroid` (used by `pylint`). The cost of updating requirements in the future is higher to me than the uncertainty around what's installed.

We should explore https://github.com/nvie/pip-tools. It's similar to Ruby gems in the sense that you have the requirement ranges *and* a form of lock file that tells pip exactly which version to install. Best of both worlds!

rlucioni:
@clintonb updated.

requirements.txt
rlucioni: @clintonb https://github.com/pypa/pipfile might also be an option in the future.

rlucioni: Also: http://nvie.com/posts/pin-your-packages/#pin-explicitly

clintonb (approved):

Issue #10, opened 2016-12-09
Log Aggregation


clintonb:
We use Splunk at edX to aggregate logs. It would be nice to have a similar solution for devstack. Splunk may be cost-prohibitive, so consider alternative solutions.

clintonb:
There is a Splunk Docker image: https://github.com/splunk/docker-splunk. It seems to work without a license.

PR #13, opened 2016-12-09
Mount service source code into corresponding containers

rlucioni:
Also ties in the clone script: repos are cloned if they don't exist when starting and opening devstack.

ECOM-6583

@clintonb @davec-edx @jmbowman

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,25 +1,28 @@
-.PHONY: devstack.reset devstack.start devstack.stop help requirements validate
+.PHONY: clone devstack.reset devstack.start devstack.stop help requirements validate
 
 help: ## Display this help message
 	@echo "Please use \`make <target>' where <target> is one of"
 	@perl -nle'print $& if m{^[\.a-zA-Z_-]+:.*?## .*$$}' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m  %-25s\033[0m %s\n", $$1, $$2}'
 
+clone: ## Clone service repos
+	./clone.py
+
 # TODO Print out help for this target. Even better if we can iterate over the services in docker-compose.yml, and
 # print the actual service names.
-devstack.open.%:    ## Open a shell into the specified service container
+devstack.open.%: clone   ## Open a shell into the specified service container
 	docker exec -it edx.devstack.$* env TERM=$(TERM) /edx/app/$*/devstack.sh open
 
-devstack.reset:     ## Remove all service containers
+devstack.reset: ## Remove all service containers
 	docker-compose down
 
-devstack.start:     ## Start all services
+devstack.start: clone ## Start all services
 	docker-compose up
 
-devstack.stop:      ## Stop all services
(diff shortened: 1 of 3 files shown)


jmbowman (approved):

Makefile
```diff
@@ -1,25 +1,28 @@
...
 	@echo "Please use \`make <target>' where <target> is one of"
 	@perl -nle'print $& if m{^[\.a-zA-Z_-]+:.*?## .*$$}' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m  %-25s\033[0m %s\n", $$1, $$2}'
 
+clone: ## Clone service repos
+	./clone.py
+
 # TODO Print out help for this target. Even better if we can iterate over the services in docker-compose.yml, and
 # print the actual service names.
-devstack.open.%:    ## Open a shell into the specified service container
+devstack.open.%: clone   ## Open a shell into the specified service container
```
clintonb, line 12: This target assumes you have services already running. I don't think it makes sense to clone at this point.

clone.py
```diff
@@ -52,27 +53,29 @@ def __init__(self, clone_url):
...
 
+        parent_path = dirname(dirname(abspath(__file__)))
+        self.path = join(parent_path, self.name)
+
     def clone(self):
         """Clone the repo."""
-        parent_path = dirname(dirname(abspath(__file__)))
-        clone_path = join(parent_path, self.name)
-        subprocess.run(['git', 'clone', self.clone_url, clone_path], check=True)
+        subprocess.run(['git', 'clone', self.clone_url, self.path], check=True)
```
davec-edx, line 61: Is it worth sending stdout/stderr to a log file?  or opening a pipe and trap calledProcessError then log or somehow report it?  If it fails, I wouldn't want to lose the cause.

Makefile
rlucioni: @clintonb so would you only do the cloning check for `devstack.start`?

clone.py
rlucioni: @davec-edx reporting on failures is what the `check=True` is for. See https://docs.python.org/3/library/subprocess.html#subprocess.run. https://github.com/edx/devstack/issues/10 covers figuring out what to do about log aggregation.

Makefile
clintonb: Yep.

clone.py
davec-edx: Ok, I had read the docs on it, but haven't used the call, I believe it's new for 3.5.  However there is this line:
    ...and stdout and stderr if they were captured.

Makefile
rlucioni: Addressed in https://github.com/edx/devstack/pull/16.

clone.py
rlucioni: Yes, `run` is new in 3.5. If something goes wrong, we'll see the messages printed to stderr.

davec-edx: 👍

PR #14, opened 2016-12-09
Add provisioning script

clintonb:
ECOM-6565

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,29 @@
 # devstack
 
 Get up and running quickly with edX services.
+
+## Getting Started
+
+All of the services can be run by following the steps below.
+
+1. Run the provision command, if you haven't already, to configure the various services with super-users (for
+   development without the auth service) and tenants (for multi-tenancy).
+
+   The username and password for the superusers are both "edx". You can access the services directly via Django admin
+   at the `/admin` path.
+
+    ```
+    $ make devstack.provision
+    ```
+
+2. Start the services.
+
+    ```
+    $ make devstack.start
+    ```
+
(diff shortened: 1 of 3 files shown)


clintonb:
@rlucioni @davec-edx @jmbowman have a look

README.md
```diff
@@ -1,3 +1,18 @@
 # devstack
 
 Get up and running quickly with edX services.
+
+## Getting Started
+
+1. Run the provision command, if you haven't already, to configure the various services with super-users (for
+   development without the auth service) and tenants (for multi-tenancy).
+
+    ``` bash
```
rlucioni: You can remove the `bash` directive, it doesn't add anything in this case.

README.md
```diff
@@ -1,3 +1,18 @@
 # devstack
 
 Get up and running quickly with edX services.
+
+## Getting Started
+
+1. Run the provision command, if you haven't already, to configure the various services with super-users (for
+   development without the auth service) and tenants (for multi-tenancy).
```
rlucioni: Mention the username and password used for these standalone superusers. As someone new to this devstack, I might not know to dig them out of `provision.sh`.

README.md
```diff
@@ -1,3 +1,18 @@
...
+
+    ``` bash
+    $ make devstack.provision
+    ```
+
+2. Start the services.
+
+    ```bash
+    $ make devstack.start
```
rlucioni: I'd also include a note about how to use `devstack.open` here. I might not know what to do when I run this and the containers are left running in the foreground.

rlucioni (approved):
Looks good. I made a few comments about the README, but documentation changes/additions can happen separately if that's easier/faster.

PR #15, opened 2016-12-09
Pulling devstack-tagged images

clintonb:
This is temporary until the configuration is merged. See https://github.com/edx/configuration/pull/3098.

ECOM-6561


clintonb:
FYI @jmbowman @rlucioni @davec-edx these images are online now.

PR #17, opened 2016-12-09
Moved DB provisioning to the provision script

clintonb:
This alleviates issues that occur when the DB-dependent services start before the MySQL container has time to run the entrypoint script.

ECOM-6563


clintonb:
FYI @davec-edx @rlucioni @jmbowman

davec-edx (approved):

PR #20, opened 2016-12-10
Allow access to OS X hosts from inside containers

rlucioni:
This adds a script for creating a loopback alias and adding a corresponding entry to /etc/hosts on OS X machines. See the README additions for more details. The script can be run with a Make target, which in turn is run by `devstack.start` to avoid any manual steps.

ECOM-6586

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,3 +27,28 @@ After the services have started, if you need shell access to one of the services
 ```
 $ make devstack.open.discovery
 ```
+
+## Loopback Alias
+
+Containers making requests to the LMS and Studio must communicate with ports exposed on the host system by the Vagrant VM.
+
+This is fine on Linux, but doesn't work out of the box on OS X. Attempting to access localhost on a container will result
+in talking to the Docker for Mac HyperKit VM, not the host machine.
+
+While it's true that you can get this to work by accessing your Mac's external IP from your containers, this isn't ideal because
+it won't work if you have no network access on your host. Your external IP also changes as you switch networks, meaning you'd have
+to change the IP accessed by your containers every time you changed networks.
+
+A better solution, borrowed from the [Docker forums](https://forums.docker.com/t/access-host-not-vm-from-inside-container/11747/10),
+is to give your host a fixed address by creating a [loopback](http://askubuntu.com/questions/247625/what-is-the-loopback-device-and-how-do-i-use-it)
+alias. This is done for you by the `devstack.start` target.
+
+The result is a fixed IP which your containers can use to access ports on your host machine. Note that the underlying script uses `sudo`;
+adding IP addresses requires root access. Also note that the alias will not survive a host reboot, which is why the `devstack.start` target
+always attempts to set up the loopback for you.
+
+Part of the loopback alias setup includes adding a line to the `/etc/hosts` file on your machine. If you want to stop using devstack, you can clean this up by opening your `/etc/hosts` file and removing this line:
+
+```
+10.254.254.254 docker.host
+```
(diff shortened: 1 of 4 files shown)


rlucioni:
@clintonb @davec-edx @jmbowman @gsong FYI.

I still need to figure out what to do with the forwarded ports in our docker-compose file. I think our options are to remove the port forwarding altogether, or attempt to forward to different, unused ports. I don't know which is best. Thoughts?

loopback.sh
```diff
@@ -0,0 +1,15 @@
+# It's not necessary to do this on Linux, so check that we're on OS X before continuing.
+if [ "$(uname)" == "Darwin" ]; then
+    # This loopback alias allows access to OS X hosts from inside a container. 10.254.254.254
+    # is a private IP likely to be unused. As is, the alias will not survive a reboot.
+    # Borrowed from https://forums.docker.com/t/access-host-not-vm-from-inside-container/11747/10.
+    sudo ifconfig lo0 alias 10.254.254.254
+
+    # This will check for a "docker.host" entry in your hosts file. It's created if it doesn't
+    # exist, so you can use URLs like "http://docker.host:8000" in your service configuration.
```
rlucioni, line 9: We might have to think harder about this one. If we want to use `docker.host` URLs in our configuration, I'll need to extend this script so that it writes to /etc/hosts on Linux, too.

rlucioni: Addressed below.

davec-edx:
What do you mean? I'm not sure what the problem is.  Why can't the ports be
opened the way they are?

README.md
```diff
@@ -27,3 +27,22 @@ After the services have started, if you need shell access to one of the services
 ```
 $ make devstack.open.discovery
 ```
+
+## Communicating with LMS and Studio on OS X
+
+Containers making requests to the LMS and Studio must communicate with ports exposed on the host system by the Vagrant VM.
+
+This is fine on Linux, but doesn't work out of the box on OS X. Attempting to access localhost on a container will result
```
gsong: Few points:

1. Why not use the [native Mac docker-engine](https://docs.docker.com/docker-for-mac/)?
2. A container's `localhost` in `net=bridge` mode will refer to the container itself, the host will have a different IP address. This is consistent across platforms.
3. Have you seen @nedbat's [write up about hooking up course-discovery container to Vagrant](https://openedx.atlassian.net/wiki/display/~ned/Running+Course+Discovery+alongside+Devstack)?

rlucioni:
@davec-edx running our containers alongside the Vagrant VM causes port clashes. LMS and Studio forward several of the same ports (e.g., Elasticsearch on 9200).

README.md
rlucioni: 1. I am; we should all be at this point. Although it doesn't rely on VirtualBox like the Docker Toolbox used to, Docker for Mac still uses a VM: HyperKit, a lightweight macOS virtualization solution built on top of Hypervisor.framework in macOS 10.10 Yosemite and higher ([reference](https://docs.docker.com/docker-for-mac/docker-toolbox/#/the-docker-for-mac-environment)).
2. I think this is what causes problems for us in accessing the LMS and Studio running in the Vagrant VM. The containers need to be able to access ports forwarded by the Vagrant VM. in order to do that they need to talk to the host itself, not themselves.
3. Yes. The approach here relies on the same loopback alias explored there.

loopback.sh
```diff
@@ -0,0 +1,19 @@
...
+    if ! grep -q "docker.host" /etc/hosts; then
```
clintonb, line 10: Minor: Include a line in the README informing users how to _remove_ this if they want to (sadly) remove devstack.

clintonb (approved):

loopback.sh
rlucioni: Added.

PR #22, opened 2017-01-03
Make README clearer about this project's relationship with Vagrant

rlucioni:
@benpatterson @clintonb FYI.


benpatterson:
Perfect; thank you

PR #30, opened 2017-01-24
Update README to make virtualenv use more clear

rlucioni:
@clintonb @mjfrey


clintonb:
Please update to reflect the fact that Python is not needed for anything besides cloning. It would be nice if we could avoid virtualenv altogether.

rlucioni:
The need for this is obviated by https://github.com/edx/devstack/pull/42.

PR #42, opened 2017-01-25
Replace Python clone script with a Bash script

rlucioni:
Eliminates the need for a virtualenv. Because @clintonb just can't be bothered to deal with Python.


clintonb (approved):
Would you add edx-platform, please?

PR #45, opened 2017-01-26
Added Make targets to backup and restore volumes

clintonb:
This could be useful for backup/restoring volume data for numerous purposes, including:

* Provisioning
* Working on multiple machines (as I am now)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,16 @@
 .PHONY: clone devstack.provision devstack.reset devstack.start devstack.stop help requirements validate
 
+backup-volumes: ## Backup all data volumes to the local host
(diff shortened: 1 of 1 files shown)


clintonb:
@edx/ecommerce @mjfrey have a look

Makefile
```diff
@@ -1,5 +1,16 @@
 .PHONY: clone devstack.provision devstack.reset devstack.start devstack.stop help requirements validate
 
+backup-volumes:
```
rlucioni: Please add help strings for these targets.

rlucioni (approved):

PR #46, opened 2017-01-26
Updated initial DB dump for edxapp

### LINKED ISSUE #32: LMS ecommerce_worker user needs staff
LMS ecommerce_worker user needed by the trusted client should be setup as staff user

clintonb:
The ecommerce_worker user is now a staff user with @example.com email address. This brings it inline with the other service users.

Fixes #32


clintonb:
FYI @mjfrey

PR #59, opened 2017-03-06
Added Support for Docker Sync

clintonb:
Docker Sync will alleviate some of the filesystem issues on OS X.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,31 @@ For example to access the Catalog/Course Discovery Service, you can run:
 $ make devstack.open.discovery
 ```
 
+### Docker Sync
+
+Docker for Mac has known filesystem issues that significantly decrease performance. In order to mitigate these issues,
+we use [Docker Sync](https://github.com/EugenMayer/docker-sync/wiki) to synchronize file data from the host machine to
+the containers. Follow the steps below to setup Docker Sync.
+
+1. Ensure all containers are stopped.
+```
+$ make devstack.stop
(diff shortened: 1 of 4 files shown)


clintonb:
FYI @edx/ecommerce @efagin @robrap @gsong @staubina @feanil

EugenMayer:
very nice comprehensive performance test here https://openedx.atlassian.net/wiki/display/OpenDev/Docker+Sync !

your test probably do only reflect reads, which are slow, but writes are even slower, the difference should then be about 50-60 times slower then native/docker-sync driven.

did you benchmark using native_osx or unison? You should in any case upgrade to docker-sync 0.4.6, a lot of bugfixes have been made to stabalize the 0.4.x branch, which has been a hard ride @clintonb

EugenMayer:
Also since you probably think about offering windows support, docker-sync now supports windows with docker for windows + ubuntu shell WSL: could be a setup for you too https://github.com/EugenMayer/docker-sync/wiki/docker-sync-on-Windows

clintonb:
Thanks @EugenMayer, for your feedback/suggestions AND your work on Docker Sync! I plan to run additional tests sometime next week.

EugenMayer:
@clintonb well wouldnt we OSX devs otherwise kill ourselfs with a coffein shock during the coffee breaks while app stacks start, execute, run tests, compile assets or just process a request.

thanks for the heads up!

PR #60, opened 2017-03-08
Upgraded to Mongo 3.0

### LINKED ISSUE #52: Update to Mongo 3.0
edX uses Mongo 3.0 in production. Devstack should reflect this fact. Use the [existing configuration](https://github.com/edx/configuration/tree/master/playbooks/roles/mongo_3_0) to pull, and configure, the container correctly. Specifically, note that we use the [WiredTiger storage engine](https://docs.mongodb.com/manual/core/wiredtiger/).

clintonb:
Fixes #52

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,8 +30,12 @@ services:
     #   - "11211:11211"
 
   mongo:
+    # We use WiredTiger in all environments. In development environments we use small files
+    # to conserve disk space, and disable the journal for a minor performance gain.
+    # See https://docs.mongodb.com/v3.0/reference/program/mongod/#options for complete details.
+    command: mongod --smallfiles --nojournal --storageEngine wiredTiger
     container_name: edx.devstack.mongo
-    image: mongo:2.6.5
+    image: mongo:3.0.14
     # ports:
     #  - "27017:27017"
     volumes:


clintonb:
**Release Notes**
While there are methods of dumping data from Mongo 2.6 and importing it into Mongo 3.0, our limited use of this version of devstack makes the conversion unnecessarily complicated. The steps below will permanently remove the Mongo data volume, create a new one, and re-provision Mongo.

```bash
# Destroy all containers
make devstack.reset

# Remove the Mongo data volume
docker volume rm devstack_mongo_data

# Re-create the demo course
make devstack.provision
```

clintonb:
@edx/devops please review

FYI @edx/ecommerce

docker-compose.yml
```diff
@@ -30,8 +30,12 @@ services:
     #   - "11211:11211"
 
   mongo:
+    # We use WiredTiger in all environments. In development environments we use small files
+    # to conserve disk space, and disable the journal for a minor performance gain.
+    # See https://docs.mongodb.com/v3.0/reference/program/mongod/#options for complete details.
+    command: mongod --smallfiles --nojournal --storageEngine wiredTiger
     container_name: edx.devstack.mongo
-    image: mongo:2.6.5
+    image: mongo:3.0
```
rlucioni: @clintonb is this consistent with the version we run in production?

clintonb: Yes. See #52.

rlucioni (approved):

docker-compose.yml
jibsheet: Good news is that when we get to 3.2/3.4 you won't have to specify wiredTiger

jibsheet (approved):
Looks good.  We're already using mongo:3.0.14 in the forums travis tests (My only question is whether you should pin to the specific subversion we use, but not a blocker, especially since 3.0 is mostly in maintenance now).

clintonb:
Pinned to `mongo:3.0.14` to be as specific as possible regarding the MongoDB version.

PR #69, opened 2017-03-23
Updated provisioning for the Catalog Service

clintonb:
Provisioning the service will now refresh the metdata data from the dependent services and update the search index.


clintonb:
FYI @brittneyexline

PR #74, opened 2017-03-28
Exit provisioning scripts on any error

jbarciauskas:
I haven't played around with timeouts, but I was definitely missing tables after student_.. in alphabetical order, so I assume there was a timeout that the shell scripts plowed through. This should hopefully make the scripts exit on any error rather than pushing ahead.

### DIFF
--- a/load-db.sh
+++ b/load-db.sh
@@ -7,6 +7,9 @@
 #
 # This will load the edxapp database from a file named exapp.sql.
 
+set -e
+set -o pipefail
+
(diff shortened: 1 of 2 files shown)


clintonb:
A bit of knowledge for those of us less-familiar with Bash: https://coderwall.com/p/fkfaqq/safer-bash-scripts-with-set-euxo-pipefail.

clintonb (approved):

Issue #64, opened 2017-03-16
Static Assets + Webpack


clintonb:
Working with nginx poses problems for dealing with static files compiled via webpack since the webpack watcher needs to run, and assets collected in `/edx/var/<app>/staticfiles`. I found it simple to modify the command in `docker-compose.yml` to bypass nginx, and run `make static.watch` in a separate shell.

`bash -c  'source /edx/app/credentials/credentials_env && cd /edx/app/credentials/credentials && make production-requirements && ./manage.py runserver 0.0.0.0:18150'`

This is not ideal. The images should support reloading static assets on subsequent requests by default. I believe @gsong works around this by having a separate container run the equivalent of `make static.watch`.

clintonb:
Two possible solutions:

1. Stop using nginx to serve static assets
2. Instruct Django to store static files in `/edx/var/<service/staticfiles/` rather than the application directory, negating the need to run `manage.py collectstatic`.

clintonb:
We no longer use nginx.

PR #81, opened 2017-04-03
Using Django runserver to serve applications

clintonb:
Our initial desire to make development look like production by using nginx and supervisor is conflicting with our goal of easing local development pains. Specifically, front-end development requires re-compiling/collecting static assets with every change. Using Django to serve the applications alleviates this pain and allows developers to maintain their current workflows.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,12 @@ devstack.destroy: ## Destroy all containers and volumes
 devstack.open.%: ## Open a shell into the specified service container
 	docker exec -it edx.devstack.$* env TERM=$(TERM) /edx/app/$*/devstack.sh open
 
+devstack.open.lms: ## Open a shell into the LMS container
+	docker exec -it edx.devstack.lms env TERM=$(TERM) /edx/app/edxapp/devstack.sh open
+
+devstack.open.studio: ## Open a shell into the Studio container
+	docker exec -it edx.devstack.studio env TERM=$(TERM) /edx/app/edxapp/devstack.sh open
+
 devstack.provision: ## Provision all services
 	./provision.sh
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -55,7 +55,7 @@ services:
 
   # edX services
   credentials:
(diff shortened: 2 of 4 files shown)


clintonb:
@AlasdairSwan @rlucioni @mjfrey @MatthewPiatetsky would you pull this branch and let me know if it solves your issues with static assets and uploads?

clintonb:

clintonb:
This will fix #64.

Makefile
```diff
@@ -26,6 +26,12 @@ devstack.destroy: ## Destroy all containers and volumes
 devstack.open.%: ## Open a shell into the specified service container
 	docker exec -it edx.devstack.$* env TERM=$(TERM) /edx/app/$*/devstack.sh open
 
+devstack.open.lms: ## Open a shell into the LMS container
```
rlucioni, line 29: I'd like to revisit the way these targets are named. We can have that conversation separately, I just wanted to note it.

docker-compose.yml
```diff
@@ -94,17 +94,27 @@ services:
     ports:
       - "18130:18130"
 
-  edxapp:
-    command: /edx/app/edxapp/devstack.sh start
-    container_name: edx.devstack.edxapp
+  lms:
+    command: bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && NO_PREREQ_INSTALL=1 paver lms -p 18000 -f --settings devstack_docker'
```
rlucioni: You may want to consider the less arcane `--port=` and `--fast` options.

rlucioni (approved):

Makefile
clintonb: That's fine. Would you file an issue so that we can have discussion there?

clintonb:
I have verified that modifying JS files locally will reflect in the data served by Django without any input other than reloading the page.

Makefile
rlucioni: https://github.com/edx/devstack/issues/83

PR #82, opened 2017-04-04
Added support for the edx.org marketing site

clintonb:
Remaining Concerns

- [x] Support docker-sync
- [x] How do we make usage of this convenient for edX engineers who need it, but not inconvenient for anyone who does not?

ECOM-6626

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 .DEFAULT_GOAL := help
 
+include *.mk
+
 # Generates a help message. Borrowed from https://github.com/pydanny/cookiecutter-djangopackage.
 help: ## Display this help message
 	@echo "Please use \`make <target>' where <target> is one of"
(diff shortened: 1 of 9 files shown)


clintonb:
FYI @edx/ecommerce @edx/helio

Makefile
```diff
@@ -26,6 +26,9 @@ devstack.destroy: ## Destroy all containers and volumes
 devstack.open.%: ## Open a shell into the specified service container
 	docker exec -it edx.devstack.$* env TERM=$(TERM) /edx/app/$*/devstack.sh open
 
+devstack.open.marketing: ## Open a shell into the specified service container
```
rlucioni: This help string can be a little more helpful.

rlucioni (commented):
Looking good.

mikedikan (approved):

PR #84, opened 2017-04-04
Use Docker's prune command for devstack destruction

rlucioni:
No need to reinvent the wheel here. I'm using `--force` since the script already prompts for confirmation.

### DIFF
--- a/destroy.sh
+++ b/destroy.sh
@@ -5,7 +5,5 @@ if [[ $REPLY =~ ^[Yy]$ ]]
 then
     echo
     docker-compose down
-    docker volume rm devstack_mysql_data
-    docker volume rm devstack_mongo_data
-    docker volume rm devstack_elasticsearch_data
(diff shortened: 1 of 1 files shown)


clintonb (changes requested):
We should only prune the things we know about for devstack. Pruning the entire system destroys things that aren't devstack. That's not good.

rlucioni:
That's fair. To be clear, the prune command only removes unused data, so anything removed here would have been unused anyway. The command supports [filtering](https://docs.docker.com/engine/reference/commandline/system_prune/#filtering), but only by creation timestamp for now.

clintonb:
Note that "unused" does not mean unwanted. For example, if I run that command with all of my containers stopped, it will get rid of all the containers and volumes. Only running containers/volumes would remain.

rlucioni:
Yeah, good point.

PR #85, opened 2017-04-04
Clean up Make targets

### LINKED ISSUE #83: Make target improvements
Some of the existing target names are hard to work with and/or redundant. I think we can do better.

To be more precise:
1. The `devstack.` prefix. What does this add? Can we get rid of it?
2. Targets that rename/mask/shadow perfectly good Docker or Docker Compose names. For example, using `start` and `reset` instead of `up` and `down`, or (less painful) `update-images` instead of `pull`. Can we stick to the names used by Docker and Docker Compose where possible? We have the help target to explain where necessary.
3. Targets for running shells. The names for these can be more explicit. For example, `make <play>-shell`. "Open" means different things to different people (e.g., open a browser).

There may be other improvements, but this seems like a good place to start.

@clintonb FYI.

rlucioni:
Resolves #83.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,57 +1,57 @@
-.PHONY: clone devstack.provision devstack.reset devstack.start devstack.stop help requirements update-images validate
+.DEFAULT_GOAL := help
 
-backup-volumes: ## Backup all data volumes to the local host
-	docker run --rm --volumes-from edx.devstack.mysql -v $$(pwd)/.dev/backups:/backup debian:jessie tar zcvf /backup/mysql.tar.gz /var/lib/mysql
-	docker run --rm --volumes-from edx.devstack.mongo -v $$(pwd)/.dev/backups:/backup debian:jessie tar zcvf /backup/mongo.tar.gz /data/db
-	docker run --rm --volumes-from edx.devstack.elasticsearch -v $$(pwd)/.dev/backups:/backup debian:jessie tar zcvf /backup/elasticsearch.tar.gz /usr/share/elasticsearch/data
-
-restore-volumes:  ## WARNING: THIS WILL OVERWRITE ALL EXISTING DATA!!! Restore all data volumes from the local host.
-	docker run --rm --volumes-from edx.devstack.mysql -v $$(pwd)/.dev/backups:/backup debian:jessie tar zxvf /backup/mysql.tar.gz
-	docker run --rm --volumes-from edx.devstack.mongo -v $$(pwd)/.dev/backups:/backup debian:jessie tar zxvf /backup/mongo.tar.gz
-	docker run --rm --volumes-from edx.devstack.elasticsearch -v $$(pwd)/.dev/backups:/backup debian:jessie tar zxvf /backup/elasticsearch.tar.gz
-
-
+# Generates a help message. Borrowed from https://github.com/pydanny/cookiecutter-djangopackage.
 help: ## Display this help message
 	@echo "Please use \`make <target>' where <target> is one of"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,57 +1,59 @@
...
 
-devstack.reset: ## Remove all service containers
-	docker-compose down
-
-devstack.start: ## Start all services (with host volumes)
+up: ## Bring up all services with host volumes
 	docker-compose -f docker-compose.yml -f docker-compose-host.yml up
 
-devstack.start.sync: ## Start all services (with docker-sync)
+sync: ## Start all services with docker-sync
```
clintonb: `up.sync`? `sync` alone implies some sort of synchronization action is taking place. While this is technically true, it's not quite what I would expect.

Makefile
```diff
@@ -1,57 +1,59 @@
-.PHONY: clone devstack.provision devstack.reset devstack.start devstack.stop help requirements update-images validate
```
clintonb, line 1: Is this no longer necessary?

gsong: It's only needed if you have a conflicting filename.

Makefile
```diff
@@ -1,57 +1,59 @@
...
-	docker run --rm --volumes-from edx.devstack.mongo -v $$(pwd)/.dev/backups:/backup debian:jessie tar zcvf /backup/mongo.tar.gz /data/db
-	docker run --rm --volumes-from edx.devstack.elasticsearch -v $$(pwd)/.dev/backups:/backup debian:jessie tar zcvf /backup/elasticsearch.tar.gz /usr/share/elasticsearch/data
-
-restore-volumes:  ## WARNING: THIS WILL OVERWRITE ALL EXISTING DATA!!! Restore all data volumes from the local host.
-	docker run --rm --volumes-from edx.devstack.mysql -v $$(pwd)/.dev/backups:/backup debian:jessie tar zxvf /backup/mysql.tar.gz
-	docker run --rm --volumes-from edx.devstack.mongo -v $$(pwd)/.dev/backups:/backup debian:jessie tar zxvf /backup/mongo.tar.gz
-	docker run --rm --volumes-from edx.devstack.elasticsearch -v $$(pwd)/.dev/backups:/backup debian:jessie tar zxvf /backup/elasticsearch.tar.gz
+.DEFAULT_GOAL := help
 
+.PHONY: help
```
clintonb: If I understand @gsong's comment correctly, this is not needed.

rlucioni: Removed.

Makefile
rlucioni: Updated to `up-sync`. Hyphenated for consistency with other targets.

Makefile
```diff
@@ -1,57 +1,57 @@
...
 
-devstack.reset: ## Remove all service containers
-	docker-compose down
-
-devstack.start: ## Start all services (with host volumes)
+up: ## Bring up all services with host volumes
 	docker-compose -f docker-compose.yml -f docker-compose-host.yml up
 
-devstack.start.sync: ## Start all services (with docker-sync)
+up-sync: ## Start all services with docker-sync
```
clintonb: Nit: "Start" or "Bring up". I have no preference as long as we are consistent.

clintonb (approved):

Makefile
rlucioni: @clintonb thanks, updated.

clintonb:
@rlucioni please update the README.

rlucioni:
@clintonb good catch, PR incoming.

PR #92, opened 2017-04-10
Add curl health checks for ecomm, studio and LMS to Travis

jbarciauskas:
Create 'make healthchecks' to run healthchecks against active services. Still need to add credentials (not currently running on my machine for some reason) and discovery (once it's been fixed).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,3 +55,6 @@ lms-shell: ## Run a shell on the LMS container
 
 studio-shell: ## Run a shell on the Studio container
 	docker exec -it edx.devstack.studio env TERM=$(TERM) /edx/app/edxapp/devstack.sh open
+
+healthchecks: ## Run a curl against all services' healthcheck endpoints to make sure they are up. This will eventually be parameterized
+	./healthchecks.sh
--- a/healthchecks.sh
+++ b/healthchecks.sh
@@ -1,0 +1,11 @@
+set -e
+set -x
+
+echo "Checking LMS heartbeat:"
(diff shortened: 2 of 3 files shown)


jbarciauskas:
fyi @edx/docker-devstack-working-group

gsong (approved):

Makefile
```diff
@@ -55,3 +55,6 @@ lms-shell: ## Run a shell on the LMS container
 
 studio-shell: ## Run a shell on the Studio container
 	docker exec -it edx.devstack.studio env TERM=$(TERM) /edx/app/edxapp/devstack.sh open
+
+healthchecks:
```
clintonb: Add help text explaining what this does.

clintonb (approved):

healthchecks.sh
```diff
@@ -0,0 +1,11 @@
+set -e
+set -x
+
+echo "Checking LMS heartbeat:"
+curl http://localhost:18000/heartbeat #LMS
+echo
+echo "Checking Studio heartbeat:"
+curl http://localhost:18010/heartbeat #LMS
```
doctoryes: Nit: Maybe use `jq` to check the heartbeat results? Or perhaps just check the HTTP response code?

jbarciauskas: This is relying on the fact that curl exits with a non-zero exit code if the response from the server is not a 200 response code. Is that good enough or do we want something more? I figured exiting with an error + visual inspection was good enough.

doctoryes: That's enough. 👍

PR #93, opened 2017-04-13
Replaced rsync with unison

clintonb:
This will allow for two-way syncing, allowing us to run commands that create files (e.g. migrations) inside the containers.

### DIFF
--- a/docker-compose-host.yml
+++ b/docker-compose-host.yml
@@ -3,31 +3,52 @@ version: "2.1"
 services:
   credentials:
     volumes:
-      - /edx/app/credentials/credentials/credentials/assets/
-      - /edx/app/credentials/credentials/credentials/static/bundles/
-      - /edx/app/credentials/credentials/node_modules/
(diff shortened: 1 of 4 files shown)


clintonb:

clintonb:
...and try it out!

gsong:
@clintonb Will this require installation of Unison on the host machine? I imagine yes?

robrap:
@gsong I think the installation of docker-sync forced you to install unison even when we were only using rsync.  Did you not have that experience?
UPDATE: When I first installed, I got:
```
Shall we continue - DID you read it? (y/N) y
       error  Could not find unison binary in path. Please install it, e.g. using "brew install unison"
```

clintonb:
@robrap is correct.

However, I really don't like the startup penalty with unison. It takes a noticeable amount of time (over 1 minute) to sync all the data *before* the containers even start.

clintonb:
This is no longer necessary given that Docker Sync has a new strategy: https://github.com/EugenMayer/docker-sync/wiki/Testin-the-new-Native-OSX-strategy

PR #94, opened 2017-04-13
Add update-lms make command

jbarciauskas:


staubina:
:+1: This looks like a helpful command.

SalahAdDin:
Yes, is a very usefull command!

clintonb:
1. What problem does this solve?
2. Is that problem specific to LMS?
3. Can we improve documentation instead of adding a Make target that obfuscates the underlying processes?

SalahAdDin:
@clintonb 

1. Don't have to reprovisioning the dev stack each time you make modifies in requirements, for example.
2. Can be applied for LMS, E-Commerce and Studio, because they are the most commonly worked applications.
3. Basicaly you have to execute these commands all time that you need if you don't make this command.

macdiesel:
This has been covered in a slightly different way via other pull requests.

PR #97, opened 2017-04-19
Pulling from the latest tag for LMS and Studio

clintonb:
The configuration changes for these images has been merged to the master
branch of edx/configuration. We can now pull from the "official" latest
tag rather than devstack.


clintonb:

PR #99, opened 2017-04-20
Set a default value for DEVSTACK_WORKSPACE, and export it to all subs…

cpennington:
…hells in Make


jbarciauskas (approved):
?= looks like exactly what we want, and tests pass. Great!

PR #100, opened 2017-04-21
Add flag to update_index to make discovery provision successfully

jbarciauskas:
Re-enable discovery as part of devstack, since a parameter was added to fix the issue that was causing it to fail.

### DIFF
--- a/healthchecks.sh
+++ b/healthchecks.sh
@@ -9,3 +9,6 @@ curl http://localhost:18010/heartbeat # Studio
 echo
 echo "Checking ecommerce health:"
 curl http://localhost:18130/health/ # Ecommerce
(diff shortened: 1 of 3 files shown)


healthchecks.sh
```diff
@@ -9,3 +9,6 @@ curl http://localhost:18010/heartbeat # Studio
 echo
 echo "Checking ecommerce health:"
 curl http://localhost:18130/health/ # Ecommerce
+echo
+echo "Checking discovery health:"
+curl http://localhost:18381/health/ # Ecommerce
```
clintonb: Discovery

clintonb (approved):

PR #102, opened 2017-04-21
Skip running update_db for edxapp, move data to edxapp.sql

jbarciauskas:
Just an idea to speed up provisioning, as update_db for the lms is almost half of the current provisioning time


jmbowman:
We've added a newer database dump in the meantime, and are getting close to a Django upgrade that should make the migrations much faster.

PR #104, opened 2017-04-24
Starting LMS and Studio with manage.py

clintonb:
Despite the environment variable overrides, Paver still seems to be
doing more than is necessary to start LMS and Studio. Skip Paver, and
run the server directly via Django's manage.py, like every other
service.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -93,7 +93,7 @@ services:
       - "18130:18130"
 
   lms:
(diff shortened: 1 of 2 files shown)


mjfrey (approved):
Yeah!  This should solve the paver CPU issue.

jbarciauskas:
Oh great. Waiting for CI to run but glad to hear this might address the CPU usage John reported

clintonb:
The build is failing because uninstall edxval causes Django to stop. We should stop uninstalling packages in this manner.

clintonb:
Related: https://github.com/edx/edx-platform/pull/14947

PR #105, opened 2017-04-25
Add command and instructions to view logs.

robrap:


Makefile
```diff
@@ -48,6 +48,9 @@ down: ## Remove all service containers and networks
 destroy: ## Remove all devstack-related containers, networks, and volumes
 	./destroy.sh
 
+logs: ## Output the Docker image logs to stdout
```
rlucioni: This should read something like: `View logs from containers running in detached mode`. There are no images here. I'd also suggest naming the target `tail`, in keeping with Unix.

README.md
```diff
@@ -94,6 +94,13 @@ Catalog/Course Discovery Service, you can run:
 make discovery-shell
 ```
 
+To see the Docker image logs, you can either use "Kitematic" (available from the
```
rlucioni: "To see logs from containers running in detached mode, ..."

clintonb (approved):
I support this change, but I lament the fact that our Makefile has grown so large. I fear that we will make another pavelib (as is the case with edx-platform) full of indecipherable incantations that prevent developers from learning what they are actually doing.

Makefile
robrap: @rlucioni Thanks for the corrections.
In terms of `tail` vs `logs`, I don't love that particular recommendation.  If I were hunting for a command for this, the term `log` would stand out much more to me.  Also, most of our commands we switched to match the docker-compose command, which in this case is `logs`.  I recommend I leave this, or compromise with `tail-logs`, but my preference would be to leave it.  Thoughts?

robrap:
@clintonb I think we need to document common tasks either way.  I'd be happy to have us shrink the Makefile and simply have a section in the docs that provides the docker-compose commands. 
 But we should probably agree on a practice for this, document it, and try to stick with it.  Do you want to discuss this in the channel?

Makefile
rlucioni: @robrap `logs` is fine, I'm also in favor of staying consistent with the names of the Docker Compose commands when we can.

rlucioni (approved):

PR #107, opened 2017-04-28
Disabled removal of Python libs on LMS and Studio

clintonb:
This environment variable instructs edxapp to NOT remove old Python libs when install prerequisites since our images/containers will never have the older versions.


cpennington (approved):
LGTM (I haven't tested it)

PR #108, opened 2017-05-02
Adding marketing site override capaibility using .env file to enable …

mikedikan:
…local debugging

@edx/docker-devstack-working-group  for review

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,5 @@
+# Change the value to the IP address of your machine/browser to enable debugging.
+XDEBUG_IP_ADDRESS=127.0.0.1
+# Provides local environment overrides if used.  See docker.settings.private.php.example for details
+DRUPAL_EXTRA_SETTINGS=/var/www/html/sites/default/docker.settings.private.php
+
--- a/docker-compose-marketing-site.yml
+++ b/docker-compose-marketing-site.yml
@@ -16,9 +16,15 @@ services:
       - DB_NAME=edxmktg
       - DB_PASSWORD=password
       - DB_USER=edxmktg001
-      - DRUPAL_EXTRA_SETTINGS=/var/www/html/sites/default/docker.settings.php
       - LMS_HOST=http://localhost:18000
       - MEMCACHE_HOST=edx.devstack.memcached
+      # Used by PhpStorm to attach the debug connection
+      - PHP_IDE_CONFIG=serverName=edx.docker
+      # For the settings below, see .env.example to provide values (default is after the ':-')
+      # Provides local environment overrides.  (PRIVATE_SETTINGS set in .env file)
+      - DRUPAL_EXTRA_SETTINGS=${DRUPAL_EXTRA_SETTINGS:-/var/www/html/sites/default/docker.settings.php}
+      # IP address of your machine to enable debugging (IP_ADDRESS set in .env file)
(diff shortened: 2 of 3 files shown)


.env.example
```diff
@@ -0,0 +1,4 @@
+# Change the value to the IP address of your machine/browser to enable debugging.
+IP_ADDRESS=127.0.0.1
+# Provides local environment overrides if used.  See docker.settings.private.php.example for details
+PRIVATE_SETTINGS=/var/www/html/sites/default/docker.settings.private.php
```
clintonb: Add a newline.

docker-compose-marketing-site.yml
```diff
@@ -16,9 +16,15 @@ services:
...
       - DB_PASSWORD=password
       - DB_USER=edxmktg001
-      - DRUPAL_EXTRA_SETTINGS=/var/www/html/sites/default/docker.settings.php
       - LMS_HOST=http://localhost:18000
       - MEMCACHE_HOST=edx.devstack.memcached
+      # Used by PhpStorm to attach the debug connection
+      - PHP_IDE_CONFIG=serverName=edx.docker
+      # For the settings below, see .env.example to provide values (default is after the ':-')
+      # Provides local environment overrides.  (PRIVATE_SETTINGS set in .env file)
+      - DRUPAL_EXTRA_SETTINGS=${PRIVATE_SETTINGS:-/var/www/html/sites/default/docker.settings.php}
```
clintonb: Why change the variable name? Keep it the same to avoid confusion.

docker-compose-marketing-site.yml
```diff
@@ -16,9 +16,15 @@ services:
...
+      # IP address of your machine to enable debugging (IP_ADDRESS set in .env file)
+      - XDEBUG_CONFIG=remote_host=${IP_ADDRESS:-127.0.0.1}
```
clintonb: XDEBUG_IP_ADDRESS

docker-compose-marketing-site.yml
mikedikan: The idea being that DRUPAL_EXTRA_SETTINGS is always set, but if needed, it can be overridden/extended with the PRIVATE_SETTINGS file referenced in a local .env file

docker-compose-marketing-site.yml
mikedikan: 👍

.env.example
mikedikan: 👍

docker-compose-marketing-site.yml
clintonb: `DRUPAL_EXTRA_SETTINGS` is not always set. We are setting it here. I'd prefer we not make new variables unnecessarily. `DRUPAL_EXTRA_SETTINGS` should default to `/var/www/html/sites/default/docker.settings.php`. If those settings need to be overridden, developers can change the value of `DRUPAL_EXTRA_SETTINGS` in the `.env` file. It is obvious that `DRUPAL_EXTRA_SETTINGS` is related to Drupal. It is not obvious what `PRIVATE_SETTINGS` is.

mikedikan:
@clintonb please take another look

clintonb (approved):

clintonb:
Remember to squash your commits.

gsong:
👍 for starting to use `.env` with `docker-compose`.

PR #111, opened 2017-05-05
Updated for compatibility with Docker Sync 0.4.2

### LINKED ISSUE #11: Use invoke for task management and command execution
[Invoke](https://github.com/pyinvoke/invoke) is a Python (2.6+ and 3.3+) task execution tool which takes inspiration from GNU Make and Ruby's Rake. It places an emphasis on minimal boilerplate and provides traditional flag-based style command-line parsing and automatic help-text generation; see the link for more.

We should use invoke instead of Make to avoid the need for arcane lines like [this](https://github.com/edx/devstack/blob/master/Makefile#L5) and unhelpful stuff like updating [this](https://github.com/edx/devstack/blob/master/Makefile#L1).

### LINKED ISSUE #110: Not compatible with latest version of docker-sync
The `dest` field has been removed as of docker-sync version 0.3+ (https://github.com/EugenMayer/docker-sync/wiki/1.2-Upgrade-Guide#dest-has-been-removed). This breaks our ability to run `make dev.sync.provision`.

clintonb:
- Replaced calls to docker-sync-daemon with docker-sync, which is now daemon by default
- Moved from rsync strategy to native_osx, which is the new default
- Using :nocopy to avoid issues when syncing to named volumes

Fixes #110

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ dev.up: ## Bring up all services with host volumes
 	docker-compose -f docker-compose.yml -f docker-compose-host.yml up -d
 
 dev.sync.daemon.start: ## Start the docker-sycn daemon
-	docker-sync-daemon start
+	docker-sync start
 
 dev.sync.provision: | dev.sync.daemon.start dev.provision ## Provision with docker-sync enabled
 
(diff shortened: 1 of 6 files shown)


clintonb:
FYI @amangano-edx

PR #112, opened 2017-05-05
Updated LMS provisioning to be idempotent

clintonb:
Any errors that occur when loading the edxapp or edxapp_csmh databases will be ignored. This will allow users to run provisioning scripts more than once.

### DIFF
--- a/provision-lms.sh
+++ b/provision-lms.sh
@@ -1,4 +1,3 @@
-set -e
 set -o pipefail
 set -x
 
@@ -6,6 +5,7 @@ set -x
 ./load-db.sh edxapp
 ./load-db.sh edxapp_csmh
 
+set -e
 # Bring LMS online
 docker-compose $DOCKER_COMPOSE_FILES up -d lms
 


clintonb:

jbarciauskas:
I'd rather not take this shortcut. I added set -e where I did specifically because the database loading process was failing silently. It turned out I had a half-loaded database and needed to reset things.

clintonb:
@jbarciauskas in that case can you prioritize fixing this or documenting a workaround? Folks no longer know what to do when they update code, and I no longer know what to tell them. The system is essentially unusable unless people completely destroy their devstacks and provision from scratch. In some cases where data setup is a tedious, manual process this is an unacceptable demand.

jbarciauskas:
I don't understand - the update process is the same as it's always been? Switch branches, then run migrations and run asset compilation right? Is there more?

clintonb:
That's easily done on one IDA. What if I want to provision all of them?  It's worth asking: what is the usecase for provisioning? Today with Vagrant it is used to update all services. Right now it's a one-time command for Docker devstack. I find that rather confusing.

robrap:
@clintonb FYI: I don't know anyone who provisions Vagrant devstack other than the initial provisioning.  I think the process we all followed in old TNL was to get a new devstack if there was a major change that was needed.

jbarciauskas:
Yeah, my understanding of the use case was just things go foobar and you want to start over because it seems easier than sorting out what went wrong. For incremental updates like branch switches, wasn't there an update-lms make target under discussion?

clintonb:
Ah! There lies the confusion. The provision target, as I originally envisioned, was meant to be run anytime after a code update. The idea was that you could update your code and run one commeand to get all of the services functions properly (e.g. dependency installation, migrations, assets).

I do not recall other discussions on provisioning.

jmbowman:
I don't think we're hitting errors in this stage anymore, people have been re-provisioning the environment successfully for a while now.

PR #113, opened 2017-05-10
Updated README

clintonb:


clintonb:

gsong (approved):

README.md
```diff
@@ -37,7 +37,8 @@ before provisioning.
 All of the services can be run by following the steps below. Note that since we
 are running many containers, you should configure Docker with a sufficient
 amount of resources. Our testing found that [configuring Docker for Mac][] with
-2 CPUs and 4GB of memory works well.
+half of system capacity works well. For example, if you have 8 cores and 16 GB of
+memory, configure Docker with 4 cores (CPUs) and 8 GB of memory.
```
jbarciauskas, line 41: I'd prefer to stick with the minimum requirements rather than say "half" -- particularly 4GB of RAM we know is a hard minimum.

jbarciauskas (changes requested):

README.md
clintonb: Developers will be at a disadvantage with 2 CPUs and 4 GB of memory. I haven't used that configuration almost since I wrote the README because it's not very usable.

jbarciauskas: OK, it's seemed mostly fine on my Macbook. We should just rephrase as minimum and highly suggested then?

nedbat:
I made #115 to match these changes in README.rst

clintonb:
See #113

PR #116, opened 2017-05-17
Add to readme: how to read logs

nedbat:


nedbat:
@clintonb @jbarciauskas is there more detail I should add?

clintonb (approved):

clintonb:
@nedbat this seems sufficient for the general case. Any additional detail would probably get into troubleshooting specific services, which seems a bit out of scope.

PR #117, opened 2017-05-18
Using a smaller image for credentials

clintonb:
Devstack has been updated to use a smaller image for the Credentials Service. This image is built without Ansible.

Related: https://github.com/edx/credentials/pull/199


clintonb:

PR #120, opened 2017-05-18
Using cached volume configuration to improve performance

clintonb:
See https://docs.docker.com/docker-for-mac/osxfs-caching/ for details.

### DIFF
--- a/docker-compose-host.yml
+++ b/docker-compose-host.yml
@@ -3,16 +3,16 @@ version: "3"
 services:
   credentials:
     volumes:
-      - ${DEVSTACK_WORKSPACE}/credentials:/edx/app/credentials/credentials
+      - ${DEVSTACK_WORKSPACE}/credentials:/edx/app/credentials/credentials:cached
   discovery:
       volumes:
-      - ${DEVSTACK_WORKSPACE}/course-discovery:/edx/app/discovery/discovery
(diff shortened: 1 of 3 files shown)


clintonb:

clintonb:
It seems the Docker version on Travis does not support the `cached` configuration, and is not dumb/smart enough to ignore it.

gsong:
@clintonb Do we still need docker-sync if we're using `cached`?

clintonb:
@gsong Maybe not. A very quick manual test provided performance that is on par with Docker Sync.  I'll run some [loadtests](https://github.com/edx/edx-load-tests/tree/master/loadtests/devstack) when I have time.

clintonb:
Load test results updated at https://openedx.atlassian.net/wiki/display/OpenDev/Load+Tests.

Performance is about on par with Docker Sync.

gsong:
@clintonb Sweet!

robrap:
@clintonb FYI: I will try to test debugging in PyCharm soon with:
1. Vagrant devstack,
2. This change with docker-sync, and 
3. This change without docker-sync.
I'll let you know what I find.

clintonb:
This change eliminates the need for Docker Sync. I haven't used Docker Sync in a few weeks. I'll share more later today.

robrap:
@clintonb Right.  I was just recommending another possible performance test to ensure that this really does eliminate the need for docker-sync, since I was running into performance problems around this earlier.

PR #123, opened 2017-05-19
Removed host volumes for static assets and node_modules

clintonb:
I suspect these volumes retain outdated data between container creation instead of pulling the latest data from the image as intended. This results in services loading without assets.

Our usage of host volumes puts us in a weird position due to the fact that we want share code from the host while excluding certain directories, and pulling that information from the image itself. Docker doesn't seem to support such a configuration. We can resolve this issue by either rebuilding images completely with every change, using a separate container to build static assets, or explicitly sharing specific code directories rather than the entire directory.

### DIFF
--- a/docker-compose-host.yml
+++ b/docker-compose-host.yml
@@ -3,31 +3,16 @@ version: "2.1"
 services:
   credentials:
     volumes:
-      - /edx/app/credentials/credentials/credentials/assets/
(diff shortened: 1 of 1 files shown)


clintonb:
@edx/docker-devstack-working-group please review.

@schenedx I suspect this is the source of your troubles with static assets.

clintonb:
Merging now since @schenedx is blocked.

PR #126, opened 2017-05-24
Add Common issues section to README.

vkaracic:
These are some that I have saved in an Everest note.


vkaracic:
Ping @clintonb

clintonb:

jmbowman (approved):
I think we can fix the code to prevent most of these problems from occurring, but documenting the workarounds is a good start.

vkaracic:
I don't think more than one reviewer is necessary for this. Merging.

PR #131, opened 2017-06-02
Remove theme argument from site creation

mjfrey:
Theme argument no longer supported in the create site management command.


mjfrey:
@clintonb

clintonb (approved):
#oops

PR #132, opened 2017-06-05
Add custom shell env variables file.

vkaracic:
This is more of a suggestion PR, so the implementation could probably be improved. For example, to do small edits I use `nano` instead of `vim` so instead of editing the default bash text editor each time I enter the shell, I thought it would be useful to have a file with the settings and other environmental variables that someone would like to use throughout the container list.

What do you think @edx/docker-devstack-working-group ?

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,4 +2,3 @@
 XDEBUG_IP_ADDRESS=127.0.0.1
 # Provides local environment overrides if used.  See docker.settings.private.php.example for details
 DRUPAL_EXTRA_SETTINGS=/var/www/html/sites/default/docker.settings.private.php
-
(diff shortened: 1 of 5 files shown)


clintonb:
What are you editing anything inside the container?

Frankly, we should not make it convenient for people to work inside the container as that is not ideal behavior. Containers are not meant to last. If you are editing code, that should be done on your host machine. If you are editing some other file (that is part of the image), that should not be done at all. Obviously, there are exceptions, but they are exceptional not the norm.

Put simply: we don't login to EC2 and edit data inside our AMIs. We shouldn't do it with our containers.

vkaracic:
Well, for one, when I troubleshoot something I sometimes go directly to the file in a 3rd party module to print out the state, since I'm having trouble working with PDB on docker. That is not something that is stored outside of the container.
But very well, I can use this commit for my personal use then.

PR #134, opened 2017-06-05
Drop the database(s) before running the sql load scripts

macdiesel:
Provisioning will fail if these databases have already been loaded or are partially loaded. Drop the table if it exists to avoid provisioning SQL load errors.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -58,8 +58,8 @@ a minimum of 2 CPUs and 4GB of memory works well.
    services with superusers (for development without the auth service) and
    tenants (for multi-tenancy).
 
-   **Note** When running the provision command databases for all systems will be
-   dropped and recreated.
+   **Note** When running the provision command databases for ecommerce and edxapp
+   will be dropped and recreated.
 
    The username and password for the superusers are both "edx". You can access
    the services directly via Django admin at the ``/admin/`` path, or login via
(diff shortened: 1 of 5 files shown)


clintonb (commented):
Update the documentation to reflect the fact that provisioning will drop these databases.

README.rst
```diff
@@ -58,6 +58,9 @@ a minimum of 2 CPUs and 4GB of memory works well.
    services with superusers (for development without the auth service) and
    tenants (for multi-tenancy).
 
+   **Note** When running the provision command databases for all systems will be
```
clintonb: This isn't quite true. Only three databases will be dropped. If we want to drop all databases, it might be simpler to do so in `provision.sql`.

ecommerce.sql
```diff
@@ -19,6 +19,8 @@
 -- Current Database: `ecommerce`
 --
 
+DROP DATABASE /*!32312 IF EXISTS*/ `ecommerce`;
```
jmbowman, line 22: Could `dump-db.sh` be updated to add these drop database commands automatically?  Otherwise I'm afraid we'll lose these with the next execution of that.

clintonb: https://dev.mysql.com/doc/refman/5.5/en/mysqldump.html#option_mysqldump_add-drop-database

macdiesel: Options, I can add --add-drop-database to the mysqldump command or we can remove the --skip-add-drop-table option. Any preference?

macdiesel: --add-drop-database may be more efficient overall.

jmbowman (approved):

PR #136, opened 2017-06-05
Update help for updating images and code.

robrap:
For those that don't read long paragraphs closely, making sure that `make pull` and `git pull` are clearly differentiated.

@edx/docker-devstack-working-group


robrap:
@adampalay: This is also something you should do before doing a fresh provisioning.

adampalay:
Thanks @robrap , turns out I was running into another error tho!

cpennington (approved):

PR #137, opened 2017-06-06
Add Makefile command to print logs of specific container.

vkaracic:
Instead of listing all the logs, or having to append the grep command each time, this command will allow outputting the logs of a specific container only. Particularly useful if you want to read the traceback of an error that has been intertwined with lines of logs from other services.

Thoughts @edx/docker-devstack-working-group ?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,9 @@ destroy: ## Remove all devstack-related containers, networks, and volumes
 logs: ## View logs from containers running in detached mode
 	docker-compose logs -f
 
+%-logs: ## View the logs of the specified service container
+	docker-compose logs -f | grep edx.devstack.$*
+
 pull: ## Update Docker images
 	docker-compose pull
 
--- a/README.rst
+++ b/README.rst
@@ -107,6 +107,13 @@ following:
 
     make logs
 
+To view the logs of a specific service container run ``make <service>-logs``.
+For example to access the logs for Ecommerce, you can run:
+
+.. code:: sh
+
+    make ecommerce-logs
+
 To reset your environment and start provisioning from scratch, you can run:
 
 .. code:: sh


Makefile
```diff
@@ -55,6 +55,9 @@ destroy: ## Remove all devstack-related containers, networks, and volumes
 logs: ## View logs from containers running in detached mode
 	docker-compose logs -f
 
+%-logs: ## View the logs of the specified service container
```
robrap, line 58: Either needs a TODO like `%-shell:` and/or some resolution on help text.

README.rst
```diff
@@ -107,6 +107,15 @@ following:
 
     make logs
 
+To view the logs of a specific service container run the following:
+
+.. code:: sh
+
+    make <target>-logs
```
robrap: Maybe use similar pattern to `shell`:
```
To view the logs of a specific service container, run ``make <service>-logs``.
For example to access the logs for Catalog/Course Discovery Service, you can run:
 
 .. code:: sh
 
     make discovery-logs
```
Note calling it "service" instead of "target" for consistency.

robrap (changes requested):
@edx/docker-devstack-working-group We still need to have a general conversation about what is important enough to add to the make file, and what becomes a separate doc of tips and tricks.  Are we ok with just adding more and more to make?

vkaracic:
Very well. I would argue that if you do decide to keep a command for the logs that you decide on this, or something like this one, I find it to be much more useful instead of scrolling through the logs of all services to find what you look for.

Makefile
vkaracic: Does the TODO on the `%-shell` mean that the help for it should print something like:
```
...
lms-shell (...)
studio-shell (...)
ecommerce-shell (...)
...
```
?

robrap:
@vkaracic I agree that this one seems close enough to the existing `logs` target.  That wasn't meant to be a blocker on this ticket.  I was just trying to kick off that conversation.

Just the minor clean-up before I'd give thumbs.

Makefile
robrap: I'm not sure.  I don't know if we should enumerate services for each of these, or have something like the following that gives a single example:
```
SERVICE-shell Run a shell on the service container (e.g. "make lms-shell" for LMS).
```

vkaracic:
@robrap comments addressed. After this PR I could work on the TODO's in the Makefile. I have a couple of ideas for more commands as well.
How many reviews would this PR need in order to merge it?

robrap (approved):
We'll ask in HipChat, but I think we should be using 1 thumb on simple PRs. 👍

robrap:
@vkaracic I say you merge when it is green.

Makefile
```diff
@@ -55,6 +55,10 @@ destroy: ## Remove all devstack-related containers, networks, and volumes
 logs: ## View logs from containers running in detached mode
 	docker-compose logs -f
 
+# TODO: Print out help for this target.
```
clintonb: Remove the TODO.

clintonb (approved):

robrap:
@clintonb I had @vkaracic add the TODO for consistency. ;)  Either way, sounds like @vkaracic is looking to solve the help issue in a follow-up PR, so that's better than a TODO. :)

Makefile
```diff
@@ -55,6 +55,9 @@ destroy: ## Remove all devstack-related containers, networks, and volumes
 logs: ## View logs from containers running in detached mode
 	docker-compose logs -f
 
+%-logs: ## View the logs of the specified service container
+	docker-compose logs -f | grep edx.devstack.$*
```
clintonb, line 59: You don't need `grep`. See https://docs.docker.com/compose/reference/logs/.

You can simply call `docker-compose logs -f $*`.

PR #139, opened 2017-06-06
Updated Docker Compose files to be compatible with version 3

clintonb:


clintonb:

PR #141, opened 2017-06-07
Cale/improve provisioning

cpennington:


provision-studio.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -e
+set -o pipefail
+set -x
+
+# Nothing special needed for studio
+docker-compose $DOCKER_COMPOSE_FILES up -d studio
```
clintonb: Add a newline.

provision.sh
```diff
@@ -38,13 +39,14 @@ echo -e "${GREEN}Creating databases and users...${NC}"
...
-./provision-lms.sh
-
-# Nothing special needed for studio
-docker-compose $DOCKER_COMPOSE_FILES up -d studio
+./provision-oauth.sh | sed "s/^/${RED}[provision-oauth.sh]${NC} /"
 
-./provision-ecommerce.sh
-./provision-discovery.sh
-./provision-credentials.sh
+cat <<CMDS | xargs --max-procs=0 --max-lines=1 -I {} bash -c {}
```
clintonb: s/--max-lines=/-L

Use -P instead of max-procs. Unfortunately, the macOS version does not support maxing out by setting the value to 0.

clintonb (approved):
Remember to squash.

PR #143, opened 2017-06-08
docker compose version

staubina:
Returns the docker compose files to 2.1.  3.0 is not currently supported by PyCharm and not needed.


clintonb (approved):

clintonb:
Nit: Update your commit message to reflect the version to which we are changing, and why.

PR #148, opened 2017-06-12
Initial support for bok-choy tests

jmbowman:
Changes needed so far for running bok-choy tests in Docker Devstack:

* Add a stock image for a standalone Selenium 2.53.1 server driving Firefox 45.0.2.  This doesn't need to change as often as edxapp, is tested and routinely updated upstream, and avoids greatly increasing the size of the edxapp containers.  There are newer images for this container which use newer versions of Selenium and Firefox, but this one best matches our current Vagrant and Jenkins configuration; updating it will require some test updates.
* Set environment variables in the lms and studio containers specifying which ports the bok-choy run should start CMS and LMS on
* Expose these ports so Firefox can reach them from the Selenium container.  They need to be different for the lms and studio containers so they don't conflict with each other.
* Set an environment variable specifying the container name, since that will be used as the hostname for browser URLs (and it's hard to discover this from inside the container otherwise)
* Sync the .git directory for the edxapp containers, since the bok-choy test setup currently performs some git operations.  We should refactor this later to avoid this sync burden, but that comes after just getting it working in the first place.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,6 +23,10 @@ services:
       - elasticsearch_data:/usr/share/elasticsearch/data
       - elasticsearch_data:/usr/share/elasticsearch/logs
 
+  firefox:
+    container_name: edx.devstack.firefox
+    image: selenium/standalone-firefox:2.53.1-beryllium
+
   memcached:
     container_name: edx.devstack.memcached
     image: memcached:1.4.24
@@ -103,10 +107,15 @@ services:
       - memcached
       - mongo
     environment:
+      BOK_CHOY_HOSTNAME: edx.devstack.lms
+      BOK_CHOY_LMS_PORT: 18003
+      BOK_CHOY_CMS_PORT: 18031
       NO_PYTHON_UNINSTALL: 1
     image: edxops/edxapp:latest
     ports:
       - "18000:18000"
+      # - "18003:18003"
+      # - "18031:18031"
 
   studio:
     command: bash -c 'source /edx/app/edxapp/edxapp_env && while true; do python /edx/app/edxapp/edx-platform/manage.py cms runserver 0.0.0.0:18010 --settings devstack_docker; sleep 2; done'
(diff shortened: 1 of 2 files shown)


jmbowman:
FYI: @edx/docker-devstack-working-group

docker-compose.yml
```diff
@@ -103,10 +107,15 @@ services:
       - memcached
       - mongo
     environment:
+      BOK_CHOY_LMS_PORT: 18003
+      BOK_CHOY_CMS_PORT: 18031
+      CONTAINER_NAME: edx.devstack.lms
```
clintonb: Can we use an environment variable name that is less implementation-specific (e.g. `BOK_CHOY_HOSTNAME`)?

docker-compose.yml
```diff
@@ -103,10 +107,15 @@ services:
...
       NO_PYTHON_UNINSTALL: 1
     image: edxops/edxapp:latest
     ports:
       - "18000:18000"
+      - "18003:18003"
```
clintonb: Do these need to be exposed to the host? If not, this doesn't need to be here. The ports are already exposed to the other containers.

docker-compose.yml
jmbowman: Potentially, although I do need something to clearly specify "am I running in a Docker container?"  Being in a container forces a lot of bok-choy-related configuration changes across (so far) 8 different settings and configuration files in edx-platform (because absolutely everything was configured to use "localhost", and now the various services are in a bunch of different containers).

docker-compose.yml
jmbowman: They don't strictly need to be, but it is useful for debugging to be able to visit the bok-choy lms and cms services directly.

docker-compose.yml
```diff
@@ -103,10 +107,15 @@ services:
       - memcached
       - mongo
     environment:
+      BOK_CHOY_HOSTNAME: edx.devstack.lms
+      BOK_CHOY_LMS_PORT: 18003
```
macdiesel, line 111: Could we use yaml anchors for these port values?

jmbowman: Currently each one is only used once (outside of commented out blocks), so I don't think it would gain us much at the moment.

jmbowman:
@clintonb @macdiesel Does this look ok to you now?

clintonb (approved):

PR #150, opened 2017-06-14
Provisioning demo data for ecommerce

clintonb:
The create_demo_data management command is now run for the E-Commerce Service. This creates a verified seat for the demo course, allowing developers to get started testing payments with minimal effort.

LEARNER-1375


clintonb:

clintonb:
FYI @edx/learner

PR #152, opened 2017-06-14
Partial revert of "Deprecate docker-sync."

ormsbee:
This partially reverts commit 9b3b738a420095482daac52e09a40429d2a9ba21.

Further profiling showed that LMS/Studio startup times about tripled when we dropped Docker Sync in favor of Docker's built in cached volume mount system (going from 7-8s to anywhere betwen 22-28s). So we're rolling back the part of that commit that removed Docker Sync and the recommendations in the README.

However, we still want to keep the minimum version at 17.05 Edge, for image building purposes. We also want to clarify that we did try using Docker's built in cached volume support, so that people don't waste time trying it again until Docker really implements "delegated" mode.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -28,14 +28,23 @@ boot2docker) are not supported.
 
 `Docker for Windows`_ may work but has not been tested and is *not supported*.
 
-Docker Sync (Deprecated)
-~~~~~~~~~~~~~~~~~~~~~~~~
+Docker Sync
+~~~~~~~~~~~
 
-Our use of Docker Sync is now deprecated since Docker for Mac 17.04 introduced
-`performance improvements for volume mounts`_.
+Docker for Mac has known filesystem issues that significantly decrease
+performance, paticularly for starting edx-platform (e.g. when you want to run a
+test). In order to mitigate these issues, we use `Docker Sync`_ to synchronize
+file data from the host machine to the containers.
 
-Docker for Mac had known filesystem issues that significantly decreased
-performance, and we were previously using `Docker Sync`_ to solve this issue.
+If you are using macOS, please follow the `Docker Sync installation
+instructions`_ before provisioning.
+
+The performance improvements provided by `cached consistency mode for volume
+mounts`_ introduced in Docker CE Edge 17.04 are still not good enough. It's
+possible that the "delegated" consistency mode will be enough to no longer need
+docker-sync, but this feature doesn't appear to have been fully implemented yet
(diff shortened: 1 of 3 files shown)


ormsbee:
I'm really unclear on how this PR causes this error:

```
ERROR: No container found for discovery_1
ERROR: No container found for discovery_1
ERROR: No container found for discovery_1
```

README.rst
```diff
@@ -28,14 +28,22 @@ boot2docker) are not supported.
...
 `Docker for Windows`_ may work but has not been tested and is *not supported*.
 
-Docker Sync (Deprecated)
-~~~~~~~~~~~~~~~~~~~~~~~~
+Docker Sync
+~~~~~~~~~~~
 
-Our use of Docker Sync is now deprecated since Docker for Mac 17.04 introduced
-`performance improvements for volume mounts`_.
+Docker for Mac has known filesystem issues that significantly decrease
```
robrap, line 34: I reordered a bit to put the details that you don't really need last.
```
Docker for Mac has known filesystem issues that significantly decrease
performance, paticularly for starting edx-platform (e.g. when you want to run a
test). In order to mitigate these issues, we use `Docker Sync`_ to synchronize file
data from the host machine to the containers.

If you are using macOS, please follow the `Docker Sync installation
instructions`_ before provisioning.

The performance improvements provided by `cached consistency mode for volume
mounts`_ introduced in Docker CE Edge 17.04 are still not good enough. It's possible that the "delegated"
consistency mode will be enough to no longer need docker-sync, but this feature doesn't appear to have been
fully implemented yet (as of Docker 17.06.0-ce-rc2, "delegated" behaves the same
as "cached").
```

robrap, line 34: Note that there is a big cost from switching between Edge and Stable.  I'm not sure which version we should go back to recommending as the minimum requirement, and possibly we should recommend a stable version?
@macdiesel ?

robrap (changes requested):
Thanks @ormsbee.  Couple of minor thoughts.

README.rst
ormsbee: It sounds like there are other reasons to leave it at Edge (with image creation). Either way, it seems like we could address that separately? So for now, we could leave the status quo recommendation of Edge, and tackle the issue of Stable vs. Edge as a separate PR if we change our minds later?

macdiesel (approved):

ormsbee:
@robrap: Made the wording change you requested.

robrap (approved):
Thanks.

PR #155, opened 2017-06-15
Restart runserver when it dies

jmbowman:
We had previously set the `NO_PYTHON_UNINSTALL` environment variable to avoid a package uninstallation triggering a runserver restart which would terminate the Docker container.  More recently, updating the requirements in edx-platform's `github.txt` requirements file has occasionally been having the same effect (I just had another PR's Travis run fail because of this).  To resolve this, I've put each of the `runserver` commands inside an infinite loop which pauses for 2 seconds and then restarts the service if it exits for any reason.  This allows us to run scripts that make invasive changes which may restart or terminate the `runserver` process, and have them run through to completion without being canceled by the container abruptly exiting.

If the runserver process is killed by an error of some sort, the exception is typically logged and can be found via `make logs`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -55,7 +55,7 @@ services:
 
   # edX services
   credentials:
-    command: bash -c 'python /edx/app/credentials/credentials/manage.py runserver 0.0.0.0:18150'
+    command: bash -c 'while true; do python /edx/app/credentials/credentials/manage.py runserver 0.0.0.0:18150; sleep 2; done'
     container_name: edx.devstack.credentials
     depends_on:
       - mysql
@@ -70,7 +70,7 @@ services:
       - "18150:18150"
 
   discovery:
-    command: bash -c 'source /edx/app/discovery/discovery_env && python /edx/app/discovery/discovery/manage.py runserver 0.0.0.0:18381'
+    command: bash -c 'source /edx/app/discovery/discovery_env && while true; do python /edx/app/discovery/discovery/manage.py runserver 0.0.0.0:18381; sleep 2; done'
     container_name: edx.devstack.discovery
     depends_on:
       - mysql
@@ -84,7 +84,7 @@ services:
       - "18381:18381"
 
   ecommerce:
-    command: bash -c 'source /edx/app/ecommerce/ecommerce_env && python /edx/app/ecommerce/ecommerce/manage.py runserver 0.0.0.0:18130'
+    command: bash -c 'source /edx/app/ecommerce/ecommerce_env && while true; do python /edx/app/ecommerce/ecommerce/manage.py runserver 0.0.0.0:18130; sleep 2; done'
     container_name: edx.devstack.ecommerce
     depends_on:
       - mysql
@@ -96,7 +96,7 @@ services:
       - "18130:18130"
 
   lms:
-    command: bash -c 'source /edx/app/edxapp/edxapp_env && python /edx/app/edxapp/edx-platform/manage.py lms runserver 0.0.0.0:18000 --settings devstack_docker'
+    command: bash -c 'source /edx/app/edxapp/edxapp_env && while true; do python /edx/app/edxapp/edx-platform/manage.py lms runserver 0.0.0.0:18000 --settings devstack_docker; sleep 2; done'
     container_name: edx.devstack.lms
     depends_on:
       - mysql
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -55,7 +55,7 @@ services:
 
   # edX services
   credentials:
-    command: bash -c 'python /edx/app/credentials/credentials/manage.py runserver 0.0.0.0:18150'
+    command: bash -c 'while true; do python /edx/app/credentials/credentials/manage.py runserver 0.0.0.0:18150; sleep 2; done'
```
clintonb, line 58: The credentials image will always have the latest dependencies because it is built with every push to edx/credentials. The proper solution for this service is to pull the latest image.

docker-compose.yml
```diff
@@ -96,27 +96,23 @@ services:
...
 
   lms:
-    command: bash -c 'source /edx/app/edxapp/edxapp_env && python /edx/app/edxapp/edx-platform/manage.py lms runserver 0.0.0.0:18000 --settings devstack_docker'
+    command: bash -c 'source /edx/app/edxapp/edxapp_env && while true; do python /edx/app/edxapp/edx-platform/manage.py lms runserver 0.0.0.0:18000 --settings devstack_docker; sleep 2; done'
     container_name: edx.devstack.lms
     depends_on:
       - mysql
       - memcached
       - mongo
-    environment:
```
clintonb: We should still keep these in place. There is no reason to uninstall any dependencies on new images.

docker-compose.yml
```diff
@@ -96,27 +96,23 @@ services:
       - "18130:18130"
 
   lms:
-    command: bash -c 'source /edx/app/edxapp/edxapp_env && python /edx/app/edxapp/edx-platform/manage.py lms runserver 0.0.0.0:18000 --settings devstack_docker'
+    command: bash -c 'source /edx/app/edxapp/edxapp_env && while true; do python /edx/app/edxapp/edx-platform/manage.py lms runserver 0.0.0.0:18000 --settings devstack_docker; sleep 2; done'
```
clintonb, line 99: Please add comments explaining that these changes should be reverted once we have automated image builds, and linking to the relevant ticket.

docker-compose.yml
jmbowman: I suspect we'll eventually move the virtualenvs out of the containers and into volumes, so this may become relevant again.  But I'm ok with leaving it set for now.

docker-compose.yml
jmbowman: Automated image builds won't fix this.  The current problem is that we have a lot of scripts which start by ensuring that all requirements are up to date, and pip always reinstalls dependencies specified as GitHub URLs (because there isn't a version number to rely on).  There's at least one of these packages that has a good chance of taking down `runserver` when pip uninstalls it in order to install it again.

Given that this is at least the second way we've found for `runserver` to get terminated while doing routine environment maintenance, I think it's a good idea to always run it more robustly like this (at least for devstack).

docker-compose.yml
jmbowman: Even if you're just locally testing if it's safe to upgrade a particular dependency to a newer version?  The image can't have the latest dependencies if you haven't committed them yet, and it seems a little extreme to require a full new image build locally for every library upgrade attempt (especially if you're iterating over changes to a required package).

docker-compose.yml
clintonb: Fine, but we should acknowledge that we are treating a symptom. The root cause is the fact that we are installing a lot of stuff from GitHub when it should, ideally, be installed from PyPI (or our own package server).

docker-compose.yml
clintonb: Rebuilding is not extreme at all. That seems to be how others work with Docker in development. We are averse to rebuilding because it takes forever (due to lack of caching for Ansible). However, the Credentials Service uses a [trimmed-down Dockerfile](https://github.com/edx/credentials/blob/master/Dockerfile). Building that image takes less than 4 minutes (most of that due to static asset compilation which is limited by a dependency on susy). It's not out of the question for a developer to build that image locally when installing new dependencies.

docker-compose.yml
clintonb: Why would we move the virtualenvs outside of the containers?

jmbowman: Not something I'm advocating immediately, but the bandwidth requirements of needing new images on every git pull are pretty high (especially for edx-platform, which comes in at around 2 GB).  Being able to pull the frequently-updated directories out into persistent volumes could allow each downloaded image to be useful for much longer, since our existing dependency update scripts could keep things current incrementally.  But that's a discussion for later.

clintonb (approved):

PR #157, opened 2017-06-20
Add documentation about running LMS tests.

vkaracic:
@edx/docker-devstack-working-group please review.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -276,6 +276,19 @@ PyCharm Integration
 
 See the `Pycharm Integration documentation`_.
 
+Running LMS and Studio Python tests
+-----------------------------------
+
+If you want to run Python tests inside the LMS or Studio container, you need to pass the
+``test_docker`` settings flag. Example:
+
+.. code:: sh
+
+    DISABLE_MIGRATIONS=1 ./manage.py lms test --settings=test_docker openedx.core.djangoapps.user_api
+
+You can also add the ``test_docker`` settings flag to the other examples detailed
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -306,6 +306,17 @@ Start over
 If you want to completely start over, run ``make destroy``. This will remove
 all containers, networks, AND data volumes.
 
+Run LMS tests
```
robrap: 1. Rename to `LMS and Studio Python tests`.
2. I would move this as a higher-level section immediately before or after PyCharm Integration, to hopefully keep people from having trouble. :)
3. Add another line at the end: `You can also add the ``test_docker`` settings flag to the other examples detailed in the edx-platform testing documentation.`, making `edx-platform testing documentation` a reference to https://github.com/edx/edx-platform/blob/master/docs/testing.rst#running-python-unit-tests

README.rst
```diff
@@ -306,6 +306,17 @@ Start over
...
+~~~~~~~~~~~~~
+
+If you want to run tests inside the LMS container, you need to pass the
```
robrap: `tests` => `python tests`

robrap (commented):
Thanks for adding this!

FYI: @macdiesel: as part of the edx-platform workflow docs work it may make sense to pull this out to another doc, but I think this is fine here until that larger work gets started.

vkaracic:
@robrap no problem! I believe I have addressed your comments. Please re-review.

robrap (approved):

PR #158, opened 2017-06-20
Added instructions for database dumps

clintonb:


clintonb:
@efischer19 give it a shot.

README.rst
```diff
@@ -252,6 +252,24 @@ For example, if you wanted to build tag ``release-2017-03-03`` for the
 E-Commerce Service, you would modify ``ECOMMERCE_VERSION`` in
 ``docker/build/ecommerce/ansible_overrides.yml``.
 
+How do I create database dumps?
+-------------------------------
+Occasionally we need to update the database dumps to speed up provisioning and generally spend less time running
+migrations. Updating these database dumps is pretty straightforward.
```
robrap: Could we also answer the question when?  Do we do this anytime we see migrations run during provisioning?  Will eat be obvious when we should dump ecommerce vs edxapp vs edxapp_csmh?

robrap (approved):
Thanks for the instructions.

Am I correct in understanding that this would never be required, but might only speed things?  

It would be great to have more clarity about when I'd do this and how I know which scripts to run.

README.rst
clintonb: The database dumps only exist because running migrations from scratch takes forever, and we have never squashed migrations. When should we make new dumps? Whenever migrations start to seem slow or, as is the case now, we want to incorporate changes that require manual intervention.

Technically, if CI is running on a daily basis, it could be updated to run migrations from scratch and update the dumps.

robrap: Would you mind adding some form of your answer into the docs?  I think it would be helpful beyond the life of this PR.

PR #159, opened 2017-06-20
Update with freshly-generated edxapp.sql

jbarciauskas:
@edx/docker-devstack-working-group


clintonb:
You will also want to include edxapp_csmh.sql. The two databases are related.

clintonb (approved):
Squash to a single commit.

jbarciauskas:
Obsoleted by #160

PR #161, opened 2017-06-23
Fix disk space issues and print logs from container after failed healthcheck

jbarciauskas:
This should fix disk space issues as well as provide more debugging output when things go south in the future.

### DIFF
--- a/healthchecks.sh
+++ b/healthchecks.sh
@@ -1,14 +1,29 @@
-set -e
 set -x
 
 echo "Checking LMS heartbeat:"
-curl http://localhost:18000/heartbeat # LMS
+curl http://localhost:18000/heartbeat
+if [[ $? -ne 0 ]]; then
+    docker-compose logs
+    exit 2
+fi
 echo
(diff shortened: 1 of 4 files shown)


jbarciauskas:
@edx/docker-devstack-working-group I think this is ready to go

healthchecks.sh
```diff
@@ -1,14 +1,29 @@
-set -e
 set -x
 
 echo "Checking LMS heartbeat:"
-curl http://localhost:18000/heartbeat # LMS
+curl http://localhost:18000/heartbeat
+if [[ $? -ne 0 ]]; then
```
robrap, line 5: I was going to recommend you use a trap for error handling, but there is a lot of confusing articles about the proper way to handle signals.  If you want to look into it, it is probably a cleaner way to create a common error handler.

robrap (approved):

healthchecks.sh
jbarciauskas: I'm going to merge this then look into it

PR #162, opened 2017-06-26
Fix exposed studio port.

robrap:


clintonb (approved):

clintonb:
I misread the switch. Why the change to from 18010? If you make this change here you will need to update the images and configuration. See https://github.com/edx/configuration/blob/master/docker/build/edxapp/Dockerfile#L34.

PR #163, opened 2017-06-26
Adjust Studio to port 18001.

robrap:


clintonb (commented):
Why?

clintonb:
If you make this change here you will need to update the images and configuration. See https://github.com/edx/configuration/blob/master/docker/build/edxapp/Dockerfile#L34.

robrap:
Decided that the original port choice of 18010 for Studio was related to edx/configuration of gunicorn + nginx.  We are going to go with consistency with configuration, rather than with the old Studio port of 8001.

PR #169, opened 2017-06-30
PLAT-1351 Document how to install/upgrade dependencies

jmbowman:
Adding some explanation of how to handle JS and Python package management in Docker devstack.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -252,6 +252,46 @@ require manual intervention. The process below details how to update the databas
    ./dump-db.sh edxapp
    ./dump-db.sh edxapp_csmh
 
+How do I upgrade Node.JS packages?
+----------------------------------
+
+JavaScript packages for Node.js are installed into the ``node_modules``
+directory of the local git repository checkout which is synced into the
+corresponding Docker container.  Hence these can be upgraded via any of the
+usual methods for that service (``npm install``,
+``paver install_node_prereqs``, etc.), and the changes will persist between
+container restarts.
+
+How do I upgrade Python packages?
+---------------------------------
+
+Unlike the ``node_modules`` directory, the ``virtualenv`` used to run Python
+code in a Docker container only exists inside that container.  Changes made to
+a container's filesystem are not saved when the container exits, so if you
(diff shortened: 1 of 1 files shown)


jmbowman:
@edx/docker-devstack-working-group This is ready for review, let me know if you think of anything that should be added or rephrased.

README.rst
```diff
@@ -252,6 +252,42 @@ require manual intervention. The process below details how to update the databas
...
+
+How do I upgrade Python packages?
+---------------------------------
+
+Unlike the ``node_modules`` directory, the ``virtualenv`` used to run Python
+code in a Docker container only exists inside that container.  Changes made to
+a container's filesystem are not saved when the container exits, so if you
+manually install or upgrade Python packages in a container (via
+``pip install``, ``paver install_python_prereqs``, etc.), they will no
+longer be present if you restart the container.  If you want to ensure that
```
robrap: Is stop vs down something people need to be aware of to understand what is happening when?  Should we include that here, or a separate section that we point to?

README.rst
```diff
@@ -252,6 +252,42 @@ require manual intervention. The process below details how to update the databas
...
+your new or upgraded packages are present in the container every time it
+starts, you have a few options:
+
+* Merge your updated requirements files and wait for the next official Docker
```
robrap: For someone new to this, I don't think they would know what an "official Docker image" is.  Maybe you should mention Jenkins or point to a section that discusses this? To a newbie (like me), this sounds like I'm waiting on something from Docker.

robrap (commented):
I think it would be fine for you to merge and iterate, or improve and then merge.

README.rst
gsong: @robrap Do you mean the fact that `stop` preserves state and `down` is just `stop` followed by `rm` and destroys state?

robrap: @gsong I mean that this PR uses the language "not saved when the container exits" and "if you restart the container".  Is this referring to `down` with destroyed state?  Is there a way to make this more clear here and/or to point to another place that describes this generally?

PR #171, opened 2017-07-01
Use `[[` the new and improved version of test

### LINKED ISSUE #170: clone.sh: line 36: [: : integer expression expected
I just got back to looking at Open edx again.

I am using docker `Version 17.06.0-ce-mac18 (18433)
Channel: stable
d9b66511e0`

When I run `make dev.clone` I am seeing this 
```
$ make dev.clone
./clone.sh
./clone.sh: line 36: [: : integer expression expected
```

azizur:
Fixes #170.

[What is the difference between test, `[` and `[[`?](http://mywiki.wooledge.org/BashFAQ/031)

### DIFF
--- a/clone.sh
+++ b/clone.sh
@@ -4,10 +4,10 @@
 # data volumes into their corresponding Docker containers to facilitate development.
 # Repos are cloned to the directory above the one housing this file.
 
-if [ -z "$DEVSTACK_WORKSPACE" ]; then
+if [[ -z "$DEVSTACK_WORKSPACE" ]]; then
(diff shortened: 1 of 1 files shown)


macdiesel:
My concern here is that you've only updated this set of tests and the tests in other files remain using [ ]. Also we may lose some portability, while I'm not that concerned about bourne shell or Korn, I would like the styling of the tests to be consistent across all files.  

I've opened a PR #176 that I think may solve your issue, could you give it a try.

macdiesel:
picked #184 over this.

PR #173, opened 2017-07-05
Add script to generate test courses in Studio

stevenzheng3791:
LEARNER-1499
https://openedx.atlassian.net/browse/LEARNER-1499
Command outside of edx-platform to create test courses in studio by calling the appropriate management command in studio. See https://github.com/edx/edx-platform/pull/15494

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,3 +98,6 @@ validate-lms-volume: ## Validate that changes to the local workspace are reflect
 
 mysql-shell: ## Run a shell on the mysql container
 	docker-compose exec mysql bash
+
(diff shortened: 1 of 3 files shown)


macdiesel:
@stevenzheng3791 Could you make this a make target instead of a python script? Also seems unnecessary to store the format in yaml then convert to json.

stevenzheng3791:
@macdiesel I've restructured things a bit. So there is a make target now that calls a very simple script that loads a json file and passes that json object into the command

macdiesel (approved):
Assuming you've tested this and it works merge away. Thank you for the changes.

PR #179, opened 2017-07-07
Troubleshooting section for getcwdu problem.

robrap:
FYI: @andy-armstrong

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -477,22 +477,52 @@ Or, you can run the following commands to clean up dangling images and volumes:
    docker rmi $(docker images -f "dangling=true" -q)
    docker volume rm $(docker volume ls -qf dangling=true)
 
+No such file or directory
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+While provisioning, some have seen the following error:
+
+.. code:: sh
+
+   ...
+       cwd = os.getcwdu()
+   OSError: [Errno 2] No such file or directory
+   make: *** [dev.provision.run] Error 1
+
+Everyone who has seen this has gotten past it, but we don't have a surefire way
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -477,6 +477,31 @@ Or, you can run the following commands to clean up dangling images and volumes:
...
+Everyone who has seen this has gotten past it, but we don't have a surefire way
+to do so. Rebooting and restarting Docker do *not* seem to correct the issue. It
+may be an issue that is exacerbated by our use of sync (which typically speeds
+up the provisioning process on Mac), so you can try the following:
+
+.. code:: sh
+
+   # repeat the following until you get past the error.
+   make stop
+   make provision
```
mjfrey: make dev.provision

README.rst
```diff
@@ -477,6 +477,31 @@ Or, you can run the following commands to clean up dangling images and volumes:
...
+
+Once you get past the issue, you should be able to continue to use sync versions
+of the make targets.
```
mjfrey: Not sure if this is true.  Once you provision without sync, you can not use sync.

mjfrey (commented):
couple of comments.

README.rst
robrap: I'm pretty sure that sync can be used or not used at any time.  It just requires that sync is first run, which is what our make targets using sync use.
@clintonb @jbarciauskas @macdiesel: Can anyone confirm?

README.rst
robrap: I'll fix.  Thanks.

README.rst
clintonb: Correct. You can swap between using Docker Sync and native volumes at anytime.

mjfrey (approved):

macdiesel (approved):

PR #182, opened 2017-07-07
Prune dangling images after provision

macdiesel:


clintonb:
Typo in the commit message. How are the dangling images created?

macdiesel:
@clintonb want to run it for me in your local env and see what happens? =)

robrap:
@macdiesel: I think this is ok.  The docs for prune are not as clear as I wish them to be, but this sounds ok to me.  Do you want to clean up dangling volumes as well?

See this part of the readme: https://github.com/edx/devstack/blame/master/README.rst#L483
I think we may want to update the README as well, because I think I was using commands from before prune existed.

macdiesel:
@robrap I've updated the docs.

Re prune volumes: I don't think we should try to manage that. The cost of deleting user data in error is too high.

robrap (approved):

README.rst
```diff
@@ -480,12 +480,12 @@ Or, you can run the following commands to clean up dangling images and volumes:
 
 .. code:: sh
 
-   docker rmi $(docker images -f "dangling=true" -q)
-   docker volume rm $(docker volume ls -qf dangling=true)
+   docker image prune -f
+   docker volume prune -f
```
clintonb: This is a dangerous command, and should have a bit more warning around it.

README.rst
```diff
@@ -480,12 +480,12 @@ Or, you can run the following commands to clean up dangling images and volumes:
...
 
 No such file or directory
 ~~~~~~~~~~~~~~~~~~~~~~~~~
-
+
```
clintonb: Nit: Tab/space is not necessary.

README.rst
robrap: If I'm reading the docs correctly, the "-f" only takes care of the dangling ones, which shouldn't be dangerous.  Let me know if you think otherwise.  I think it is not like the system prune we documented above.

clintonb: I see your point, but I disagree with the assumption that everyone's machine is in a similar state. For example, I might have volumes I want to keep despite having no associated containers. This command gets rid of those volumes without warning or recourse. That's something that should be emphasized. The past few months have shown me that not everyone in the organization has the same base level knowledge of Docker. We should be cautious about leaving footguns (as Cale calls them) lying around.

robrap: @clintonb This section is just a troubleshooting section in the README.  It was only the first call (dangling image deletion) that @macdiesel added to provisioning.  Also, I've deleted dangling volumes many times. I could argue that you would need to know what you are doing to have a dangling volume that you wouldn't want deleted. :)  I don't think it would hurt to add a comment here that these commands will delete dangling images and volumes without warning.

jmbowman (approved):
I agree there should be a brief warning in the README about the consequences of pruning volumes, otherwise looks good.

clintonb (approved):
Squash your commits.

PR #190, opened 2017-07-19
Add ecommerce course generator script

stevenzheng3791:
LEARNER-1817
https://openedx.atlassian.net/browse/LEARNER-1817

Command that lives in docker devstack that can provision test courses in ecommerce. An equivalent command for studio already exists. See https://github.com/edx/devstack/pull/173 . The final goal will be to have a single make target that will provision each of ecommerce, studio, and mktg with the same course(s).


test-course-generator/test-course-ecommerce.json
```diff
@@ -0,0 +1,17 @@
+{
+	"courses": [
+		{
+			"store": "split",
```
clintonb: The E-Commerce service should not care about Mongo data storage.

clintonb:
https://github.com/edx/ecommerce/pull/1392 must be merged before this is merged.

test-course-generator/test-course-ecommerce.json
```diff
@@ -0,0 +1,16 @@
+{
+	"courses": [
+		{
+			"organization": "test-course-generator",
+			"number": "1",
+			"run": "1",
+			"fields": {
```
BenjiLee: Tabs vs spaces: https://softwareengineering.stackexchange.com/questions/57/tabs-versus-spaces-what-is-the-proper-indentation-character-for-everything-in-e

Fun read about tabs vs spaces: https://stackoverflow.blog/2017/06/15/developers-use-spaces-make-money-use-tabs/

~I don't care which one you use, but in context of this file, your tabs seem to have more spaces than the usual 2 or 4.~

Nvm: Looks like github might be displaying 6 spaces to me.  

Either way, you should keep the tabbing/spacing consistent with other json files in this repo. Especially https://github.com/edx/devstack/blob/master/test-course-studio.json

BenjiLee (approved):

PR #197, opened 2017-07-27
Add memory check warning before dev.up

macdiesel:
Print out a low memory warning before starting docker containers with make.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ dev.clone: ## Clone service repos to the parent directory
 dev.provision.run: ## Provision all services with local mounted directories
 	DOCKER_COMPOSE_FILES="-f docker-compose.yml -f docker-compose-host.yml" ./provision.sh
 
-dev.provision: | dev.provision.run stop ## Provision dev environment with all services stopped
+dev.provision: | check-memory dev.provision.run stop ## Provision dev environment with all services stopped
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -108,3 +108,6 @@ create-course-studio: ## Generates a course on studio using the configurations i
 
 create-course-ecommerce: ## Generates a course on ecommerce using the configurations in test-course-ecommerce.json
 	./course-generator/create-course-ecommerce.sh
+
+check-memory:
+	@if [ `docker info --format '{{json .}}' | python -c "import sys, json; print json.load(sys.stdin)['MemTotal']"` -le 2147483648 ]; then echo "\033[0;31mWarning, System Memory is set too low!!! Increase Docker memory to be at least 2 Gigs\033[0m"; fi || exit 0
```
clintonb: Add a newline.

clintonb (approved):

Makefile
macdiesel: done

Makefile
```diff
@@ -108,3 +108,7 @@ create-course-studio: ## Generates a course on studio using the configurations i
...
```
staubina: Is 2 GB enough?

macdiesel: We'll start here and increment as necessary.

PR #198, opened 2017-07-28
Add Makefile targets and README for easier PDB debugging.

doctoryes:
How's this PR for a direction in enabling PDB debugging of our services using Docker devstack?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,7 +112,7 @@ studio-static: ## Rebuild static assets for the Studio container
 static: | credentials-static discovery-static ecommerce-static lms-static studio-static ## Rebuild static assets for all service containers
 
 studio-attach: ## Attach to the Studio container process to use the debugger & see logs.
-	docker attach `docker ps -aqf "name=edx.devstack.cms"`
+	docker attach `docker ps -aqf "name=edx.devstack.studio"`
 
 healthchecks: ## Run a curl against all services' healthcheck endpoints to make sure they are up. This will eventually be parameterized
 	./healthchecks.sh
--- a/Makefile
+++ b/Makefile
@@ -82,11 +82,17 @@ restore:  ## Restore all data volumes from the host. WARNING: THIS WILL OVERWRIT
 credentials-shell: ## Run a shell on the credentials container
 	docker exec -it edx.devstack.credentials env TERM=$(TERM) bash
 
+credentials-attach: ## Attach to the credentials container process to use the debugger & see logs.
+	docker attach `docker ps -aqf "name=edx.devstack.credentials"`
+
 e2e-shell: ## Start the end-to-end tests container with a shell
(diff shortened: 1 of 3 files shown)


clintonb (commented):
Is there a reason we cannot make this the default? I remain concerned about the proliferation of make targets.

jmbowman:
So far, I can't find any documented drawbacks to just always allocating an interactive tty for these services; I agree with Clinton that would be simpler if it doesn't cause any other problems for us.

doctoryes:
Performance (& security?) concerns are what held me back from integrating it into the main target. I'll assume no security worries since devstack is only for development purposes.

I'll change this PR to add the interactive tty to the main target.

Makefile
```diff
@@ -111,4 +120,3 @@ create-course-ecommerce: ## Generates a course on ecommerce using the configurat
 
 check-memory:
 	@if [ `docker info --format '{{json .}}' | python -c "import sys, json; print json.load(sys.stdin)['MemTotal']"` -lt 2147483648 ]; then echo "\033[0;31mWarning, System Memory is set too low!!! Increase Docker memory to be at least 2 Gigs\033[0m"; fi || exit 0
-
```
clintonb: Restore the newline.

doctoryes:
Ready for re-review!

Makefile
```diff
@@ -82,12 +82,21 @@ restore:  ## Restore all data volumes from the host. WARNING: THIS WILL OVERWRIT
...
 	docker exec -it edx.devstack.lms env TERM=$(TERM) /edx/app/edxapp/devstack.sh open
 
+lms-attach: ## Attach to the LMS container process to use the debugger & see logs.
+	docker attach `docker ps -aqf "name=edx.devstack.lms"`
+
 studio-shell: ## Run a shell on the Studio container
 	docker exec -it edx.devstack.studio env TERM=$(TERM) /edx/app/edxapp/devstack.sh open
 
+studio-attach: ## Attach to the Studio container process to use the debugger & see logs.
+	docker attach `docker ps -aqf "name=edx.devstack.cms"`
```
jmbowman: Shouldn't this be "edx.devstack.studio"?

README.rst
```diff
@@ -366,6 +366,42 @@ PyCharm Integration
...
+
+and your attached session will offer an interactive PDB prompt when the breakpoint is hit.
+
+To detach from the container, you'll need to stop the container with:
+
+.. code:: sh
+
+    make stop
+
+or a manual Docker command to bring down the container.
```
jmbowman: Give an example of the manual command?

Makefile
efischer19: 👍 
```
ericfischer@Eric-Fischers-MacBook-Pro:~/work/devstack$ make studio-attach
docker attach `docker ps -aqf "name=edx.devstack.cms"`
"docker attach" requires exactly 1 argument(s).
See 'docker attach --help'.

Usage:  docker attach [OPTIONS] CONTAINER

Attach local standard input, output, and error streams to a running container
make: *** [studio-attach] Error 1
```

macdiesel (approved):

PR #202, opened 2017-08-02
Mount code in /edx/src for local pip installs

efischer19:
This is key to the development workflows of edx-proctoring and edx-ora2, among others. It more or less copies the Vagrant functionality seen here: https://github.com/edx/configuration/blob/master/vagrant/release/devstack/Vagrantfile#L35

Work in progress, will submit for review once I've confirmed this actually works as expected.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -341,6 +341,11 @@ starts, you have a few options:
   container is started.  For example, the part of the studio command which
   reads ``...&& while true; do...`` could be changed to
   ``...&& pip install my-new-package && while true; do...``.
+* In order to work on locally pip-installed repos like edx-ora2, first clone
+  them into ../src (relative to this directory). Then, inside your lms shell,
+  you can ``pip install -e /edx/src/edx-ora2``. If you want to keep this code
+  installed across stop/starts, modify ``docker-compose.yml`` as mentioned
+  above.
 
 Switching branches
 ------------------
--- a/docker-compose-host.yml
+++ b/docker-compose-host.yml
@@ -13,9 +13,11 @@ services:
   lms:
     volumes:
       - ${DEVSTACK_WORKSPACE}/edx-platform:/edx/app/edxapp/edx-platform:cached
+      - ${DEVSTACK_WORKSPACE}/src:/edx/src:cached
   studio:
     volumes:
       - ${DEVSTACK_WORKSPACE}/edx-platform:/edx/app/edxapp/edx-platform:cached
+      - ${DEVSTACK_WORKSPACE}/src:/edx/src:cached
   forum:
     volumes:
       - ${DEVSTACK_WORKSPACE}/cs_comments_service:/edx/app/forum/cs_comments_service:cached


efischer19:
So this "works", but I ran into an interesting issue.

ORA contains a file upload feature, and one of the potential backend implementations is a filesystem based storage. In testing this feature, a file will be created on the local filesystem named `u"noël.txt"`. This fails out-of-the-box with the following error: `UnicodeEncodeError: 'ascii' codec can't encode character u'\xeb' in position 42: ordinal not in range(128)`.

The way I got things working was to set the `LC_ALL` environment variable to `en_US.UTF-8`, after running `sudo locale-gen en_US.UTF-8` to fix the `/bin/bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)` warning I was seeing.

My questions are:
- Is this a reasonable fix for my issue? It seems like we ought to either guarantee that ORA is running on a system that respects unicode filepaths, or remove that test and only use ascii file paths.
- Is "generate the locale, and set it for `LC_ALL`" the type of thing we ought to do on a base image?

I've also called this discrepancy out on the [currently being updated dev documentation for edx-ora2](https://openedx.atlassian.net/wiki/spaces/EDUCATOR/pages/9765004/ORA+Developer+Guide#Using+ORA+with+docker+devstack).

README.rst
```diff
@@ -324,6 +324,11 @@ starts, you have a few options:
   container is started.  For example, the part of the studio command which
   reads ``...&& while true; do...`` could be changed to
   ``...&& pip install my-new-package && while true; do...``.
+* In order to work on locally pip-installed repos, clone them into ../src
+  (relative to this directory). Then, inside your lms shell, you can do
```
macdiesel: strike "do something like"

docker-compose.yml
```diff
@@ -133,6 +133,7 @@ services:
       # - "18031:18031"
     volumes:
       - edxapp_lms_assets:/edx/var/edxapp/staticfiles/
+      - ../src:/edx/src
```
macdiesel: - s/..src/edxapp_lms_src
- Add edxapp_lms_src to the volumes list on line 158

macdiesel: Should this be done for studio as well?

efischer19: I'm a little confused by this - how will the new `edxapp_lms_src` docker volume be connected to the currently-exists local folder `../src/`?

efischer19: yeah, I'll add it

macdiesel: This will make a new volume that docker will manage so you don't need to worry about creating this src directory.

macdiesel:
@efischer19 I'm going to default to the opinions of @edx/devops on the filesystem encoding. Assuming we want to support it you should add the command to enable this to either the Dockerfile or the ansible role.

feanil:
@jmbowman fixed this issue in the base xenial common files here: https://github.com/edx/configuration/commit/8d2e0fc734f9a1f46da9963d2bd0d4bb9e2ed6b5

It seems like maybe that didn't get pushed to docker-hub or @efischer19 's images are out of date?

jmbowman:
@feanil It looks like our base xenial image was last updated 4 months ago: https://hub.docker.com/r/edxops/xenial-common/tags/ .  Definitely due for a fresh upload.

feanil:
I've made an internal PR to update our automation.  Hopefully we'll get a fresh build out soon.

efischer19:
@macdiesel, I've tested this as it currently exists and it's working for my ora2 development flow. Can you review for merge when you have a chance?

docker-compose-host.yml
```diff
@@ -13,10 +13,12 @@ services:
   lms:
     volumes:
       - ${DEVSTACK_WORKSPACE}/edx-platform:/edx/app/edxapp/edx-platform:cached
+      - ${DEVSTACK_WORKSPACE}/src:/edx/src:cached
```
clintonb, line 16: What happens if this directory doesn't exist on the host machine?

efischer19: I just tested this - the directory is created if it does not exist ahead of time, no errors to speak of.

clintonb (approved):

PR #204, opened 2017-08-02
Add forums IDA

macdiesel:


docker-compose-sync.yml
```diff
@@ -16,6 +16,9 @@ services:
   studio:
     volumes:
       - edxapp-sync:/edx/app/edxapp/edx-platform:nocopy
+  forums:
```
jmbowman: Putting on my DDD "ubiquitous language" hat: we have "forums" here, "forum" above, and our [glossary](http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/glossary.html#f) redirects "forum" to "discussion" (and that documentation seems to have standardized on "discussion" throughout).  We should pick one and stick with it.

jmbowman (commented):
I presume there will need to be some config changes in edx-platform to actually utilize this; is the plan to do that separately after this is merged?  And a minor nitpick, a couple of files are missing trailing newlines.

docker-compose-sync.yml
macdiesel: Should be forum, forgot to change this one.

jmbowman (approved):

SalahAdDin:
@macdiesel Don't forget add the properly repository for forums.

macdiesel:
@SalahAdDin I'm not sure I follow?

SalahAdDin:
It works very well, thank you!

PR #205, opened 2017-08-03
Add note to readme about memory limit and error code 137

mikedikan:


mikedikan:

README.rst
```diff
@@ -543,6 +543,20 @@ up the provisioning process on Mac), so you can try the following:
...
+~~~~~~~~~~~~
+
+While provisioning, some have seen the following error:
+
+.. code:: sh
+
+   ...
+   Build failed running pavelib.assets.update_assets: Subprocess return code: 137
+
+This is an indication that your docker process ran out of memory during execution.
```
robrap: I'd note that return code 137 basically means some went majorly wrong and died.  We have found that this often can be fixed by adjusting memory, but if that doesn't work, more sleuthing will be required.

mikedikan: comment (I believe) addressed - please take another look :)

macdiesel (approved):

README.rst
```diff
@@ -543,6 +543,21 @@ up the provisioning process on Mac), so you can try the following:
...
+~~~~~~~~~~~~
+
+While provisioning, some have seen the following error:
+
+.. code:: sh
+
+   ...
+   Build failed running pavelib.assets.update_assets: Subprocess return code: 137
+
+This is an indication that your docker process died during execution.  Most likely,
```
robrap, line 556: Nit: What does it actually kill?  I'm unclear if it is the original docker process that dies, or just the additional command you ran, etc.?

README.rst
```diff
@@ -543,6 +543,21 @@ up the provisioning process on Mac), so you can try the following:
...
+this is due to running out of memory (this has historically been the case at edX).
+If you are set to 2GB (docker for mac default) up it to 4GB (the edX recommendation).
```
robrap: Some minor changes I would make are detailed here:
```
... Most likely, this is due to running out of memory. If you are set to 2GB (docker for mac default), increase it to 4GB (the current recommendation). If you are set to 4GB, try 6GB.
```

robrap (approved):
I recommended some minor changes, but this would get the job done as-is as well.  Thanks for creating the PR!

README.rst
mikedikan: The entire process (make provision) dies, resulting in a exit in the middle of provisioning (very annoying, as we need to restart provisioning)  Should I try to make that more clear?  Suggested language would be helpful if so.

README.rst
mikedikan: Sounds like you want to strip out the personality here 😝

robrap: I like the personality, but am concerned for people where english might be their second language.  I also removed references to "edX" that weren't needed, or could have been "Open edX", yadda yadda yadda.

README.rst
robrap: Just wondering.  I don't think it needs any changes.  Someone will know what it means when they see the error. :)

staubina: @robrap You can get this error in any container as a note.  I had this error in my MySql container when I ran into this issue.

staubina (approved):

PR #206, opened 2017-08-07
Create master command for course-generator tool

stevenzheng3791:
LEARNER-2103
https://openedx.atlassian.net/browse/LEARNER-2103

Important things to note:
1. There is now one script that handles the provisioning for each of the services (can still choose which services to use)
2. There is a master test-course.json (contains a complete list of configurations) that can be edited. This is the course that will be created when you run ```make create-test-course```. (You can define other course config files but to use them you will need to run the course provisioning script directly)
3. There is a course-builder script that will build the json for a list of courses specified in the script, for developers requiring a suite of courses. I've added a list of course-mode permutations to this script, but the intention is that the list, as well as the functions, can be modified to suit the developer. Running ```make build-courses``` will run this script and automatically provision those courses.


stevenzheng3791:
I consider this a rough draft, but let me know what you guys think. I'd like to be able to demo course mode permutations since that was an important goal, which is the primary reason behind the course-builder script, even though that's not necessarily part of the AC

Makefile
```diff
@@ -103,11 +103,15 @@ vnc-passwords: ## Get the VNC passwords for the Chrome and Firefox Selenium cont
 mysql-shell: ## Run a shell on the mysql container
 	docker-compose exec mysql bash
 
-create-course-studio: ## Generates a course on studio using the configurations in test-course.json
-	./course-generator/create-course-studio.sh ./course-generator/test-course.json
-
-create-course-ecommerce: ## Generates a course on ecommerce using the configurations in test-course.json
-	./course-generator/create-course-ecommerce.sh ./course-generator/test-course.json
+# Provisions studio, ecommerce, and marketing course(s) in test-course.json
+create-test-course: ## NOTE: can only provision courses to marketing if you have access to it
```
BenjiLee: This note is vague. What do you mean by access and who is needing access?

stevenzheng3791: I will fix the note to be more clear. So, by default it will attempt to create the course in marketing, but for those outside of edX, if my understanding is correct, then that attempt would fail

BenjiLee:
The marketing site is not working for me so I can't confirm it all working together but looks like LMS and ecommerce and linking correctly.

course-generator/create-courses.sh
```diff
@@ -0,0 +1,46 @@
+# Script that provisions studio, ecommerce, marketing with courses
+# USAGE: ./create-courses [--studio] [--ecommerce] [--marketing] course-config.json
+studio=false
+ecommerce=false
+marketing=false
+echo "Parsing options"
+for arg in "$@"; do
+    if [ $arg == "--studio" ]; then
+    	studio=true
+    	echo "Course creation for STUDIO is turned on"
```
BenjiLee: Wording of these log messages are kind of awkward. Turning something on sounds like we hit a binary switch somewhere. Perhaps something like "Creating course for {}" or "Course will be created for {}"

course-generator/create-courses.sh
```diff
@@ -0,0 +1,46 @@
...
+    elif [ $arg == "--ecommerce" ]; then
+    	ecommerce=true
+    	echo "Course creation for ECOMMERCE is turned on"
+    elif [ $arg == "--marketing" ]; then
+    	marketing=true
+    	echo "Course creation for MARKETING is turned on"
+    fi
+done
+echo "Done parsing options. Make sure correct services are turned on"
```
BenjiLee: Is there a way to verify if a service is turned on when we check for each service above?

Makefile
```diff
@@ -103,11 +103,15 @@ vnc-passwords: ## Get the VNC passwords for the Chrome and Firefox Selenium cont
...
-create-course-studio: ## Generates a course on studio using the configurations in test-course.json
-	./course-generator/create-course-studio.sh ./course-generator/test-course.json
-
-create-course-ecommerce: ## Generates a course on ecommerce using the configurations in test-course.json
-	./course-generator/create-course-ecommerce.sh ./course-generator/test-course.json
+# Provisions studio, ecommerce, and marketing with course(s) in test-course.json
+create-test-course: ## NOTE: marketing course creation is not available for those outside edx
+	./course-generator/create-courses.sh --studio --ecommerce --marketing course-generator/test-course.json
+
+# Run the course json builder script and create those courses
```
bderusha: what are "those" courses?

Makefile
```diff
@@ -103,11 +103,15 @@ vnc-passwords: ## Get the VNC passwords for the Chrome and Firefox Selenium cont
 mysql-shell: ## Run a shell on the mysql container
 	docker-compose exec mysql bash
 
-create-course-studio: ## Generates a course on studio using the configurations in test-course.json
-	./course-generator/create-course-studio.sh ./course-generator/test-course.json
-
-create-course-ecommerce: ## Generates a course on ecommerce using the configurations in test-course.json
-	./course-generator/create-course-ecommerce.sh ./course-generator/test-course.json
+# Provisions studio, ecommerce, and marketing with course(s) in test-course.json
+create-test-course: ## NOTE: marketing course creation is not available for those outside edx
```
bderusha: NIT: `edX` Capital X

Makefile
```diff
@@ -103,11 +103,15 @@ vnc-passwords: ## Get the VNC passwords for the Chrome and Firefox Selenium cont
...
+	./course-generator/create-courses.sh --studio --ecommerce --marketing course-generator/test-course.json
+
+# Run the course json builder script and create those courses
+build-courses: ## NOTE: marketing course creation is not available for those outside edx
```
bderusha: NIT: `edX` Capital X

Makefile
bderusha: A more thorough description of what this make target does would be helpful here, and any steps required beforehand (like updating a config file, etc)

course-generator/create-courses.sh
stevenzheng3791: Actually, I meant this to be a way for the users to know that they entered the options correctly. It seems like there could be a way to check if the services are actually working (mainly that you can activate the shell for them), I'll look into that.

stevenzheng3791:
@BenjiLee @bderusha Ready for another review

bderusha (approved):
looks good! 👍

PR #207, opened 2017-08-10
Mem check and e2e command updates

jmbowman:
* I noticed that the check-memory target was failing under Python 3.5 in Travis; updated it to work with Python 2 and 3. This shouldn't be causing any real trouble, but it's an obvious error in the job output that can be distracting when trying to debug other issues.
* A couple of basic enterprise e2e tests were recently added which aren't meant to be run as part of the main test suite; updated the recommended command for running the e2e tests to match what that repo is now using.


macdiesel (approved):
Thank you.

PR #208, opened 2017-08-14
PLAT-1661 Add static asset make targets

jmbowman:
Add make targets to rebuild static assets in each container or for all containers.  The IDAs all store the built assets inside the source directory being mapped in from the host, so they don't need separate asset volumes like the ones we added for LMS and Studio.  The provisioning scripts for each service already build static assets, so nothing needs to be added there.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,20 @@ lms-shell: ## Run a shell on the LMS container
 studio-shell: ## Run a shell on the Studio container
 	docker exec -it edx.devstack.studio env TERM=$(TERM) /edx/app/edxapp/devstack.sh open
 
+%-static: ## Rebuild static assets for the specified service container
+	docker exec -t edx.devstack.$* bash -c 'source /edx/app/$*/$*_env && cd /edx/app/$*/$*/ && make static'
+
+credentials-static: ## Rebuild static assets for the credentials container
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -91,6 +91,20 @@ lms-shell: ## Run a shell on the LMS container
 studio-shell: ## Run a shell on the Studio container
 	docker exec -it edx.devstack.studio env TERM=$(TERM) /edx/app/edxapp/devstack.sh open
 
+%-static: ## Rebuild static assets for the specified service container
+	docker exec -t edx.devstack.$* bash -c 'source /edx/app/$*/$*_env && cd /edx/app/$*/$*/ && make static'
+
+credentials-static: ## Rebuild static assets for the credentials container
+	docker exec -t edx.devstack.credentials bash -c 'make static'
```
macdiesel, line 98: can't use the %-static: target for this?

jmbowman: discovery and ecommerce use an env file to activate the virtualenv, etc.; credentials doesn't have such a file (or use a virtualenv), so it can't use the same command.

macdiesel: I <3 snowflakes.

macdiesel (approved):

PR #210, opened 2017-08-15
Watch edx-platform assets

efischer19:
# [PLAT-1663](https://openedx.atlassian.net/browse/PLAT-1663)

Essentially, I'm modifying the launch command for both `lms` and `studio` to kick off a `paver watch_assets` background process.

Work in progress, pairs with https://github.com/edx/edx-platform/pull/15817. This seemingly works on my local setup.

FYI @arizzitano @macdiesel

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -116,7 +116,7 @@ services:
       - "18130:18130"
 
   lms:
-    command: bash -c 'source /edx/app/edxapp/edxapp_env && while true; do python /edx/app/edxapp/edx-platform/manage.py lms runserver 0.0.0.0:18000 --settings devstack_docker; sleep 2; done'
+    command: bash -c 'cd /edx/app/edxapp/edx-platform && source ../edxapp_env && (nohup paver watch_assets -b &) && while true; do python manage.py lms runserver 0.0.0.0:18000 --settings devstack_docker; sleep 2; done'
     container_name: edx.devstack.lms
     depends_on:
       - mysql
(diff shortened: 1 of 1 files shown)


clintonb (changes requested):
Docker containers are supposed to run one process. If you need to run another process, spin up another container. It would be ideal if that container was based on an image that is specially-purposed for asset compilation.

clintonb:
https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#each-container-should-have-only-one-concern

clintonb:
@gsong and I discussed some ideas about this a couple hours ago at lunch. Let me know if you'd like to chat further.

macdiesel:
@clintonb this may be a bit of an overstatement, I have found this in the docker documentation:
https://docs.docker.com/engine/admin/multi-service_container/

"It is generally recommended that you separate areas of concern by using one service per container. That service may fork into multiple processes (for example, Apache web server starts multiple worker processes). **It’s ok to have multiple processes**, but to get the most benefit out of Docker, avoid one container being responsible for multiple aspects of your overall application."

To be honest, I'm not sure that this is something that would benefit from being in multiple containers as it's tied directly to django serving of assets while in development mode. 

Perhaps what would be better here would be to start using the supervisor process to launch both the Django app as well as the asset watcher.

efischer19:
This (how/if to handle multi-process containers) seems like it's going to be a fairly in-depth discussion, I've created a wiki page here: https://openedx.atlassian.net/wiki/spaces/~efischer/pages/163090911/Docker+multi-process+ruminations

gsong:
Addressed with #230.

PR #211, opened 2017-08-16
Deprecate provision command

macdiesel:


Makefile
```diff
@@ -38,8 +38,8 @@ dev.sync.requirements: ## Install requirements
 dev.sync.up: dev.sync.daemon.start ## Bring up all services with docker-sync enabled
 	docker-compose -f docker-compose.yml -f docker-compose-sync.yml up -d
 
-provision: | check-memory ## Provision all services using the Docker volume
-	./provision.sh
+provision: | dev.provision ## Provision all services using the Docker volume
+	echo "\033[0;31mThis command will be deprecated in a future release, use dev.provision\033[0m"
```
jmbowman: Is this going to appear at the end of a many-minute provisioning process with hundreds of lines of output?  If so, would it be more prominent to just put the deprecation warning in the help comment, which appears in the `make help` output?

jmbowman (approved):

PR #214, opened 2017-08-18
Rename compose-host to compose.override

gsong:
Follow the Docker community convention: https://docs.docker.com/compose/extends/

> When you run `docker-compose up` it reads the overrides automatically.

1. This would allow developers to explore `docker-compose` commands much more easily.
2. It's the Docker way.

/cc @andy-armstrong @clintonb


andy-armstrong:
@gsong I started to do the same thing yesterday, but then noticed that dev.sync.up needs to be able to use a different set of volumes, which it gets from docker-compose-sync.yml instead of docker-compose-hosts.yml. Given the discussion about sync no longer being necessary maybe that's okay, but I'd want to be sure about that before making this change. What do you think?

andy-armstrong:
Also, @clintonb is on vacation, so maybe @jmbowman or @macdiesel should review it instead.

gsong:
@andy-armstrong The docker-sync version will just continue to use its own set of docker-compose files, so we should be OK. This just allows you to use `docker-compose` in the most normal case—development mode without docker-sync.

macdiesel (commented):
With PR #211 we have begun to make the assumption that this repo and setup will only be used for a development environment (i.e. no full stack). Assuming this is the case, would it be better to add these volumes from the override file directly to the docker-compose.yml file rather than to load them in as overrides? 

The overrides file seems redundant and hides key configuration information from an inexperienced user.

gsong:
Sounds like a plan. This way we can leave override as a local git ignored tweak that each developer can make on their own. I'll make that adjustment soon.

kdmccormick:
Closing this as outdated.

PR #215, opened 2017-08-22
Ignore `local.mk` to store convenience make targets.

UmanShahzad:
I was getting tired of typing `docker-compose restart <service>` which is slightly longer than `make lms-restart` or just `make restart` for everything, so added these here.

Along with it, updated some relevant parts of the documentation (and very minor spacing issues that were causing bad rendering).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -208,6 +208,8 @@ simply use the ``docker-compose restart`` command:
 -  lms
 -  studio
 
+If you'd like to add some convenience make targets, you can add them to a `local.mk` file, ignored by git.
+
 Payments
 --------
 
--- a/.gitignore
+++ b/.gitignore
@@ -97,3 +97,5 @@ ENV/
 .docker-sync/
 .env
 
+# Personal makefile extensions
+local.mk


clintonb (changes requested):
I like the documentation changes, but would prefer we not try to create a Make target for every Docker command. Just use Docker! 

Also, why do you need to consistently restart containers?

UmanShahzad:
@clintonb Haha, right, I didn't like the solution too much either in general, it was just for my convenience. But in that case, it'd be more reasonable to git-ignore some file type which we can then include in the main Makefile, allowing people their own commands without having unstaged work sitting around. Since we already include `*.mk`, let's just move what's in there into `Makefile` and gitignore it for the future?

Also, I was messing around a little with the containers which lead me to have to restart them, but not anymore. Still, since the above possibility came up, maybe we can do something about it.

UmanShahzad:
@clintonb I pushed a solution that implements my above comment's suggestion.

Now I put the restart commands into my `convenience.mk`, and will probably be adding much more convenience commands :)

Makefile
```diff
@@ -146,3 +146,18 @@ build-courses: ## NOTE: marketing course creation is not available for those out
 
 check-memory:
 	@if [ `docker info --format '{{json .}}' | python -c "from __future__ import print_function; import sys, json; print(json.load(sys.stdin)['MemTotal'])"` -lt 2147483648 ]; then echo "\033[0;31mWarning, System Memory is set too low!!! Increase Docker memory to be at least 2 Gigs\033[0m"; fi || exit 0
+
```
clintonb: Revert these changes. There is already a separate file for the marketing site. Marketing site targets should always be there as it is not part of Open EdX.

README.rst
```diff
@@ -194,19 +196,18 @@ Useful Commands
 ---------------
 
 Sometimes you may need to restart a particular application server. To do so,
-simply use the ``docker-compose restart`` command:
```
clintonb: Revert your changes about restarts.

README.rst
```diff
@@ -682,6 +684,26 @@ docker-sync, but this feature hasn't been fully implemented yet (as of
 Docker 17.06.0-ce, "delegated" behaves the same as "cached").  There is a
 GitHub issue which explains the `current status of implementing delegated consistency mode`_.
 
+.. _available services:
+
+Available Services
```
clintonb: This should be close to the top of the document, not the bottom.

clintonb (changes requested):
Start small. Revert all of these changes, because you've now gone in a worse direction by combining marketing targets in this manner. Just ignore one file, local.mk, and document how users can add their own targets.

UmanShahzad:
@clintonb I will be getting to this now -- sorry for the delay!

UmanShahzad:
@clintonb Thanks for the review. I undid the Makefile changes, allowing instead for a single `local.mk` for convenience make targets. Also updated the documentation to move the service list up to a better context, and added a note about the existence of `local.mk`.

README.rst
```diff
@@ -169,6 +171,26 @@ is ``edx``.
 | verified   | verified@example.com   |
 +------------+------------------------+
 
+.. _available services:
```
clintonb: I'm on the fence about this section. The table below informs the user which services are installed. I'm not certain what additional value this new section adds. I'm also concerned it will grow stale, as the table below his since it lacks a URL for the forums service.

What value do you gain from this list that is not gained from the table or opening the docker-compose file? How did you determine which services are available?

UmanShahzad: I'd prefer to just scratch it, it doesn't really matter.

UmanShahzad:
@clintonb In order to keep the diff clean and let https://github.com/edx/devstack/pull/224 take care of the doc nits, I'll be just totally undoing the `README` changes minus the note about `local.mk`, and this PR will just be about allowing convenience make targets through that `local.mk`.

clintonb (approved):

PR #219, opened 2017-08-29
(WIP) first attempt at running webpack container from image (which must be …

michaelroytman:
…built locally)


macdiesel:
@MichaelRoytman @efischer19 How is this different from #230?

efischer19:
@macdiesel I believe this was a precursor to that. Instead of building a webpack container specifically for studio-frontend, we expanded to include a full asset watcher. This PR can be closed.

PR #220, opened 2017-08-30
Updated discovery provisioning

clintonb:
Provisioning the Discovery Service container now properly creates a Site instance.

PLAT-1715


clintonb:

PR #223, opened 2017-09-07
Watch assets in separate container

efischer19:


efischer19:
This kind of works, but tests are failing due to some sort of "no space on device" issue. I'll keep poking next week, but feel free to take a look at what's already here.

macdiesel (approved):
You can ignore the disk space error and merge this if the tests pass locally.

docker-compose.yml
```diff
@@ -149,6 +149,19 @@ services:
     volumes:
       - edxapp_lms_assets:/edx/var/edxapp/staticfiles/
 
+  lms_watcher:
+    command: bash -c 'cd /edx/app/edxapp/edx-platform && source ../edxapp_env && while true; do paver watch_assets; sleep 2; done'
+    container_name: edx.devstack.lms_watcher
+    depends_on:
+      - lms
+    stdin_open: true
```
gsong: Is there a reason for needing `stdin` and `tty`?

docker-compose.yml
```diff
@@ -149,6 +149,19 @@ services:
     volumes:
       - edxapp_lms_assets:/edx/var/edxapp/staticfiles/
 
+  lms_watcher:
+    command: bash -c 'cd /edx/app/edxapp/edx-platform && source ../edxapp_env && while true; do paver watch_assets; sleep 2; done'
+    container_name: edx.devstack.lms_watcher
+    depends_on:
+      - lms
```
gsong: Does this service really depend on `lms` service starting?

efischer19: Yeah you're right - without the lms process we can still watch assets, it's just kind of pointless b/c nothing is consuming them. No reason to force that dependence.

docker-compose.yml
efischer19: Nope, I just copy-pasted :)

I've tested and things look fine with this removed, I'll update.

efischer19:
Failure is the disk space issue again, merging.

PR #224, opened 2017-09-08
Fix syntax problems with README.

doctoryes:
Made corrections during a read-through of the whole README.


macdiesel (approved):
Thank you.

PR #230, opened 2017-09-19
Efischer/watcher timeout

efischer19:
WIP, nomerge pls

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,7 +120,7 @@ lms-shell: ## Run a shell on the LMS container
 	docker exec -it edx.devstack.lms env TERM=$(TERM) /edx/app/edxapp/devstack.sh open
 
 lms-attach: ## Attach to the LMS container process to use the debugger & see logs.
-	docker attach `docker ps -aqf "name=edx.devstack.lms"`
+	docker attach `docker ps -aqf "name=edx.devstack.lms$$"`
 
 studio-shell: ## Run a shell on the Studio container
 	docker exec -it edx.devstack.studio env TERM=$(TERM) /edx/app/edxapp/devstack.sh open
@@ -140,7 +140,7 @@ studio-static: ## Rebuild static assets for the Studio container
 static: | credentials-static discovery-static ecommerce-static lms-static studio-static ## Rebuild static assets for all service containers
 
 studio-attach: ## Attach to the Studio container process to use the debugger & see logs.
-	docker attach `docker ps -aqf "name=edx.devstack.studio"`
+	docker attach `docker ps -aqf "name=edx.devstack.studio$$"`
 
 healthchecks: ## Run a curl against all services' healthcheck endpoints to make sure they are up. This will eventually be parameterized
 	./healthchecks.sh
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 ########################################################################################################################
 #
 # When adding a new target:
-#   - If you are adding a new service make sure the dev.clean target will fully reset said service.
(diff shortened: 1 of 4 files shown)


efischer19:
This is seemingly working for me locally, when paired with https://github.com/edx/edx-platform/pull/16051

Makefile
```diff
@@ -119,12 +122,18 @@ update-db: | studio-update-db lms-update-db discovery-update-db ecommerce-update
 lms-shell: ## Run a shell on the LMS container
 	docker exec -it edx.devstack.lms env TERM=$(TERM) /edx/app/edxapp/devstack.sh open
 
+lms-watcher-shell: ## Run a shell on the LMS watcher container
+	docker exec -it edx.devstack.lms_watcher env TERM=$(TERM) /edx/app/edxapp/devstack.sh open
+
 lms-attach: ## Attach to the LMS container process to use the debugger & see logs.
-	docker attach `docker ps -aqf "name=edx.devstack.lms"`
+	docker attach `docker ps -aqf "name=edx.devstack.lms$$"`
```
gsong: Why not just attach directly to the container name: `docker attach edx.devstack.lms`?

efischer19: That's a good question! @doctoryes, I think you added this logic in https://github.com/edx/devstack/pull/198 - does George's suggestion seem reasonable, or was there a specific reason for filtering the results of `docker ps`?

doctoryes: No specific reason - guess I didn't know that the container name itself would just work.

doctoryes: Simpler is nearly always better!

macdiesel:
@efischer19 do you think there should be a target to stop just the watchers? dev.watchers.stop or something similar.

macdiesel:
Or a target to start just the watchers? I guess you would just run dev.up.all even if you had already had the other set of containers already running?

efischer19:
Yeah, that's the part I'm least sure of - what's most useful to most workflows? Should we add both the `dev.up.all` target as well as a new watchers-only target? Just the latter? How have we been building the Makefile to this point?

I'll update so that dev.up.all/stop both work for *everything*, and give the watchers their own start/stop as well. Totally open to further discussion here, but IMO it feels like it'll be less hassle to have too many targets than it would be to not have enough.

gsong:
I like @macdiesel's idea. That way `dev.up.all` can just be a compositional task.

efischer19:
Updated. `dev.up` and `stop` both continue to function as before. New targets `dev.up.watchers` and `stop.watchers` allow you to start and stop asset watching as desired, while `dev.up.all` and `stop.all` combine the previous targets compositionally.

macdiesel (approved):
Thanks for the PR @efischer19.

Makefile
```diff
@@ -54,11 +59,16 @@ provision: | dev.provision ## This command will be deprecated in a future releas
 
 stop: ## Stop all services
 	(test -d .docker-sync && docker-sync stop) || true ## Ignore failure here
-	docker-compose stop
+	docker-compose -f docker-compose.yml stop
```
gsong, line 62: Remove `-f` option.

Makefile
```diff
@@ -38,6 +38,11 @@ dev.repo.reset: ## Attempts to reset the local repo checkouts to the master work
 dev.up: | check-memory ## Bring up all services with host volumes
 	docker-compose -f docker-compose.yml -f docker-compose-host.yml up -d
 
+dev.up.watchers: | check-memory ## Bring up asset watcher containers
+	docker-compose docker-compose-watchers.yml up -d
```
gsong, line 42: Is this missing a `-f`?

PR #234, opened 2017-09-25
Updated README to include a tip about resetting a database

BenjiLee:


README.rst
```diff
@@ -578,6 +578,25 @@ Start over
...
+Resetting a database
+~~~~~~~~~~~~~~~~~~~~
+
+In case you botched a migration or just want to start with a clean database.
+
+1. Open up the mysql shell and drop the database for the desired service. 
+
+```
+>>> make mysql-shell
+>>> DROP DATABASE (insert database here)
```
macdiesel: I think you need to run the mysql command first so you can be at the mysql prompt before you can run this sql statement.

BenjiLee: Updated

README.rst
```diff
@@ -578,6 +578,26 @@ Start over
 If you want to completely start over, run ``make destroy``. This will remove
 all containers, networks, AND data volumes.
 
+Resetting a database
+~~~~~~~~~~~~~~~~~~~~
+
+In case you botched a migration or just want to start with a clean database.
+
+1. Open up the mysql shell and drop the database for the desired service.
```
macdiesel, line 586: Thinking about this more, wouldn't running something like: ```./load-db.sh edxapp``` fix the issue was well without all the shell finagling?

BenjiLee: Doesn't that require a database file to exist? If that's the case, should I try to add the sql file for course-discovery?

macdiesel: This seems preferable to me. @clintonb is there a reason why this wasn't done for discovery during the initial phase of work?

clintonb: The edxapp and ecommerce database dumps were created to speed up provisioning because running migrations is slow for these services—edxapp because it's still on Django 1.8, ecommerce because there are a lot of models. Running migrations for credentials and discovery is quite fast, thus no need for a database dump.

Given that we haven't kept the current DB dumps updated, I would not advise recommending this procedure for other services until we have a method of automating DB dumps (assuming we want that).

macdiesel: Thank you.

PR #237, opened 2017-09-26
Added volume to store discovery's assets

clintonb:
This will ensure that uploads remain despite containers being dropped and re-created.

LEARNER-2469


clintonb:

macdiesel (approved):

clintonb:
The build failure is unrelated to the new volume. Merging.

PR #242, opened 2017-10-05
Marketing site updates

clintonb:
- Added make target to bring containers down
- Added volume for all application data
- Added volume for private key to download data from production


clintonb:
FYI @edx/learner this change coupled with https://github.com/edx/edx-mktg/pull/1656 should allow folks to run the `sync-prod-db` and `sync-prod-files` targets from the `/edx/app/edx-mktg/edx-mktg` directory inside the container.

PR #245, opened 2017-10-16
Update marketing make tasks.

robrap:
- Add stop.
- Fix message for down.
- Add help-marketing.

FYI: @edx/docker-devstack-working-group 

**Before merge**
- [ ] Squash

### DIFF
--- a/marketing.mk
+++ b/marketing.mk
@@ -1,7 +1,14 @@
+help-marketing: ## Display this help message
+	@echo "Please use \`make <target>' where <target> is one of"
+	@perl -nle'print $& if m{^[\.a-zA-Z_-]+:.*?## .*$$}' $(MAKEFILE_LIST) | grep marketing | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m  %-25s\033[0m %s\n", $$1, $$2}'
+
 marketing-shell: ## Run a shell on the marketing site container
 	docker exec -it edx.devstack.marketing env TERM=$(TERM) bash
 
-down-marketing:   ## Bring up all services (including the marketing site) with host volumes
+stop-marketing:   ## Stop all services (including the marketing site) with host volumes
+	docker-compose -f docker-compose.yml -f docker-compose-host.yml -f docker-compose-marketing-site.yml -f docker-compose-marketing-site-host.yml stop
+
+down-marketing:   ## Bring down all services (including the marketing site) with host volumes
 	docker-compose -f docker-compose.yml -f docker-compose-host.yml -f docker-compose-marketing-site.yml -f docker-compose-marketing-site-host.yml down
 
 up-marketing:   ## Bring up all services (including the marketing site) with host volumes


marketing.mk
```diff
@@ -1,7 +1,14 @@
+help-marketing: ## Display this help message
```
robrap, line 1: Reviewers: If you don't like my hacky "grep marketing" I can kill this.  But it would be nice to add this to the marketing wiki page.

marketing.mk
```diff
@@ -1,7 +1,14 @@
...
+	@echo "Please use \`make <target>' where <target> is one of"
+	@perl -nle'print $& if m{^[\.a-zA-Z_-]+:.*?## .*$$}' $(MAKEFILE_LIST) | grep marketing | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m  %-25s\033[0m %s\n", $$1, $$2}'
+
 marketing-shell: ## Run a shell on the marketing site container
 	docker exec -it edx.devstack.marketing env TERM=$(TERM) bash
 
-down-marketing:   ## Bring up all services (including the marketing site) with host volumes
+stop-marketing:   ## Stop all services (including the marketing site) with host volumes
+	docker-compose -f docker-compose.yml -f docker-compose-host.yml -f docker-compose-marketing-site.yml -f docker-compose-marketing-site-host.yml down
```
robrap: Oops.  down => stop coming. :)

marketing.mk
```diff
@@ -1,7 +1,14 @@
+help-marketing: ## Display this help message
+	@echo "Please use \`make <target>' where <target> is one of"
+	@perl -nle'print $& if m{^[\.a-zA-Z_-]+:.*?## .*$$}' $(MAKEFILE_LIST) | grep marketing | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m  %-25s\033[0m %s\n", $$1, $$2}'
```
macdiesel, line 3: Think we could do this with python instead of perl similar to what we do here: https://github.com/edx/devstack/blob/master/Makefile#L201 ?

robrap: @macdiesel.  Thanks.  I mostly wanted the stop command.  I will probably just have to drop the help for now.  I'm in a place where I'm feeling wildly slow and behind on my current task, so it's bad timing on resolving this help issue.

Note that I just copy/pasted the other help task and added the grep:
https://github.com/edx/devstack/blob/master/Makefile#L20

robrap:
Sorry to do this @macdiesel, but I just reverted the help task.  Hope that is ok so I can just close this PR out.  Thanks.

marketing.mk
macdiesel: Did not realize it was a copy/pasta out of the makefile. If it already exists lets take it. Sorry for the confusion.

robrap:
@macdiesel: I reverted the revert, so the help is back.  I'll merge once I see a thumb and green build.  Thanks!

macdiesel (approved):

PR #248, opened 2017-10-27
Add lms-restart and studio-restart make targets

thallada:
Kills the running Django development server and lets the watcher process restart it. Thanks @doctoryes for the original kill command.


efischer19 (approved):
LGTM, great addition 👍

doctoryes:
👍

PR #252, opened 2017-11-06
Update PyCharm test setup docs for pytest

jmbowman:
This seems to work for me, but I'm not a heavy user of running tests through PyCharm; let me know if this looks incorrect or incomplete to anyone.

### DIFF
--- a/docs/pycharm_integration.rst
+++ b/docs/pycharm_integration.rst
@@ -197,34 +197,27 @@ configuration with the following options:
 Setup a Run/Debug Configuration for python tests for LMS or Studio
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To run and debug unit tests, create a **"Django server"** type Run/Dubug
-configuration (not "Django tests" as is done for IDAs) with the following
-options:
+To run and debug unit tests, edit the **"Defaults -> Python tests -> py.test"** type Run/Dubug
+configuration with the following options:
 
-1. Leave host/port blank
-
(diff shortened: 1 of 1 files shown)


clintonb (approved):
It's probably worth noting that, if you attempt to run a single test class that subclasses `TestCase` (by right-clicking on the class), PyCharm will not use `py.test`, but will default to using `unittest`.

jmbowman:
After tinkering around with it for a while and hunting through the PyCharm bug tracker, I discovered that the Django support feature currently overrides the default test runner.  So you need to either disable Django support or (starting in 2017.3 EAP) opt not to use the Django test runner.  I've updated the doc accordingly.

docs/pycharm_integration.rst
```diff
@@ -197,34 +197,34 @@ configuration with the following options:
...
-5. Python Interpreter: Choose the Docker Compose interpreter for this
-   service.
+Then make two changes in PyCharm's preferences:
 
-6. Working directory: /edx/app/edxapp/edx-platform
+1. Set "Tools -> Python Integrated Tools -> Default test runner" to "py.test".
 
-7. Path mappings (add mapping):
+2. In "Languages & Frameworks -> Django", uncheck "Enable Django Support".
+   Starting in PyCharm 2017.3 (the EAP of which is already available), this
```
clintonb, line 220: We should have a ticket to update these docs once 2017.3 is released.

clintonb (approved):

docs/pycharm_integration.rst
jmbowman: Done: [PLAT-1787](https://openedx.atlassian.net/browse/PLAT-1787)

macdiesel (approved):
Tested and works in my environment.

PR #256, opened 2018-01-16
Put make pull higher up in the docs

iloveagent57:
* Make the down/pull/dev.up cycle clear at the top of the docs. 
* Add convenience make directive to pop into a mysql edxapp shell.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -49,6 +49,20 @@ You will also need the following installed:
 - make
 - python pip (optional for MacOS)
 
+Using the Latest Images
+-----------------------
+
+New images for our services are published frequently.  Assuming that you've followed the steps in `Getting Started`_
+below, run the following sequence of commands if you want to use the most up-to-date versions of the devstack images.
+
+.. code:: sh
+
+    make down
+    make pull
+    make dev.up
+
+This will stop any running devstack containers, pull the latest images, and then start all of the devstack containers.
+
 Getting Started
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -21,6 +21,22 @@ FYI
...
+Daily Development
+-----------------
+
+New images for our services are published frequently, so it's best to get in the habit of always running
+the most up-to-date images.  Assuming that you've followed the steps in `Getting Started`_ below, *always* run the
+following sequence of commands before starting any development work:
+
+.. code:: sh
+
+    make down
```
robrap: The only reason I would hesitate on this is that the latest is not always stable (including right now), and it just means that many more people getting stuck on issues.  A counter-argument is that it puts a lot more pressure on and infrastructure that maintains a stable set of images.

robrap: @iloveagent57: 
1. I didn't event get a response before you merged?
2. Why not just have people use "make dev.reset" which does the same, but also updates static and dbs?

iloveagent57: Sorry, saw the green button and got excited.  I changed the language to be more specific to running the latest version of images, rather than "you should to this every day".

It looks like "make dev.reset" will reset the local repo to origin/master, which you probably don't want if you're in the middle of development.

robrap: True.  Thanks for responding. :)

PR #259, opened 2018-01-19
FEDX-500 unblock everyone

efischer19:
Prevents the ENOENT error described in [FEDX-500](https://openedx.atlassian.net/browse/FEDX-500)

As outlined in https://docs.docker.com/compose/compose-file/compose-file-v2/#volumes, this will create a volume for node_modules that is not mounted in the host at all, which will hopefully resolve the weird filesystem issues were fighting ATM.

Early returns are positive on a clean install, I've made it past the initial `paver install_prereqs` step. Host reports an empty `node_modules/` under edx-platform, docker says it's fine (and I've confirmed in a separate shell that `edx-platform/node_modules/.bin/coffee` exists).

I'll let the full provision sequence finish before declaring this "good"

### DIFF
--- a/docker-compose-host.yml
+++ b/docker-compose-host.yml
@@ -4,21 +4,31 @@ services:
   credentials:
     volumes:
       - ${DEVSTACK_WORKSPACE}/credentials:/edx/app/credentials/credentials:cached
+      - credentials_node_modules:/edx/app/credentials/credentials/node_modules
   discovery:
       volumes:
       - ${DEVSTACK_WORKSPACE}/course-discovery:/edx/app/discovery/discovery:cached
+      - discovery_node_modules:/edx/app/discovery/discovery/node_modules
   ecommerce:
     volumes:
       - ${DEVSTACK_WORKSPACE}/ecommerce:/edx/app/ecommerce/ecommerce:cached
+      - ecommerce_node_modules:/edx/app/ecommerce/ecommerce/node_modules
   lms:
     volumes:
       - ${DEVSTACK_WORKSPACE}/edx-platform:/edx/app/edxapp/edx-platform:cached
+      - edxapp_node_modules:/edx/app/edxapp/edx-platform/node_modules
       - ${DEVSTACK_WORKSPACE}/src:/edx/src:cached
   studio:
     volumes:
       - ${DEVSTACK_WORKSPACE}/edx-platform:/edx/app/edxapp/edx-platform:cached
+      - edxapp_node_modules:/edx/app/edxapp/edx-platform/node_modules
       - ${DEVSTACK_WORKSPACE}/src:/edx/src:cached
   forum:
     volumes:
       - ${DEVSTACK_WORKSPACE}/cs_comments_service:/edx/app/forum/cs_comments_service:cached
       
+volumes:
+  credentials_node_modules:
+  discovery_node_modules:
+  ecommerce_node_modules:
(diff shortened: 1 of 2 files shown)


provision-lms.sh
```diff
@@ -13,7 +13,6 @@ for app in "${apps[@]}"; do
     docker-compose $DOCKER_COMPOSE_FILES up -d $app
 done
 
-docker-compose exec lms bash -c 'rm -f /edx/app/edxapp/edx-platform/.prereqs_cache/Node_prereqs.sha1 && rm -Rf /edx/app/edxapp/edx-platform/node_modules'
```
efischer19, line 16: I got a weird error here initially, so I just removed this line in hopes that it's no longer needed.

efischer19: `rm: cannot remove '/edx/app/edxapp/edx-platform/node_modules': Device or resource busy`

per the initial failing Travis run on this

efischer19:
Initial attempt had separate anonymous volumes for each of `lms` and `studio`, now trying with a single (hopefully shared) named volume.

The next step if this doesn't work will be to try adding `:delegated` to further hammer home the point that what *actually* matters is the container's view of things.

efischer19:
This has succeeded, with the following caveats:

- `ecommerce`: independently hit the r.js issue, will likely need its own `node_modules/` volume as well
- `discovery`: independently hit the r.js issue, will likely need its own `node_modules/` volume as well. Also currently has `fsevents` in its [package-lock](https://github.com/edx/course-discovery/blob/master/package-lock.json#L2059)
- 'credentials': does not currently hit the r.js issue, but I suspect that's because it's running an older version of npm. I propose giving it a `node_modules/` volume as well to prevent seeing it in the future.

arizzitano (approved):
Re `credentials` running a pinned, older version of npm -- we REALLY ought to fix that. Aside from that I think this is a good fix, and it addresses some of the problems around users modifying `node_modules` from the host. One potential problem stands out to me though: there's no ability to install new npm packages from the host. @efischer19 what would happen if someone added a new folder into `node_modules` in the host (say, via an erroneous npm install)? Would it mess up the container's `node_modules`? Do you think it's worthwhile to add a `make` command to install a new npm package into the container?

efischer19:
My opinion: not being able to install new packages from the host is a feature, not a bug. It will prevent issues like the recent `fsevents` install now that we have package-locks everywhere. I think it would be good for people to get into the habit of running `make <whatever>-shell` before making any sort of non-code changes to a repo to further the "consistent, replicable installs" idea.

I'm totally open to adding whatever make commands make people's lives easier though.

docker-compose-host.yml
```diff
@@ -4,21 +4,31 @@ services:
   credentials:
     volumes:
       - ${DEVSTACK_WORKSPACE}/credentials:/edx/app/credentials/credentials:cached
+      - credentials_node_modules:/edx/app/credentials/credentials/node_modules
```
macdiesel, line 7: It's my understanding that cached provides a significant speed increase. Is there a reason we wouldn't want to use it here as well?

efischer19: The caching options (cached, consistent, detached) only apply to bind-mounted host directories; these are named docker volumes that do not have a corresponding host directory.

macdiesel (approved):

macdiesel:
reopening to restart CI

PR #261, opened 2018-01-24
Simplify the instructions on how images are built

jibsheet:
Everything but credentials appears to use tools-edx-jenkins and
configuration:master now, especially after docker-devstack-idas was
merged.


jmbowman (approved):
👍 , thanks for catching this.  It looks like we're still using the "devstack" tag for the elasticsearch container, do you think that should be called out?

jibsheet:
Yeah, I noticed elasticsearch later, and it's not in tools-edx-jenkins.  I didn't honestly even know how the one we use is built, but I found it and added a note.  I suspect we actually want to clean out some of this, but that's a larger thing than fixing some of the docs to have correct pointers.

PR #266, opened 2018-02-01
Add devpi container to cache PyPI pkgs.

doctoryes:
https://openedx.atlassian.net/browse/PLAT-1912

PR is dependent on this companion configuration PR: https://github.com/edx/configuration/pull/4309

devpi is a standalone caching layer in front of PyPI. This PR:
- adds the container to the Open edX Docker devstack container suite
- adds a target to shell into the devpi container
- adds documentation about the devpi container and its function

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,7 +112,7 @@ restore:  ## Restore all data volumes from the host. WARNING: THIS WILL OVERWRIT
 	docker run --rm --volumes-from edx.devstack.mongo -v $$(pwd)/.dev/backups:/backup debian:jessie tar zxvf /backup/mongo.tar.gz
 	docker run --rm --volumes-from edx.devstack.elasticsearch -v $$(pwd)/.dev/backups:/backup debian:jessie tar zxvf /backup/elasticsearch.tar.gz
 
-# TODO: Create a devstack.sh file in the devpi container to activate the devpi venv/clear the entire cache.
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -98,6 +98,10 @@ restore:  ## Restore all data volumes from the host. WARNING: THIS WILL OVERWRIT
 	docker run --rm --volumes-from edx.devstack.mongo -v $$(pwd)/.dev/backups:/backup debian:jessie tar zxvf /backup/mongo.tar.gz
 	docker run --rm --volumes-from edx.devstack.elasticsearch -v $$(pwd)/.dev/backups:/backup debian:jessie tar zxvf /backup/elasticsearch.tar.gz
 
+# TODO: Create a devstack.sh file in the devpi container to activate the devpi venv/clear the entire cache.
```
bmedx: Nit: I think we don't want this to automatically clear the cache? I'd expect it to:
- activate the venv
- "devpi use" the default index
- "devpi login" the root user

Is that sensible?

doctoryes: I meant that slash to separate two options:
- shell-in (open)
- cache clear (cache-clear?)

Makes sense to do all you mention upon "open". I'll change the comment.

bmedx (approved):

macdiesel (commented):
🚢 🇮🇹

PR #267, opened 2018-02-01
[WIP] add volumes for persisting local edx videos

pwnage101:
This WIP PR comes from the edx-in-a-box/edx-for-inmates hackathon group.  It creates a local persistent video storage volume on the LMS and Studio containers.

Related:
* https://github.com/edx/edx-platform/pull/17365/
* https://github.com/edx/edx-val/pull/122


pwnage101:
The problem with this is that we don't need the volume for any other purpose besides edx-in-a-box, so it would be weird to merge this as-is.

regisb:
@pwnage101 is this still relevant? If not, can we close this? You should feel free to re-open it later, if necessary.

PR #269, opened 2018-02-02
First part of being able to start XQueue in docker devstack

jibsheet:
This primarily covers
* using the existing XQueue docker image
* adding rabbitmq
* adding XQueue to MySQL
* Have provisioning set up the rabbit users needed
* Make a shell for rabbit/xqueue
  Since it lacks the devstack.sh for now, just run bash
* Change to the devstack settings file

Coming later
* Start a second container for the consumer, before this, the xqueue
  docker container rain rsyslogd, gunicorn and the consumer.  This
  simplifies to closer to the other IDAs

### DIFF
--- a/docker-compose-xqueue.yml
+++ b/docker-compose-xqueue.yml
@@ -1,0 +1,17 @@
+version: "2.1"
+
+services:
+  rabbitmq:
+    image: rabbitmq:3.6.9
+    container_name: edx.devstack.rabbit
+
+  xqueue:
+    container_name: edx.devstack.xqueue
+    image: edxops/xqueue:latest
+    command: bash -c 'source /edx/app/xqueue/venvs/xqueue/bin/activate && while true; do SERVICE_VARIANT=xqueue python /edx/app/xqueue/xqueue/manage.py runserver 0.0.0.0:18040 --settings xqueue.devstack; sleep 2; done'
+    volumes:
+      - ${DEVSTACK_WORKSPACE}/xqueue:/edx/app/xqueue/xqueue:cached
+    depends_on: # even though we need mysql, we can't refer to it because it's started in the other compose file
+      - rabbitmq 
+    ports:
+      - 18040:18040
--- a/provision-xqueue.sh
+++ b/provision-xqueue.sh
@@ -1,0 +1,31 @@
+set -e
+set -o pipefail
+set -x
+
+# Bring up RabbitMQ and XQueue
+docker-compose $DOCKER_COMPOSE_FILES up -d rabbitmq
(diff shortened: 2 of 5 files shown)


jibsheet:
@macdiesel @jmbowman  Do either of you want to comment on this?  From some testing, it should basically never enter into a standard developer loop (except for make down which will also turn off xqueue).  Provisioning and starting xqueue is separate from our standard containers.  My intent is to keep it that way, and I'll be eventually ripping out Rabbit.

macdiesel (approved):

provision-xqueue.sh
```diff
@@ -0,0 +1,31 @@
...
+set -o pipefail
+set -x
+
+# Bring up RabbitMQ and XQueue
+docker-compose $DOCKER_COMPOSE_FILES up -d rabbitmq
+docker-compose $DOCKER_COMPOSE_FILES up -d xqueue
+
+# This works in case you provision xqueue without having other services up
+# Bring the database online.
+docker-compose up -d mysql
```
jmbowman, line 11: Do you need `$DOCKER_COMPOSE_FILES` here?

docker-compose-xqueue.yml
```diff
@@ -0,0 +1,17 @@
...
+
+services:
+  rabbitmq:
+    image: rabbitmq:3.6.9
+    container_name: edx.devstack.rabbit
+
+  xqueue:
+    container_name: edx.devstack.xqueue
+    image: edxops/xqueue:latest
+    command: bash -c 'source /edx/app/xqueue/venvs/xqueue/bin/activate && while true; do SERVICE_VARIANT=xqueue python /edx/app/xqueue/xqueue/manage.py runserver 0.0.0.0:18040 --settings xqueue.devstack; sleep 2; done'
```
jmbowman, line 11: The provisioning script uses `xqueue.aws_settings` instead; should they be consistent?

jmbowman (approved):
Just a couple of questions, otherwise looks good.

docker-compose-xqueue.yml
jibsheet: Yeah - I added the devstack.py later on than provisioning.  I'll switch to it and reprovision xqueue to confirm it

provision-xqueue.sh
jibsheet: I stole it from the standard provision.sh which doesn't (it assumes mysql will be in docker-compose.yml which is the default consumed by docker-compose unless we pass in the list of -f files).  I noticed that there were a bunch of places where we oscillate between calling docker <cmd> vs docker-compose <cmd> and sometimes use DOCKER_COMPOSE_FILES and don't.  I let it be for mysql rather than trying to clean it all up.

PR #274, opened 2018-02-21
Clone edx-themes and mount for whitelabel configuration.

doctoryes:
https://openedx.atlassian.net/browse/PLAT-1975
This PR adds code to clone edx-themes (separately from the usual cloning) and adds volume mount points for edx-themes in the appropriate containers.

### DIFF
--- a/repo.sh
+++ b/repo.sh
@@ -27,11 +27,19 @@ repos=(
     "https://github.com/edx/xqueue.git"
 )
 
+private_repos=(
+    # Needed to run whitelabel tests.
+    "https://github.com/edx/edx-themes.git"
+)
+
 name_pattern=".*edx/(.*).git"
 
-clone ()
+_clone ()
 {
-    for repo in ${repos[*]}
+    # for repo in ${repos[*]}
+    repos_to_clone=("$@")
+
+    for repo in "${repos_to_clone[@]}"
     do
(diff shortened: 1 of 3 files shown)


repo.sh
```diff
@@ -27,11 +27,19 @@ repos=(
     "https://github.com/edx/xqueue.git"
 )
 
+private_repos=(
+    # Needed to run whitelabel tests.
+    "https://github.com/edx/edx-themes.git"
```
doctoryes, line 32: Is it OK to name a private GitHub repo here?

coryleeio (approved):

repo.sh
bmedx: Seems ok to me, but I think we need further documentation about why this won't ever work for people outside of the org. Maybe in the Makefile comments and devstack README?

bmedx (approved):

repo.sh
doctoryes: Yeah - I think what I'll do is just split the make targets which clone these into a separate edx-specific Makefile. So then it'd be run with:
`make -f Makefile.edx dev.clone_whitelabel`

doctoryes:
@bmedx What do you think about the edx-specific Makefile I've added?

bmedx:
Looks good! Much clearer, I think.

PR #276, opened 2018-02-23
Add mgmt cmd script to setup whitelabel tests.

doctoryes:
https://openedx.atlassian.net/browse/PLAT-1893
Call each mgmt cmd against the appropriate container. Assumes that devstack has been started using the new `make -f Makefile.edx dev.up.e2e_wl_tests` target, which mounts edx-themes into the appropriate containers.

### DIFF
--- a/provision-whitelabel.sh
+++ b/provision-whitelabel.sh
@@ -1,0 +1,55 @@
+set -e
+set -o pipefail
+
+if [ -z "$DEVSTACK_WORKSPACE" ]; then
+    DEVSTACK_WORKSPACE=..
+elif [ ! -d "$DEVSTACK_WORKSPACE" ]; then
+    echo "Workspace directory $DEVSTACK_WORKSPACE doesn't exist"
+    exit 1
+fi
+
+# For colored status messages below.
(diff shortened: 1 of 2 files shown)


provision-whitelabel.sh
```diff
@@ -0,0 +1,55 @@
...
+wl_course_run="2018"
+wl_modulestore="split"
+wl_dns_name="wl-ci"
+wl_theme_path="${app_base_path}/edx-themes/edx-platform"
+
+#
+# Run the needed Django mgmt commands against each container.
+#
+echo "${GREEN}Creating sites and themes in LMS...${NC}"
+cmd="source ${edxapp_base_path}/edxapp_env && ${edxapp_base_path}/edx-platform/manage.py --settings=lms.envs.devstack_docker create_sites_and_configurations --dns-name ${wl_dns_name} --theme-path ${wl_theme_path}"
```
bmedx: This is all good, but once we have devstack specific config files in edx-themes we'll need to add another flag to point this script to those files instead of the stage ones.

bmedx (approved):

PR #277, opened 2018-02-27
Use edx-themes scripts for whitelabel tests and provisioning.

bmedx:


bmedx:
Superceded by the superior PR 278

PR #278, opened 2018-02-27
Move provision script to themes repo and add whitelabel run target.

doctoryes:
Use the appropriate file from edx-themes to provision and run the whitelabel tests. Also add test cleanup target.


bmedx (approved):
Nice! This is far better than my working version of these changes. 🥇

PR #283, opened 2018-03-21
Set COMPOSE_PROJECT_NAME var in provision scripts

thallada:
This sets the project name to a consistent value `"devstack"` regardless of the folder name during provisioning. This is a follow up to https://github.com/edx/devstack/pull/280 which set the variable within the Makefile.

I tested this by renaming my `devstack` folder to `devstack-test`, ran `sh provision-discovery.sh`, and then `docker volume list | grep devstacktest` which returned no results.

### DIFF
--- a/provision-credentials.sh
+++ b/provision-credentials.sh
@@ -3,6 +3,8 @@
 # NOTE (CCB): We do NOT call provision-ida because it expects a virtualenv.
 # The new images for Credentials do not use virtualenv.
 
+export COMPOSE_PROJECT_NAME='devstack'
+
 name=credentials
 port=18150
 
--- a/provision-discovery.sh
+++ b/provision-discovery.sh
@@ -1,4 +1,6 @@
 # Provisioning script for the discovery service
+export COMPOSE_PROJECT_NAME='devstack'
+
 ./provision-ida.sh discovery 18381
 
 docker-compose exec discovery bash -c 'rm -rf /edx/var/discovery/*'
--- a/provision-e2e.sh
+++ b/provision-e2e.sh
@@ -2,6 +2,8 @@ set -e
 set -o pipefail
 set -x
 
+export COMPOSE_PROJECT_NAME='devstack'
+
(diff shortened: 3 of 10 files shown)


robrap:
I'm not sure how @macdiesel wants to summarize, but just capturing his thoughts on the PR so it doesn't get lost in HipChat.

> [1:44 PM] Brian Beggs: I don't have a good answer as to why the .env file is in the gitignore. @JeremyBowman may have better context on that as he was on the project much longer than I.
> [1:45 PM] Brian Beggs: I searched through  the chat room history and at one point it was suggested that it be used for something else by George Song and there didn't seem to be any objections at the time.
> [1:46 PM] Brian Beggs: looks like it was even in a pull request at some point?
> [1:46 PM] Brian Beggs: https://github.com/edx/devstack/pull/108
> [1:46 PM] Brian Beggs: sorry that's .env.example
> [1:47 PM] Brian Beggs: I think there may have been a desire to let people have their own .env files locally to override whatever they wanted and to not have one checked in to the repo
> [1:51 PM] Brian Beggs: looking further at that PR that is when the .env was introduced to the .gitignore so I'm certain this was done not to override anyones local .env file they may have
> [1:52 PM] Brian Beggs: Could someone explain to me why we shouldn't just let this go?
> [1:56 PM] Brian Beggs: oh because of integration outside of the make files
> [1:57 PM] Brian Beggs: I'm a +1 on removing .env from the .gitignore and setting the COMPOSE_PROJECT_NAME there
> [1:59 PM] Brian Beggs: though one last question, should studio-frontend's compose items be inside of devstack instead of inside it's own repo?
> [1:59 PM] Brian Beggs: that would remove this confusion
> [1:59 PM] Brian Beggs: also having the Dockerfile in that repo strays from the current convention of having those in the configuration repository
> [2:00 PM] Brian Beggs: One might argue that studio-frontend should conform to our current standards
> [2:00 PM] Brian Beggs: sorry for the rollercoster of emotions here, just typing as I'm working through discovery on this.

PR #285, opened 2018-03-29
Two small fixes

nedbat:
- There was no mention of setting up file sharing properly.
- Setting Docker to 2.0GiB memory actually has a lower number than 2*2^30, so make the numbers match.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -231,7 +231,7 @@ build-courses: ## NOTE: marketing course creation is not available for those out
 	rm course-generator/tmp-config.json
 
 check-memory: ## Check if enough memory has been allocated to Docker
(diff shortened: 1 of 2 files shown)


nedbat:
@jmbowman @mduboseedx Thoughts?

jmbowman (approved):

nedbat:
@jmbowman any idea why travis is failing?

jmbowman:
@nedbat There are a few different issues with the devstack repo Travis builds right now, fixes are currently being coordinated in the Devstack HipChat channel.  Some of it is fallout from the Django 1.11 upgrade (management commands that didn't have any tests and don't work with 1.11, etc.)

PR #289, opened 2018-04-13
Add devpi dependency to lms and studio

jmbowman:
Both lms and studio now count on devpi being up when updating dependencies, but we hadn't captured this in the explicit container dependencies (which a problem during provisioning, when only a subset of the containers are launched).  It's not strictly needed (pip will eventually fall back to PyPI), but installing dependencies without it is slower and generates lots of warnings like this:

```
Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPConnection object at 0x7fd046336290>: Failed to establish a new connection: [Errno -2] Name or service not known',)': /root/pypi/+simple/edx-i18n-tools/
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -130,6 +130,7 @@ services:
     command: bash -c 'source /edx/app/edxapp/edxapp_env && while true; do python /edx/app/edxapp/edx-platform/manage.py lms runserver 0.0.0.0:18000 --settings devstack_docker; sleep 2; done'
     container_name: edx.devstack.lms
     depends_on:
+      - devpi
       - mysql
       - memcached
       - mongo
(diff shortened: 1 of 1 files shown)


michaelyoungstrom (approved):

estute:
I'm guessing the rationale for only using this for lms/studio is due to the fact that the platform has so many dependencies. Is the eventual plan to use this for all IDAs that are part of devstack?

estute (approved):

jmbowman:
Yes, we want to start using this for other IDAs also; there just hasn't been a ticket for doing that work yet.  It should help speed up re-installation of packages for any local development, and introduces some interesting options for local testing of package release candidates in other code that uses them.

PR #296, opened 2018-04-23
EDUCATOR-2632 | Add script to provision a retirement service account user

iloveagent57:
The devstack counterpart to https://github.com/edx/edx-platform/pull/18027

Creates a new retirement service account user and DOT application from `provision-lms.sh`.

### DIFF
--- a/provision-lms.sh
+++ b/provision-lms.sh
@@ -38,3 +38,6 @@ docker-compose exec lms bash -c '/edx/app/edx_ansible/venvs/edx_ansible/bin/ansi
 for app in "${apps[@]}"; do
     docker-compose exec $app bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && paver update_assets --settings devstack_docker'
 done
+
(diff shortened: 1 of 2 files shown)


ssemenova (approved):
I guess, ok, yes, this is better than my "add code to the SQL file" plan.

provision-lms.sh
```diff
@@ -38,3 +38,6 @@ docker-compose exec lms bash -c '/edx/app/edx_ansible/venvs/edx_ansible/bin/ansi
 for app in "${apps[@]}"; do
     docker-compose exec $app bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && paver update_assets --settings devstack_docker'
 done
+
+# Provision a retirement service account user
+./provision-retirement-user.sh retirement retirement_service
```
doctoryes: It wouldn't hurt to line this up with the name for the user in edx-internal, which is "retirement_service_worker".

doctoryes (approved):

PR #298, opened 2018-04-30
Add analytics pipeline containers

rao-abdul-mannan:
This PR will add analytics pipeline docker containers.
###### _First git clone this repo and then from root execute following commands._
clone other repos as well:
```make dev.clone```
##### Provision analytics pipeline
```make dev.provision.analytics_pipeline```

##### Access analytics pipeline shell
```make dev.up.analytics_pipeline```     # bring up the containers first
```make analytics-pipeline-shell```

##### stop/start analytics pipeline 
```make stop.analytics_pipeline```       # stop
```make dev.up.analytics_pipeline```    # start

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -229,7 +229,7 @@ mongo-shell: ## Run a shell on the mongo container
 
 ### analytics pipeline commands
 
-dev.provision.analytics_pipeline: | check-memory dev.provision.analytics_pipeline.run stop stop.analytics_pipeline
+dev.provision.analytics_pipeline: | check-memory dev.provision.analytics_pipeline.run stop.analytics_pipeline stop
 
 dev.provision.analytics_pipeline.run:
 	DOCKER_COMPOSE_FILES="-f docker-compose.yml -f docker-compose-host.yml -f docker-compose-analytics-pipeline.yml" ./provision-analytics-pipeline.sh
--- a/Makefile
+++ b/Makefile
@@ -235,7 +235,7 @@ dev.provision.analytics_pipeline.run:
 	DOCKER_COMPOSE_FILES="-f docker-compose.yml -f docker-compose-host.yml -f docker-compose-analytics-pipeline.yml" ./provision-analytics-pipeline.sh
 
 analytics-pipeline-shell: ## Run a shell on the Analytics pipeline container
-	docker exec -it edx.devstack.test.analytics_pipeline env TERM=$(TERM) /edx/app/analytics_pipeline/devstack.sh open
+	docker exec -it edx.devstack.analytics_pipeline env TERM=$(TERM) /edx/app/analytics_pipeline/devstack.sh open
 
 dev.up.analytics_pipeline: | check-memory ## Bring up analytics_pipeline
 	docker-compose -f docker-compose.yml -f docker-compose-analytics-pipeline.yml -f docker-compose-host.yml up -d analyticspipeline
--- a/Makefile
+++ b/Makefile
@@ -244,7 +244,7 @@ pull.analytics_pipeline: ## Update Analytics pipeline Docker images
 	docker-compose -f docker-compose-analytics-pipeline.yml pull --parallel
 
 stop.analytics_pipeline:
-	docker-compose -f docker-compose.yml -f docker-compose-analytics-pipeline.yml stop
+	docker-compose -f docker-compose-analytics-pipeline.yml stop
 
 # Provisions studio, ecommerce, and marketing with course(s) in test-course.json
 # Modify test-course.json before running this make target to generate a custom course
--- a/Makefile
+++ b/Makefile
@@ -85,7 +85,7 @@ stop.xqueue:
 
 down: ## Remove all service containers and networks
 	(test -d .docker-sync && docker-sync clean) || true ## Ignore failure here
-	docker-compose -f docker-compose.yml -f docker-compose-watchers.yml -f docker-compose-xqueue.yml down
+	docker-compose -f docker-compose.yml -f docker-compose-watchers.yml -f docker-compose-xqueue.yml -f docker-compose-analytics-pipeline.yml down
 
 destroy: ## Remove all devstack-related containers, networks, and volumes
 	./destroy.sh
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: Makefile [hunk 1]
```diff
@@ -227,6 +227,25 @@ mysql-shell-edxapp: ## Run a mysql shell on the edxapp database
...
+	docker exec -it edx.devstack.test.analytics_pipeline env TERM=$(TERM) /edx/app/analytics_pipeline/devstack.sh open
+
+dev.up.analytics_pipeline: | check-memory ## Bring up analytics_pipeline
+	docker-compose -f docker-compose.yml -f docker-compose-analytics-pipeline.yml -f docker-compose-host.yml up -d analyticspipeline
+
+pull.analytics_pipeline: ## Update Analytics pipeline Docker images
+	docker-compose -f docker-compose-analytics-pipeline.yml pull --parallel
+
+stop.analytics_pipeline:
+	docker-compose -f docker-compose.yml -f docker-compose-analytics-pipeline.yml stop
```
jmbowman, line 247: This and `stop.xqueue` should both be added to `stop.all`.  In that case, this target should only reference the analytics pipeline docker-compose file; that would let developers shut down the analytics pipeline containers while keeping the rest of devstack running.

docker-compose-analytics-pipeline.yml
```diff
@@ -0,0 +1,121 @@
+version: "2.1"
+
+services:
+  namenode:
+    image: edxops/analytics_pipeline_hadoop_namenode:latest
```
jmbowman, line 5: I hadn't noticed before, but looking at the Dockerfiles in the configuration repo, these 4 hadoop images seem to be just minor variations on the same base image.  How difficult would it be to create a single `edxops/hadoop` image which could be used for all 4 hadoop containers here, with the differences configured in this file?

Makefile
rao-abdul-mannan: Right, will update in next commit.

docker-compose-analytics-pipeline.yml
rao-abdul-mannan: I don't think we can as some  require volume, different directory & even CMD scripts.

macdiesel: I would prefer not to revisit this now but a ticket to go back and take a look the next time we need to upgrade these things. Could you put in this ticket @rao-abdul-mannan?

docker-compose-analytics-pipeline.yml
```diff
@@ -0,0 +1,121 @@
...
+    container_name: edx.devstack.analytics_pipeline.namenode
+    hostname: namenode
+    environment:
+      - CLUSTER_NAME=devstack
+    ports:
+      - 127.0.0.1:50070:50070
+    command: ["/run.sh"]
+    volumes:
+      - namenode_data:/hadoop/dfs/name
```
macdiesel, line 14: @rao-abdul-mannan What are your thoughts about making this and the datanote filesystem mounts? This would allow the data to persist across provisions. Thoughts?

provision-analytics-pipeline.sh
```diff
@@ -0,0 +1,62 @@
...
+
+# Analytics pipeline has dependency on lms but we only need its db schema & not full lms. So we'll just load their db
+# schemas as part of analytics pipeline provisioning. If there is a need of a full fledge LMS, then provision lms
+# by following their documentation.
+if [[ ! -z "`docker exec -i edx.devstack.mysql mysql -uroot -se "SELECT SCHEMA_NAME FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME='edxapp'" 2>&1`" ]];
+then
+  echo -e "${GREEN}LMS DB exists, skipping lms schema load.${NC}"
+else
+  echo -e "${GREEN}LMS DB not found, load the schema.${NC}"
+  docker exec -i edx.devstack.mysql mysql -uroot --max_allowed_packet=1073741824 mysql < provision.sql
```
macdiesel: What do you think about maybe changing this to run paver update_db --settings=docker-devstack instead? This would guarantee that the user of the system gets the most up to date LMS database.

macdiesel (approved):
I'm good with this as it is, though I do have a few questions. It's ok to merge this, add some tickets and iterate on a few of these points.

docker-compose-analytics-pipeline.yml
rao-abdul-mannan: This will mount the datanode's data to the external volume and yes it will persist across provisions unless deleted

docker-compose-analytics-pipeline.yml
rao-abdul-mannan: @macdiesel Sure.

provision-analytics-pipeline.sh
rao-abdul-mannan: @macdiesel This would require provisioning LMS and then running those commands. I just tried them out and it was successful, but doing this change would mean copying almost half the commands from provision-lms.sh file.

rao-abdul-mannan: Let me commit those changes.

provision-analytics-pipeline.sh
```diff
@@ -0,0 +1,69 @@
...
+then
+  echo -e "${GREEN}LMS DB exists, skipping lms schema load.${NC}"
+else
+  # Provisioning lms db with migrations. Update above comment if these are approved.
+  echo -e "${GREEN}LMS DB not found, load the schema.${NC}"
+  docker exec -i edx.devstack.mysql mysql -uroot --max_allowed_packet=1073741824 mysql < provision.sql
+  docker exec -i edx.devstack.mysql mysql -uroot --max_allowed_packet=1073741824 edxapp < edxapp.sql
+  docker-compose $DOCKER_COMPOSE_FILES up -d lms studio
+  docker-compose exec lms bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && NO_PYTHON_UNINSTALL=1 paver install_prereqs'
+  #Installing prereqs crashes the process
```
brianhw, line 41: nit: space after '#'

provision-analytics-pipeline.sh
```diff
@@ -0,0 +1,69 @@
...
+done
+
+# Analytics pipeline has dependency on lms but we only need its db schema & not full lms. So we'll just load their db
+# schemas as part of analytics pipeline provisioning. If there is a need of a full fledge LMS, then provision lms
+# by following their documentation.
+if [[ ! -z "`docker exec -i edx.devstack.mysql mysql -uroot -se "SELECT SCHEMA_NAME FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME='edxapp'" 2>&1`" ]];
+then
+  echo -e "${GREEN}LMS DB exists, skipping lms schema load.${NC}"
+else
+  # Provisioning lms db with migrations. Update above comment if these are approved.
```
brianhw, line 35: I don't understand the comment here.   What should be approved here, and what should the update be if they are approved?

provision-analytics-pipeline.sh
```diff
@@ -0,0 +1,69 @@
...
+
+# materialize hadoop directory structure
+echo -e "${GREEN}Initializing Hadoop directory structure...${NC}"
+
+until curl http://namenode:50070/jmx?qry=Hadoop:service=NameNode,name=NameNodeStatus|grep -q 'active'; do
+  printf "Waiting for namenode!"
+  sleep 5
+done
+
+sleep 10 # for datanode & other services to activate
```
brianhw, line 62: You had an explicit wait for namenode above.   Are there comparable ways to identify if the other services are active?

brianhw (commented):
Looks good to me -- a few non-blocking nits.

PR #299, opened 2018-05-07
Add demo program when provisioning

mikix:
Add some scripts to provision a demo program.

It will automatically do so when doing a normal provision. Or you can manually do it for your current devstack by running ./programs/provision.sh.

Errors are not fatal, to match the rest of the provisioning scripts, and because configuring one IDA can still be done if an earlier one fails. We print errors, so they can still be investigated.

I chose to do this work as tiny Python scripts rather than tiny management commands, because it seemed so use-case-specific. It does mean that refactoring in the IDAs might break these scripts... But I think that's an acceptable risk.

I've also **added a cache-programs step during dev.up**. If we don't do this, everytime you start a container, you'd have to manually cache the programs to have it work at all in the LMS. It adds ~7s to every dev.up time... But guh I hate caching manually. Looking for opinions on the tradeoff.

I chose to ignore bundling, since that wasn't our immediate concern. Which meant I could ignore ecommerce. I also ignored marketing, since we don't care about interactions there right now.

But this is a good base for future-us to work on to add such support.

Once this lands, we should ping devs about it, as an FYI.

https://openedx.atlassian.net/browse/LEARNER-5141

### DIFF
--- a/programs/discovery.py
+++ b/programs/discovery.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python3
+# Run like so:
+# ./manage.py shell -c "`cat discovery.py`"
+
+import urllib.request as request
+from course_discovery.apps.core.models import Partner
+from course_discovery.apps.course_metadata.models import (
+    Course, CourseRun, Organization, Program, ProgramType, SeatType
+)
+
+DEMO_IMAGE_URL = 'http://edx.devstack.lms:18000/asset-v1:edX+DemoX+Demo_Course+type@asset+block@images_course_image.jpg'
+
+
+# Make sure micromasters type exists
+micromasters, _ = ProgramType.objects.get_or_create(slug='micromasters', defaults={'name': 'MicroMasters'})
+micromasters.applicable_seat_types.add(
+    SeatType.objects.get_or_create(slug='verified', defaults={'name': 'Verified'})[0],
+    SeatType.objects.get_or_create(slug='professional', defaults={'name': 'Professional'})[0],
+    SeatType.objects.get_or_create(slug='credit', defaults={'name': 'Credit'})[0],
+)
+
+# Add a demo program
+edx_partner = Partner.objects.get(short_code='edx')  # created during normal provision
+program, _ = Program.objects.update_or_create(
+    marketing_slug='demo-program',
+    defaults={
+        'title': 'edX Demonstration Program',
+        'type': micromasters,
+        'status': 'active',
+        'partner': edx_partner,
+        'overview': 'A demo program for testing.',
+        'total_hours_of_effort': 4,
+        'min_hours_effort_per_week': 1,
+        'max_hours_effort_per_week': 4,
+        'one_click_purchase_enabled': True,
+        'card_image_url': DEMO_IMAGE_URL,
+    },
+)
+
+# Now, after an ID has been created, connect the program to other models
+
+course = Course.objects.get(key='edX+DemoX')
+program.courses = [course]
+
+try:
+    # This run causes run-time exceptions, because it uses old style key.
+    deprecated_run = CourseRun.objects.get(key='edX/DemoX/Demo_Course')
+    program.excluded_course_runs = [deprecated_run]
+except CourseRun.DoesNotExist:
+    # This key only seems to be in some existing devstacks, don't worry if it doesn't exist
+    pass
+
+edx_org = Organization.objects.get(key='edX')
+program.authoring_organizations = [edx_org]
+program.credit_backing_organizations = [edx_org]
+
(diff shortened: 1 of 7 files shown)


programs/discovery.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python3
+# Run like so:
+# ./manage.py shell -c "`cat discovery.py`"
+
+import urllib.request as request
+from course_discovery.apps.core.models import Partner
+from course_discovery.apps.course_metadata.models import Course, CourseRun, Organization, Program, ProgramType
+
+# Add a demo program
+micromasters = ProgramType.objects.get(slug='micromasters')
```
bderusha: Can we be sure that this exists?  Should this be a `get_or_create` instead?

programs/discovery.py
```diff
@@ -0,0 +1,42 @@
...
+edx_partner = Partner.objects.get(short_code='edx')
```
bderusha: Same question ^

programs/lms.py
```diff
@@ -0,0 +1,34 @@
...
+    'internal_api_url': 'https://example.com/api',  # required but unused
+    'service_username': 'discovery_worker',
+})
+
+# Tell LMS about discovery
+SiteConfiguration.objects.update_or_create(
+    site=Site.objects.get(domain='example.com'),
+    defaults={
+        'enabled': True,
+        'values': {'COURSE_CATALOG_API_URL': 'http://edx.devstack.discovery:18381/api/v1/'},
```
bderusha: NIT: Can we pull this url (or maybe just the base url) into a variable at the top of the file so it's easier to find/update if the devstack config changes?

bderusha (commented):
Save for a few minor comments this code looks fine to me.

I would like to have some conversation about just how special this case is.  Ideally we do want the ability to write seed data for a service to be the norm for any service or new feature so that it is easy/easier to develop with.

Should this be the pattern we use for all data seeding going forward?  What would this look like if we did not consider it a special case?

programs/discovery.py
mikix: Ah great catch. I hadn't remembered creating those manually, but I guess I did. Updated code to use get_or_create.

programs/discovery.py
mikix: The edx partner is explicitly made by the provision scripts before ours runs. (See provision-discovery.sh)

programs/lms.py
mikix: Done.

mikix:
To your broader point about seeding in general...

This is not a special case really. When I said use-case-specific, I meant more like "setting up the specific demo course that edx.org developers prefer to test against in devstack" felt like too specific a task for a management command.

Now, we could add more general management commands and drive those with specific arguments from devstack. That often makes sense, like for create_user or whatever. (In this PR's case, I would have made "create_program" for discovery and "configure_programs" for LMS maybe? But that already starts getting weirdly specific with configure_programs. Toggling a few settings feels like too small a task for a management command to me, but maybe my taste is off there?)

We have a couple "drivers" today that set up parts of the devstack: ansible commands in the configuration repo and devstack scripts.  But both mostly drive management commands.

So I did go a bit off on my own with the "piping django commands to an IDA" strategy. But it felt appropriately script-y for this task to me. I think the same strategy would make sense for future similar devstack demo setup for new features.

If you feel odd about the django-code route, I'd be happy to move these bits into management commands in their respective IDAs.

programs/discovery.py
```diff
@@ -0,0 +1,53 @@
...
+from course_discovery.apps.core.models import Partner
+from course_discovery.apps.course_metadata.models import (
+    Course, CourseRun, Organization, Program, ProgramType, SeatType
+)
+
+DEMO_IMAGE_URL = 'http://edx.devstack.lms:18000/asset-v1:edX+DemoX+Demo_Course+type@asset+block@images_course_image.jpg'
+
+
+# Make sure micromasters type exists
+micromasters = ProgramType.objects.get_or_create(slug='micromasters', defaults={'name': 'MicroMasters'})[0]
```
bderusha: NIT: python style note.   Prefer `micromasters, __ = ProgramType.objects.get_or_create(slug='micromasters', defaults={'name': 'MicroMasters'})`

bderusha (approved):
To capture our offline conversation.  This is a reasonable approach moving forward as seed data should be owned by the provisioning scripts and the bash escaping stuff helps us write cleaner python scripts.  👍

programs/discovery.py
mikix: Done. The next few lines where I use get_or_create for passed-in arguments can't really do the same trick, but I fixed the ones where we can.

PR #301, opened 2018-05-09
Fix demo program provisioning on BSD/Macs

mikix:
And some minor other fixes, commented below.

### DIFF
--- a/course-generator/build-course-json.sh
+++ b/course-generator/build-course-json.sh
@@ -1,4 +1,4 @@
-# Script to build course configurations in proper json format that can be passed into provision-courses.sh
+# Script to build course configurations in proper json format that can be passed into create-courses.sh
 # See test-course.json for a master list of course configurations
 # TODO: Link documentation for course-generator-tool
 # USAGE: ./build-course-json.sh [course-config-file]
--- a/programs/provision.sh
+++ b/programs/provision.sh
@@ -28,7 +28,7 @@ notice() {
 reset_color() {
     SWITCH="\033["
     NORMAL="${SWITCH}0m"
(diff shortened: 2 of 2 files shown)


course-generator/build-course-json.sh
```diff
@@ -1,4 +1,4 @@
-# Script to build course configurations in proper json format that can be passed into provision-courses.sh
+# Script to build course configurations in proper json format that can be passed into create-courses.sh
```
mikix, line 1: Minor typo I saw when doing some other work today. It's referencing the script in the same folder by the wrong name.

programs/provision.sh
```diff
@@ -28,7 +28,7 @@ notice() {
 reset_color() {
     SWITCH="\033["
     NORMAL="${SWITCH}0m"
-    echo "${NORMAL}"
+    echo -n "${NORMAL}"
```
mikix, line 31: Don't add an extra newline in our output when going back to the normal color.

programs/provision.sh
```diff
@@ -50,7 +50,7 @@ provision_ida() {
     cmd=${2:-"shell"}
 
     # Escape double quotes and backticks from the Python
-    PROGRAM_SCRIPT="$(sed 's/\("\|`\)/\\\1/g' < "$BASEDIR/$service.py")"
+    PROGRAM_SCRIPT="$(sed -E 's/("|`)/\\\1/g' < "$BASEDIR/$service.py")"
```
mikix, line 53: Uses a regex format that is standard across POSIX, rather than the default format, which apparently can differ.

bderusha (approved):
🎸

PR #302, opened 2018-05-10
Move over to pip-tools to comply with OEP-18

michaelyoungstrom:
Move this repo over to pip-tools to comply with OEP-18.


post-pip-compile.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# Remove any cruft from a requirements file generated by pip-tools which we don't want to keep
+
+function show_help {
+    echo "Usage: post-pip-compile.sh file ..."
+    echo "Remove any cruft left behind by pip-compile in the given requirements file(s)."
+    echo ""
+    echo "Updates the instructions for re-generating each requirements file, and removes"
+    echo "\"-e\" prefixes which were added to GitHub URLs only so that pip-compile could"
+    echo "process them correctly."
```
jmbowman: It looks like you can omit the bit about "-e" prefixes here.

jmbowman (approved):
One help text suggestion, otherwise 👍

PR #303, opened 2018-05-14
setup analytics pipeline travis build tests

rao-abdul-mannan:
This PR adds support for 
- Running analytics pipeline acceptance tests using their docker images/containers , as part of travis build. 
- Optimizes analytics docker devstack provisioning.

##### Analytics pipeline tests finish within 30mins which is still lower than lms provisioning & test which is around 34mins.

### DIFF
--- a/.travis/run.sh
+++ b/.travis/run.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+set -e
+set -x
+
+if [[ $DEVSTACK == 'lms' ]]; then
+    make dev.provision
+    make dev.up
+    sleep 60  # LMS needs like 60 seconds to come up
+    make healthchecks
+    make validate-lms-volume
+    # Disable e2e-tests until either:
+    # * tests are less flaky
+    # * We have a way to test the infrastructure for testing but ignore the test results.
+    # See PLAT-1712
+    # - make e2e-tests
+    make up-marketing-detached
+fi
+
+if [[ $DEVSTACK == 'analytics_pipeline' ]]; then
+    make dev.provision.analytics_pipeline
+    make dev.up.analytics_pipeline
+    sleep 30 # hadoop services need some time to be fully functional and out of safemode
+    make analytics-pipeline-devstack-test
+fi
--- a/Makefile
+++ b/Makefile
@@ -251,11 +251,17 @@ dev.up.analytics_pipeline: | check-memory ## Bring up analytics_pipeline
 	docker-compose -f docker-compose.yml -f docker-compose-analytics-pipeline.yml -f docker-compose-host.yml up -d analyticspipeline
 
 pull.analytics_pipeline: ## Update Analytics pipeline Docker images
-	docker-compose -f docker-compose-analytics-pipeline.yml pull --parallel
+	docker-compose -f docker-compose.yml -f docker-compose-analytics-pipeline.yml pull --parallel
+
+analytics-pipeline-devstack-test:
+	docker exec -u hadoop -i edx.devstack.analytics_pipeline bash -c 'sudo chown -R hadoop:hadoop /edx/app/analytics_pipeline && source /edx/app/hadoop/.bashrc && make develop-local && make docker-test-acceptance-local ONLY_TESTS=edx.analytics.tasks.tests.acceptance.test_internal_reporting_database && make docker-test-acceptance-local ONLY_TESTS=edx.analytics.tasks.tests.acceptance.test_user_activity'
 
 stop.analytics_pipeline:
 	docker-compose -f docker-compose.yml -f docker-compose-analytics-pipeline.yml stop
 	docker-compose up -d mysql      ## restart mysql as other containers need it
+
+hadoop-application-logs-%: ## Hadoop logs by application Id
+	docker exec -it edx.devstack.analytics_pipeline.nodemanager yarn logs -applicationId $*
 
 # Provisions studio, ecommerce, and marketing with course(s) in test-course.json
 # Modify test-course.json before running this make target to generate a custom course
--- a/provision-analytics-pipeline.sh
+++ b/provision-analytics-pipeline.sh
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+
 set -e
 set -o pipefail
 set -x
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -243,7 +243,11 @@ dev.up.analytics_pipeline: | check-memory ## Bring up analytics_pipeline
 	docker-compose -f docker-compose.yml -f docker-compose-analytics-pipeline.yml -f docker-compose-host.yml up -d analyticspipeline
 
 pull.analytics_pipeline: ## Update Analytics pipeline Docker images
-	docker-compose -f docker-compose-analytics-pipeline.yml pull --parallel
+	docker-compose  -f docker-compose.yml -f docker-compose-analytics-pipeline.yml pull --include-deps analyticspipeline
+
+analytics-pipeline-devstack-test:
+	docker exec -it edx.devstack.analytics_pipeline make docker-test-acceptance-local ONLY_TESTS=edx.analytics.tasks.tests.acceptance.test_hive
+	docker exec -it edx.devstack.analytics_pipeline make docker-test-acceptance-local ONLY_TESTS=edx.analytics.tasks.tests.acceptance.test_user_activity
```
brianhw: I think test_user_activity is good enough, as it tests hive and mysql.  That should be enough to cover most things that devstack developers would be likely to break.  I think test_hive is redundant, and if we want to replace it with a warehouse-related test, I would do that next.  While either test_internal_reporting_country or test_internal_reporting_user_activity would be quick (~3 minutes) and would test Vertica loading, I think that test_internal_reporting_database would be better, in terms of testing sqoop as well as vertica, and times at 4 minutes.

rao-abdul-mannan: @brianhw Thanks, i'll use `test_internal_reporting_database` in place of `test_hive`.

.travis/run.sh
```diff
@@ -0,0 +1,25 @@
...
+    # * We have a way to test the infrastructure for testing but ignore the test results.
+    # See PLAT-1712
+    # - make e2e-tests
+    make up-marketing-detached
+fi
+
+if [[ $DEVSTACK == 'analytics_pipeline' ]]; then
+    make dev.provision.analytics_pipeline
+    make dev.up.analytics_pipeline
+    sleep 30 # hadoop services need sometime to be fully functional and out of safemode
```
brianhw: nit:   sometime => some time

provision-analytics-pipeline.sh
```diff
@@ -25,17 +27,22 @@ do
   sleep 1
 done
 
+# In the event of a fresh MySQL container, wait a few seconds for the server to restart
```
brianhw: nit:  period at the end.

provision-analytics-pipeline.sh
```diff
@@ -25,17 +27,22 @@ do
...
+# This can be removed once https://github.com/docker-library/mysql/issues/245 is resolved.
+sleep 20
+
 # Analytics pipeline has dependency on lms but we only need its db schema & not full lms. So we'll just load their db
 # schemas as part of analytics pipeline provisioning. If there is a need of a full fledge LMS, then provision lms
```
brianhw: nit:  full fledge => fully fledged.

provision-analytics-pipeline.sh
```diff
@@ -25,17 +27,22 @@ do
...
 then
   echo -e "${GREEN}LMS DB exists, skipping lms schema load.${NC}"
 else
-  # Provisioning lms db with migrations. Update above comment if these are approved.
-  echo -e "${GREEN}LMS DB not found, load the schema.${NC}"
-  docker exec -i edx.devstack.mysql mysql -uroot --max_allowed_packet=1073741824 mysql < provision.sql
-  docker exec -i edx.devstack.mysql mysql -uroot --max_allowed_packet=1073741824 edxapp < edxapp.sql
+  echo -e "${GREEN}LMS DB not found, provisioning lms schema.${NC}"
+  # docker exec -i edx.devstack.mysql mysql -uroot -e "set global max_allowed_packet=1073741824;set global wait_timeout = 99999;SET global net_read_timeout = 99999;"
+  # without above query, subsequent queries will fail.
```
brianhw: But the above query is not commented out.   Should this be removed now?

brianhw (approved):
Looks good, and runs successfully on Travis!   A few optional nits, but one question about a commented-out query.   But otherwise good to go. 👍

provision-analytics-pipeline.sh
rao-abdul-mannan: @brianhw Yeah will do that.

Makefile
```diff
@@ -251,12 +251,18 @@ dev.up.analytics_pipeline: | check-memory ## Bring up analytics_pipeline
 	docker-compose -f docker-compose.yml -f docker-compose-analytics-pipeline.yml -f docker-compose-host.yml up -d analyticspipeline
 
 pull.analytics_pipeline: ## Update Analytics pipeline Docker images
-	docker-compose -f docker-compose-analytics-pipeline.yml pull --parallel
+	docker-compose  -f docker-compose.yml -f docker-compose-analytics-pipeline.yml pull --parallel
+
+analytics-pipeline-devstack-test:
+	docker exec -u hadoop -i edx.devstack.analytics_pipeline bash -c 'sudo chown -R hadoop:hadoop /edx/app/analytics_pipeline && source /edx/app/hadoop/.bashrc && make develop-local && make docker-test-acceptance-local ONLY_TESTS=edx.analytics.tasks.tests.acceptance.test_internal_reporting_database && make docker-test-acceptance-local ONLY_TESTS=edx.analytics.tasks.tests.acceptance.test_user_activity'
```
jmbowman: Doesn't need to be done right now, but the script used here is long enough that it should probably be an actual script file in the container.

Makefile
```diff
@@ -251,12 +251,18 @@ dev.up.analytics_pipeline: | check-memory ## Bring up analytics_pipeline
...
 
 stop.analytics_pipeline:
 	docker-compose -f docker-compose.yml -f docker-compose-analytics-pipeline.yml stop
 	docker-compose up -d mysql      ## restart mysql as other containers need it
 
+hadoop-applicationLogs-%: ## Hadoop logs by application Id
```
jmbowman: I think this is the only target in the Makefile that uses camel case; could it just be `hadoop-logs-%` instead?

.travis/run.sh
```diff
@@ -0,0 +1,25 @@
...
+    # - make e2e-tests
+    make up-marketing-detached
+fi
+
+if [[ $DEVSTACK == 'analytics_pipeline' ]]; then
+    make dev.provision.analytics_pipeline
+    make dev.up.analytics_pipeline
+    sleep 30 # hadoop services need some time to be fully functional and out of safemode
+    make analytics-pipeline-devstack-test
+fi
```
jmbowman: Nit: missing a newline at the end of the file

Makefile
```diff
@@ -251,12 +251,18 @@ dev.up.analytics_pipeline: | check-memory ## Bring up analytics_pipeline
 	docker-compose -f docker-compose.yml -f docker-compose-analytics-pipeline.yml -f docker-compose-host.yml up -d analyticspipeline
 
 pull.analytics_pipeline: ## Update Analytics pipeline Docker images
-	docker-compose -f docker-compose-analytics-pipeline.yml pull --parallel
+	docker-compose  -f docker-compose.yml -f docker-compose-analytics-pipeline.yml pull --parallel
```
jmbowman: Nit: there's an extra space after `docker-compose`

provision-analytics-pipeline.sh
```diff
@@ -54,15 +60,15 @@ docker-compose $DOCKER_COMPOSE_FILES exec analyticspipeline bash -c '/edx/app/ha
 # materialize hadoop directory structure
 echo -e "${GREEN}Initializing Hadoop directory structure...${NC}"
 
-until curl http://namenode:50070/jmx?qry=Hadoop:service=NameNode,name=NameNodeStatus|grep -q 'active'; do
+until curl http://127.0.0.1:50070/jmx?qry=Hadoop:service=NameNode,name=NameNodeStatus|grep -q 'active'; do
```
jmbowman, line 63: `edx.devstack.analytics_pipeline.namenode` should also work as the hostname.

jmbowman (approved):
Looks pretty good!  Found a few minor things that might be worth tweaking, but otherwise seems ready to go.

Makefile
rao-abdul-mannan: @jmbowman This logs are for application ( by their id ) & not actually the container logs, so I did name it that way. `hadoop-application-logs` seems a little longer but I think its more clear than just `hadoop-logs`.

Makefile
rao-abdul-mannan: @jmbowman Only reason its been that long, is because of the mounted directory being owned by root & we actually run every command under different user `hadoop` which fails without changing ownership.

rao-abdul-mannan: And I have only seen this happen with travis, provisioning analyticstack on local machine has no such issues.

provision-analytics-pipeline.sh
rao-abdul-mannan: I did try it but it doesn't work. Only works if you add hostname entry in /etc/hosts file

PR #305, opened 2018-05-16
Since a requimentes folder exist, make needs to know to still run 'ma…

feanil:
…ke requirements'


jibsheet (approved):

feanil:
The errors are unrelated to this change so merging this.

PR #306, opened 2018-05-17
Create a fresh/clean devstack with one script

fysheets:
The idea for this stems because @efischer19 created some scripts that we can use on our team to blow away and recreate our devstacks with one command. This process takes about 30-35 minutes, which is awesome. Eric also runs this weekly (dare I say daily) to ensure his local setup does encounter errors due to staleness.

I took this idea and tried to leverage it with a new hire recently who needed devstack set up, but does not necessarily need to know how it works / the details of what commands to run and when. I also wanted see if we can use this sort of script creation for our internship program. I see this as a way of helping folks get onboard faster with a running local setup, then be able to leverage the extensive Makefile/devstack README to learn the details as needed.

This PR is mostly to get the conversation started (thanks @doctoryes) and I realize it will need some changes, especially since we can't clean/remove the Git workspace repo if the script that's running is living within that repo. Ultimately I think this sort of functionality could be useful for devs in general, but I know getting folks on board with that is a bigger hurdle. If we can come up with a solution that will at least help our internship group here in a couple of weeks and then iterate on it from there, that'd be swell.

### DIFF
--- a/new-devstack.sh
+++ b/new-devstack.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+read -p "This will delete ALL docker and local Git data. Would you like to proceed? [y/n] " -r
+if [[ $REPLY =~ ^[Yy]$ ]]
+then
+    # 1. Ensure all docker containers, images, and volumes are cleaned up
+    # Stop all containers
+    docker stop $(docker ps -a -q)
+    # Delete all containers
+    docker rm $(docker ps -a -q)
+    # Delete all images
+    docker rmi $(docker images -q)
+    # Delete all volumes
+    docker volume rm -f $(docker volume ls -q)
+
+    # 2. Remove and remake workspace directory
+    rm -rf ~/workspace
+    mkdir ~/workspace/
+    cd ~/workspace
+
+    # 3. Rebuild the system
+    git clone https://github.com/edx/devstack.git # git@github.com:edx/devstack.git
+    cd devstack
+    make requirements
+    make dev.clone
+    make dev.provision
+    make dev.up
+
+    # 4. Possible way to handle optional/additional repo-cloning? i.e. our team always needs studio-frontend
+fi


new-devstack.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+read -p "This will delete ALL docker and local Git data. Would you like to proceed? [y/n] " -r
+if [[ $REPLY =~ ^[Yy]$ ]]
+then
+    # 1. Ensure all docker containers, images, and volumes are cleaned up
```
fysheets, line 6: On the quick read of `destroy.sh` for the `make destroy` command, I wasn't immediately sure of the differences on the methods of destruction here vs there. What is below is simply what I have from the scripts that I've been running locally.

new-devstack.sh
```diff
@@ -0,0 +1,30 @@
...
+
+    # 3. Rebuild the system
+    git clone https://github.com/edx/devstack.git # git@github.com:edx/devstack.git
+    cd devstack
+    make requirements
+    make dev.clone
+    make dev.provision
+    make dev.up
+
+    # 4. Possible way to handle optional/additional repo-cloning? i.e. our team always needs studio-frontend
```
fysheets, line 29: In my mind, this is a nice to have, but would probably be useful as we start breaking apart the monolith (assuming that leads to more repos).

new-devstack.sh
```diff
@@ -0,0 +1,30 @@
...
+    # Delete all volumes
+    docker volume rm -f $(docker volume ls -q)
+
+    # 2. Remove and remake workspace directory
+    rm -rf ~/workspace
+    mkdir ~/workspace/
+    cd ~/workspace
+
+    # 3. Rebuild the system
+    git clone https://github.com/edx/devstack.git # git@github.com:edx/devstack.git
```
fysheets, line 22: The `dev.clone` command is using https as well, so I'm wondering if there was discussion at some point on using https for these scripts rather than ssh for a reason? I know that I prefer to use ssh when cloning, but we realized last week/today that 1. if you're brand new you need to first create your SSH key and link it to Git before running the script and 2. the developer onboarding docs don't mention SSH currently.

efischer19 (commented):
My only concern is that it's kind of weird to have a script that destroys the devstack directory live *inside* that directory.

Would it be better to make this an addition to the README, something like "In your terminal: run these 4 docker commands, delete the directory that contains devstack, recreate it, then clone devstack and run these 4 make commands"?

new-devstack.sh
```diff
@@ -0,0 +1,30 @@
...
+    docker rm $(docker ps -a -q)
+    # Delete all images
+    docker rmi $(docker images -q)
+    # Delete all volumes
+    docker volume rm -f $(docker volume ls -q)
+
+    # 2. Remove and remake workspace directory
+    rm -rf ~/workspace
+    mkdir ~/workspace/
+    cd ~/workspace
```
doctoryes, line 19: To eliminate concerns about running from the same dir which this script is deleting, perhaps the script could be changed to something like:
```
mkdir ~/workspace_new
cd ~/workspace_new
# Do all the cloning/making...
mv ~/workspace ~/workspace_old
mv ~/workspace_new ~/workspace
rm -rf ~/workspace_old
```
*OR* - there are git commands that can be used, like `git reset --hard` or `git clean`, to clean the existing repos without deleting them all.

new-devstack.sh
doctoryes: It makes sense to use `https` links here and most everywhere in devstack. (There's some edx-only code that clones private GH repos that'd use SSH links, but that's mostly irrelevant.) We don't want the existence of a GitHub account and SSH keypair to be a barrier to getting the platform running for brand-new developers. Sure - they'd eventually need to GH account to submit changes but there's no need for that barrier upfront.

new-devstack.sh
doctoryes: An extensible way of doing this would be to check for the existence of a file. If it exists locally, it would contain additional repo links that would be cloned along with the cloning happening above. And the filename could be added to `.gitignore`.

new-devstack.sh
jmbowman: We recently switched Jenkins from using https to ssh for git clones because it turned out to be far more reliable.  Not sure if that's as much of an issue on a local machine which isn't doing as much in parallel, but something to consider (especially if you're planning to clone several repos over and over again).

fysheets: Cool -- good to know on both sides, I think to start having it use https would make sense and keep it aligned with our other commands currently, but it could be an update down the road to have everything use ssh instead.

new-devstack.sh
fysheets: I can see pros/cons to both of these options. I think cleaning the repos without deleting them is nice because there isn't a need to re-download each of the repos again, but I like the completeness of starting with a completely clean state. @jmbowman do you have any opinions on this piece of the script as a whole? Eric had initially had this split into multiple scripts (one for deleting everything and a separate one for cloning/creating) and it may make sense to have 2 or 3 make commands possibly for this -- one that will do both those tasks, but then one that would just do the creation fresh? If it's your first time setting up, you don't need to do the wipe clean piece of the script. By doing it that way, it can probably also leverage the `make destroy` command for the clean up step as well?

fysheets:
Thanks everyone for the comments. I think I have a better understanding for what we can do for right now, and some possible ways to extend this in the future. If there are any more comments, you guys know where to find me :)

PR #309, opened 2018-05-22
Simplify shell rule to access all containers

rao-abdul-mannan:
This PR simplifies existing shell target to allow shell access to all containers ( containers with devstack.sh have explicit shell rules )

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,23 +132,19 @@ restore:  ## Restore all data volumes from the host. WARNING: THIS WILL OVERWRIT
 	docker run --rm --volumes-from edx.devstack.mongo -v $$(pwd)/.dev/backups:/backup debian:jessie tar zxvf /backup/mongo.tar.gz
 	docker run --rm --volumes-from edx.devstack.elasticsearch -v $$(pwd)/.dev/backups:/backup debian:jessie tar zxvf /backup/elasticsearch.tar.gz
 
-# TODO: Create a devstack.sh file in the devpi container to activate the devpi venv within the shell.
-devpi-shell: ## Run a shell on the devpi container
-	docker exec -it edx.devstack.devpi env TERM=$(TERM) /bin/bash
-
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -140,6 +140,11 @@ devpi-shell: ## Run a shell on the devpi container
 %-shell: ## Run a shell on the specified service container
 	docker exec -it edx.devstack.$* env TERM=$(TERM) /edx/app/$*/devstack.sh open
 
+# Not every container has directory structure like /edx/app nor do they have devstack.sh. Without these, above shell rule does not work.
+# Which is why this rule has been added to allow bash shell access to every container.
+%-bash: ## Run a bash shell on the specified service container
+	docker exec -it edx.devstack.$* /bin/bash
```
jmbowman: I was going to suggest setting `TERM` as done below for credentials, but I think that was actually a workaround for a bug which has been fixed for a while now: https://github.com/moby/moby/issues/9299#issuecomment-246687815 .

We don't have that many IDAs in devstack, and we have some targets like `credentials-shell` below for which this command would work fine.  How about adding explicit `-shell` targets for any missing IDAs which have `devstack.sh`, and then make this `%-shell` and remove the old `%-shell`, `credentials-shell`, `devpi-shell` and `forum-shell`?  I'd like to keep the targets consistent, instead of having to remember if a particular service uses `-shell` or `-bash`.

rao-abdul-mannan:
Build failing due to a recent change in ecommerce repo. Author is notified of this issue.
```
File "/edx/app/ecommerce/ecommerce/ecommerce/core/management/commands/create_demo_data.py", line 67, in handle
```
I will rerun the build once it is fixed.

jmbowman:
I think you'll need to add explicit `discovery-shell` and `ecommerce-shell` targets because those were using the old `%-shell` implementation, but other than that 👍

macdiesel (approved):

PR #310, opened 2018-05-22
Update readme section about requirements

BenjiLee:
Even though docker-compose comes with MacOS, a user should still go through the pip install `make requirement` steps for other non docker-compose packages.

@jmyatt


jmyatt:
LGTM

jmyatt (approved):

PR #313, opened 2018-05-28
Support for offline-installable snapshots

jmbowman:
Scripts for creating and restoring reasonably comprehensive snapshots of devstack which can be installed and used without a network connection.  Created to avoid a bandwidth crisis during the Open edX Conference workshops by distributing devstack on USB flash drives.

This doesn't yet include any of the optional containers like analytics, etc.  The resulting output is currently about 7 GB, and requires an additional 25 GB of space on the target system once installed.

This also lays the groundwork for other options like downloading pre-provisioned data volumes along with the latest Docker images.  All of the compressed volumes combined are about 194 MB, significantly smaller than any of the current Open edX service images.

### DIFF
--- a/scripts/extract_snapshot_linux.sh
+++ b/scripts/extract_snapshot_linux.sh
@@ -1,0 +1,26 @@
+#!/bin/sh
+
+# Installs Open edX devstack on Linux from a local snapshot of repository,
+# image, and volume tarballs.
+
+set -e
+
+# Extract all of the Open edX source code repositories needed to run devstack
+for tarball in repositories/*.tar.gz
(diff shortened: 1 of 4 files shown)


jmbowman:
I've been getting a lot of requests for hosting the output of this somewhere, and if we do that we should probably also automate periodic builds.  It's about 7 GB, do we have anyplace appropriate for that to live?

feanil (approved):
We could host in on s3 with a CDN in front of it.  We should coordinate with @nedbat  and maybe make a 'files.openedx.org' location where we host this.

PR #314, opened 2018-06-01
Expose spark UI port to analytics pipeline container

rao-abdul-mannan:
This PR fixes issue with spark UI port exposed on wrong container.


macdiesel (commented):
You don't want the web UI exposed on the master node?

rao-abdul-mannan:
@macdiesel No, reason is that we launch spark jobs from pipeline container and we explicitly tell it to connect itself with spark ( master + worker ) standalone cluster so UI is launched on pipeline container.

macdiesel (approved):

brianhw (approved):
Seems to work for me.   👍

PR #315, opened 2018-06-05
Add notes service

pwnage101:
* I also needed to change provision-ida.sh to accept a client name, since
  the notes service bindings in the LMS are hard-coded to use a client
  name which doesn't follow the naming scheme of other IDAs.  This led
  to a tiny refactor  or other provision scripts.
* The decision to use port 18120 was based on the fact that that port
  number is already assumed to be notes in a few other places.

PLAT-2170

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -296,6 +296,8 @@ meant to be user-facing, the "homepage" may be the API root.
 +---------------------+-------------------------------------+
 | LMS                 | http://localhost:18000/             |
 +---------------------+-------------------------------------+
+| Notes/edx-notes-api | http://localhost:18120/api/v1/      |
++---------------------+-------------------------------------+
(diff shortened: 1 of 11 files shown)


pwnage101:
at the time of writing, this is still blocked on a few things:

* edx-notes-api needs a make target for "migrate"
* ~~we still need to actually build and push the docker container for notes.~~  Actually, there's already a docker container which is used for running CI in travis, which I'll try to re-use.

pwnage101:
FYI, the "pr" travis tests failed due to networking, but there wasn't much of a difference between my branch and master, so I think the "push" test should be sufficient.

bmedx (approved):
👍

doctoryes (approved):
FYI @edx/devops - @pwnage101 is including the notes IDA in devstack!

PR #317, opened 2018-06-05
WIP: Alpha Windows support

jdmulloy:


jmbowman:
As it stands, this PR unblocks getting pretty far in provisioning devstack on Windows, without seeming to break anything on Mac or Linux.  Provisioning doesn't complete yet on AppVeyor, but that's probably a RAM configuration issue; I'll create a ticket to fix that in a separate PR.  I've added a test script that covers what does work already, since the default AppVeyor configuration fails without one (it expects to find an MSVC build file if you don't provide a build script, even when there's no `appveyor.yml` file present).

appveyor.yml
```diff
@@ -0,0 +1,41 @@
...
+branches:
+  only:
+    - master
+
+environment:
+  DEVSTACK_WORKSPACE: x:/devstack
+  SHALLOW_CLONE: 1
+
+install:
+- curl -fsSL -o make-4.2.1-without-guile-w32-bin.zip https://sourceforge.net/projects/ezwinports/files/make-4.2.1-without-guile-w32-bin.zip/download
```
michaelyoungstrom, line 12: How do you feel about possibly storing this in S3 instead of downloading through sourceforge?

Makefile
```diff
@@ -102,7 +117,7 @@ down: ## Remove all service containers and networks
 	docker-compose -f docker-compose.yml -f docker-compose-watchers.yml -f docker-compose-xqueue.yml -f docker-compose-analytics-pipeline.yml down
```
estute, line 117: Judging from the other examples, I imagine you may need to run this as a bash command. There are a few others, but turning off the devstack is more important than some of the lesser used make targets. However, I understand if you just want to get this merged in for the sake of enabling CI.

estute (approved):
In general, LGTM. I am not familiar with appveyor or some of these windows commands, but I can see what is going on and it doesn't look like this will break devstacks for the other OSes.

Makefile
jmbowman: Yeah, I think I'll hold off on switching more of these until I can verify on an actual Windows machine that the change is needed.

appveyor.yml
jmbowman: I'm hoping to try a change soon that could eliminate the need to download it at all, so not sure it's worth the effort of switching - [TE-2763](https://openedx.atlassian.net/browse/TE-2763)

PR #320, opened 2018-06-13
Link local src volumes to IDAs

BenjiLee:
Gives the other containers access to the local src directory.


robrap (approved):
I see no problem with this.  Might want @jmbowman to just give his perspective.

robrap (approved):
Still approve.

PR #322, opened 2018-06-14
LEARNER-5358 Adds records-help-url config parameter

spencerhance:
https://github.com/edx/credentials/pull/400 from [LEARNER-5358](https://openedx.atlassian.net/browse/LEARNER-5358) adds in an optional configuration parameter for the records-help-url.  Since it is left empty by default, I think it makes sense to add a value to the provision script.

This can't be merged until https://github.com/edx/credentials/pull/400 has been merged since create_or_update_site will not accept unknown args


mikix (commented):
Looks fine, but why not use the real URL? (This is the one Grant provided in https://github.com/edx/edx-documentation/pull/1704#issuecomment-396697629)

https://support.edx.org/hc/en-us/sections/360001216693-Learner-records

spencerhance:
@mikix Good point, is there something to be said about the dev environment using dummy data though?  The page that it also seeds the certificates URL with looks like a dummy faq page.

I don't really have a preference though 😄

mikix (approved):
Yeah, you won me over. Especially since that page is not confusing in the sense that it doesn't have any content at all, so it's not like we're linking to an FAQ about Duck Dynasty or whatever.

Love it as is.

PR #324, opened 2018-07-02
Switch base image used for credentials repo.

dianakhuang:
Switch from using `devstack-slim` to `latest`.

### DIFF
--- a/provision-credentials.sh
+++ b/provision-credentials.sh
@@ -9,20 +9,20 @@ port=18150
 docker-compose $DOCKER_COMPOSE_FILES up -d $name
 
 echo -e "${GREEN}Installing requirements for ${name}...${NC}"
-docker exec -t edx.devstack.${name}  bash -c 'make requirements' -- "$name"
+docker exec -t edx.devstack.${name}  bash -c 'source /edx/app/credentials/credentials_env && cd /edx/app/credentials/credentials && make requirements && make production-requirements' -- "$name"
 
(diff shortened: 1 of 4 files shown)


dianakhuang:
@jmbowman thanks for all the help! I've added the fixes for those two issues. Let me know if there's anyone else who needs to review this.

jmbowman:
The latest commit I see is still from yesterday; did you forget to push a new one?

dianakhuang:
Yes, I did. 🤦‍♀️ Thanks for catching that.

provision-credentials.sh
```diff
@@ -9,20 +9,20 @@ port=18150
 docker-compose $DOCKER_COMPOSE_FILES up -d $name
 
 echo -e "${GREEN}Installing requirements for ${name}...${NC}"
-docker exec -t edx.devstack.${name}  bash -c 'make requirements' -- "$name"
+docker exec -t edx.devstack.${name}  bash -c 'source /edx/app/credentials/credentials_env && cd /edx/app/credentials/ && make requirements && make production-requirements' -- "$name"
```
jmbowman: Should these commands be using `/edx/app/credentials/credentials/` as the `cd` destination instead?

And I think we missed creating a ticket to update credentials' requirements handling for OEP-18, because at the time it wasn't tagged as a service to be included in the Hawthorn release.  Implementing that should eliminate the need for two requirements targets here (among other benefits).

jmbowman (approved):

PR #325, opened 2018-07-09
TE-2639 Support for named release images

jmbowman:
Use [docker-compose variable substitution](https://docs.docker.com/compose/compose-file/#variable-substitution) to more easily support using devstack Docker images for named Open edX releases.

We'll want to hold off on merging this until ``edx-e2e-tests`` has been tagged for release and we have a full set of ``hawthorn.master`` images so we can test it properly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -351,14 +351,17 @@ How do I build images?
 
 There are `Docker CI Jenkins jobs`_ on tools-edx-jenkins that build and push new
 Docker images to DockerHub on code changes to either the configuration repository or the IDA's codebase. These images
-are tagged ``latest``. Images that require tags other than ``latest`` are built and pushed by hand (see NOTES below).
+are tagged according to the branch from which they were built (see NOTES below).
 If you want to build the images on your own, the Dockerfiles are available in the ``edx/configuration`` repo.
 
 NOTES:
 
-1. edxapp and IDAs use the ``latest`` tag since their configuration changes have been merged to master branch of
-   ``edx/configuration``.
-2. The elasticsearch used in devstack is built using elasticsearch-devstack/Dockerfile and the ``devstack`` tag.
+1. edxapp and IDAs use the ``latest`` tag for configuration changes which have been merged to master branch of
+   their repository and ``edx/configuration``.
+2. Images for a named Open edX release are built from the corresponding branch
+   of each repository and tagged appropriately, for example ``hawthorn.master``
+   or ``hawthorn.rc1``.
+3. The elasticsearch used in devstack is built using elasticsearch-devstack/Dockerfile and the ``devstack`` tag.
 
 BUILD COMMANDS:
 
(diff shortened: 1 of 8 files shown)


README.rst
```diff
@@ -351,14 +351,16 @@ How do I build images?
...
 NOTES:
 
-1. edxapp and IDAs use the ``latest`` tag since their configuration changes have been merged to master branch of
-   ``edx/configuration``.
-2. The elasticsearch used in devstack is built using elasticsearch-devstack/Dockerfile and the ``devstack`` tag.
+1. edxapp and IDAs use the ``latest`` tag for configuration changes which have been merged to master branch of
+   their repository and ``edx/configuration``.
+2. Images for a named Open edX release are built from the corresponding branch
+   of each repository and tagged appropriately, for example ``hawthorn.master``
+   or ``hawthorn.rc1``.
```
nedbat, line 363: For devstack, people should probably always be on a ".master" branch.  But for their own development, they will need to be on their own fork, right?  Do we have a way to support that?

README.rst
```diff
@@ -385,6 +387,21 @@ For example, if you wanted to build tag ``release-2017-03-03`` for the
 E-Commerce Service, you would modify ``ECOMMERCE_VERSION`` in
 ``docker/build/ecommerce/ansible_overrides.yml``.
 
+How do I run the images for a named Open edX release?
+-----------------------------------------------------
+
+1. Set the OPENEDX_RELEASE environment variable to the appropriate image tag;
+   "hawthorn.master", "zebrawood.rc1", etc.
```
nedbat: "Note that unlike a server install, OPENEDX_RELEASE should not have the 'open-release/' prefix."

README.rst
```diff
@@ -385,6 +387,21 @@ For example, if you wanted to build tag ``release-2017-03-03`` for the
...
+2. Check out the correct branch (such as ``open-release/hawthorn.master``) in
+   the local checkout of each service repository.  ``make dev.checkout``
+   should handle this for you once you've set the environment variable above.
```
nedbat: Can we say this as "Use `make dev.checkout` to check out the correct branch in the local checkout of each service repository."  ?  Is there ever a reason they'd want to not use dev.checkout?

nedbat: Also, "once you've set the OPENEDX_RELEASE environment variable above."

README.rst
```diff
@@ -385,6 +387,21 @@ For example, if you wanted to build tag ``release-2017-03-03`` for the
...
+3. ``make pull`` to get the correct images.
+
+All ``make`` targes and ``docker-compose`` calls should now use the correct
```
nedbat: "targes"

docker-compose.yml
```diff
@@ -24,7 +24,7 @@ services:
 
   elasticsearch:
     container_name: edx.devstack.elasticsearch
-    image: edxops/elasticsearch:devstack
+    image: edxops/elasticsearch:${OPENEDX_RELEASE:-latest}
```
nedbat: This had been "devstack", but now defaults to "latest"

nedbat (changes requested):

README.rst
jmbowman: Most changes can still use the pre-built images for the branch they were derived from.  If they make more extensive changed to dependencies or such, they can build their own images using the instructions above.

nedbat: Oh, right.  I keep getting confused between pulling images, and pulling code.

nedbat (approved):

PR #328, opened 2018-07-24
Add python setup.py install to lms and studio commands

iloveagent57:
So we don't have to do this manually anymore when dealing with django app plugin changes.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -127,7 +127,7 @@ services:
       - "18130:18130"
 
   lms:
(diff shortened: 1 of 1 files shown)


nasthagiri (approved):
Thanks!

macdiesel (approved):

regisb:
Hi @iloveagent57! Is this change still necessary? If yes, can you please describe which issue this is solving? If not, we should close this.

PR #331, opened 2018-08-10
Go to a useful directory when entering marketing shell.

iloveagent57:
Previously, `make marketing-shell` would send you to some useless directory inside the marketing container.  Now it goes to the repo root.


schenedx (approved):
👍

PR #351, opened 2018-11-26
Always delete all the .pyc files when changing the code

nedbat:
In the last week, I have stumbled over this twice (moving between master and Hawthorn).  Clean up the .pyc files automatically so you will really be running the code you just pulled.

Are there other places we should do this?

### DIFF
--- a/repo.sh
+++ b/repo.sh
@@ -107,6 +107,7 @@ _checkout_and_update_branch ()
         git fetch origin ${OPENEDX_GIT_BRANCH}:${OPENEDX_GIT_BRANCH}
         git checkout ${OPENEDX_GIT_BRANCH}
     fi
+    find . -name '*.pyc' -not -path './.git/*' -delete 
 }
 
 clone ()


nedbat:
@jmbowman @doctoryes Opinions?

repo.sh
```diff
@@ -107,6 +107,7 @@ _checkout_and_update_branch ()
         git fetch origin ${OPENEDX_GIT_BRANCH}:${OPENEDX_GIT_BRANCH}
         git checkout ${OPENEDX_GIT_BRANCH}
     fi
+    find . -name '*.pyc' -not -path './.git/*' -delete
```
jmbowman, line 110: Should this use `-prune` instead for efficiency?  Something like `find . -name '*.pyc' -path ./.git -prune -o -delete`, I think.

jmbowman (commented):
My first instinct was to run `make clean` instead, but I guess that also clobbers other things that we may not want removed implicitly.

repo.sh
nedbat: I used the line from https://nedbatchelder.com/blog/201505/be_careful_deleting_files_around_git.html.  That blog post says, "find doesn’t like -prune and -delete together."  Maybe I didn't do it right?

jmbowman: Ah, ok; missed the relevant part of the man page under `-delete`.  Somebody found [a workaround using `xargs`](https://www.linuxquestions.org/questions/linux-general-1/find-with-prune-and-delete-902604/#post4470568), but not sure if it's worth switching.

jmbowman (approved):

PR #353, opened 2018-12-05
Give every container a meaningful hostname

nedbat:


nedbat:
@jmbowman I wasn't sure if CONTAINER_NAME is the best name, but couldn't think of a better one.

jmbowman (approved):
So far I'm not seeing a better approach.  @coryleeio , can you think of a better way to figure out which container you're inside from within it (or any reasons why allowing this is a terrible idea)?

coryleeio:
@jmbowman @nedbat 

There is a --hostname arg for docker run:
docker run -it --rm --name=hosty --hostname=hosty ubuntu:latest /bin/bash
> $@hosty:/# 

and it looks like there is a hostname counterpart in compose that you can use, I would probably use that instead if it doesn't break anything.
https://docs.docker.com/compose/compose-file/#domainname-hostname-ipc-mac_address-privileged-read_only-shm_size-stdin_open-tty-user-working_dir

I think you can just add hostname: thing to each thing in the compose file instead of the ENV

coryleeio:
eg:
container_name: edx.devstack.elasticsearch
hostname: edx.devstack.elasticsearch

nedbat:
Genius! Don't know why I didn't look into how to give them a hostname.

I've made it reversed from @coryleeio's suggestion:  "lms.devstack.edx".  That way the short form is "lms", and the default prompt is: "root@lms:/edx/app/edxapp/edx-platform#", which is just what we want.

docker-compose.yml
```diff
@@ -138,7 +148,7 @@ services:
     stdin_open: true
     tty: true
     environment:
-      BOK_CHOY_HOSTNAME: edx.devstack.lms
+      BOK_CHOY_hostname: lms.devstack.edx
```
jmbowman: There are several places in edx-platform that reference this by the previous name and value (in order to point browsers at the correct container for bok-choy tests); seems like the name should stay all caps and the value may need to be reversed in edx-platform where specified.

nedbat: Oops, didn't mean to change that...!

PR #378, opened 2018-12-28
TE-2791 XQueue and snapshot fixes

jmbowman:
* Make XQueue install development dependencies during provisioning (mainly to get `pyinotify` for more efficient file change polling)
* Added help strings to a few `make` targets so they'll actually appear in `make help`
* Fix creation of offline devstack installers to work again after a change in how the images are specified in `docker-compose.yml`

### DIFF
--- a/provision-xqueue.sh
+++ b/provision-xqueue.sh
@@ -18,6 +18,8 @@ do
 done
 
 docker exec -i edx.devstack.mysql mysql -uroot mysql < provision-xqueue.sql
+# Update dependencies
+docker-compose $DOCKER_COMPOSE_FILES exec xqueue bash -c 'source /edx/app/xqueue/xqueue_env && cd /edx/app/xqueue/xqueue && make requirements'
 # Run migrations
 docker-compose $DOCKER_COMPOSE_FILES exec xqueue bash -c 'source /edx/app/xqueue/xqueue_env && cd /edx/app/xqueue/xqueue && python manage.py migrate'
 # Add users that graders use to fetch data, there's one default user in Ansible which is part of our settings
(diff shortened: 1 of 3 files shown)


scripts/snapshot.py
```diff
@@ -73,6 +73,7 @@ def process_compose_file(filename, output_dir):
     for service_name in services:
         service = services[service_name]
         image = service['image']
+        image = re.sub(r'\$.*', 'latest', image)
```
estute, line 76: ^ what's going on here? Do you just want to hardcode it to always use `latest`?

jmbowman: The script as originally written was choking on the new syntax to determine the image tag from an environment variable if present.  I could have put some more sophisticated logic in here to parse the environment variable name and use its value if set, but it didn't seem that we'd want to create offline installers for non-master versions of devstack often enough to be worth the additional code complexity.

provision-xqueue.sh
```diff
@@ -18,6 +18,8 @@ do
 done
 
 docker exec -i edx.devstack.mysql mysql -uroot mysql < provision-xqueue.sql
+# Update dependencies
+docker-compose $DOCKER_COMPOSE_FILES exec xqueue bash -c 'source /edx/app/xqueue/xqueue_env && cd /edx/app/xqueue/xqueue && make requirements'
```
michaelyoungstrom, line 22: These 3 "docker-compose...." lines are all so similar, can/should we combine them into 1 line?

michaelyoungstrom (approved):

provision-xqueue.sh
jmbowman: We could, although that might introduce a fair bit of distance in the console from each command being run and its corresponding output (since the commands would all be batched together at the top).  The repetition isn't great, but I hesitate to change anything that might make debugging problems less obvious.

PR #379, opened 2019-01-07
Add gradebook to devstack

zacharis278:
Followup PR incoming for the gradebook project.  This will start the container, but the server won't be accessible without those changes since the port will move from 1991 to 1994.

Gradebook changes: https://github.com/edx/gradebook/pull/73

### DIFF
--- a/docker-compose-host.yml
+++ b/docker-compose-host.yml
@@ -33,9 +33,14 @@ services:
   forum:
     volumes:
       - ${DEVSTACK_WORKSPACE}/cs_comments_service:/edx/app/forum/cs_comments_service:cached
+  gradebook:
+    volumes:
+      - ${DEVSTACK_WORKSPACE}/gradebook:/edx/app/gradebook:cached
+      - gradebook_node_modules:/edx/app/gradebook/node_modules
       
 volumes:
   credentials_node_modules:
   discovery_node_modules:
   ecommerce_node_modules:
   edxapp_node_modules:
+  gradebook_node_modules:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -231,6 +231,16 @@ services:
     volumes:
       - devpi_data:/data
 
+  gradebook:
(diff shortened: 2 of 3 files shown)


docker-compose.yml
```diff
@@ -231,6 +231,16 @@ services:
     volumes:
       - devpi_data:/data
 
+  gradebook:
+    command: bash -c 'npm install && npm run start'
+    working_dir: '/edx/app/gradebook'
+    container_name: edx.devstack.gradebook
+    image: node:10
```
schenedx, line 238: Are we always going to use this base image?

schenedx (commented):
Are we going to create our own docker image for gradebook? Or a node v10 docker image is good enough?

docker-compose.yml
zacharis278: yeah after having a conversation with @coryleeio we decided this doesn't really justify adding another image to the configuration repo.  Everything we had in the original can be configured via docker-compose

schenedx: Great. We don't have to maintain our own image, which is lower overhead for us.

docker-compose-host.yml
```diff
@@ -33,9 +33,14 @@ services:
...
+    volumes:
+      - ${DEVSTACK_WORKSPACE}/gradebook:/edx/app/gradebook:cached
+      - gradebook_node_modules:/edx/app/gradebook/node_modules
       
 volumes:
   credentials_node_modules:
   discovery_node_modules:
   ecommerce_node_modules:
   edxapp_node_modules:
+  gradebook_node_modules:
```
schenedx: #NIT new line please

docker-compose.yml
```diff
@@ -231,6 +231,16 @@ services:
...
+    ports:
+      - "1994:1994"
```
schenedx, line 240: Please reference the gradebook repo change for this new port.

schenedx (approved):
👍 Good work

PR #385, opened 2019-02-11
Add new SSO/backend service client id/secret to provision.

doctoryes:
The `create_or_update_site` mgmt command in ecommerce was changed to allow the storage of two separate DOT applications in the oauth_settings of SiteConfiguration. This PR changes the provisioning command to match that change in a backward-compatible way.


pwnage101 (commented):
do we need to do something here about the LMS public url?

doctoryes:
@pwnage101 Right! I added it.

pwnage101 (approved):

PR #390, opened 2019-02-28
Make repo.sh error if it finds a non-empty but non-git directory when…

inventhouse:
… cloning

I was getting a less-obvious error when running `make dev.provision` so I didn't realize that it never got to any of the provisioning.  If there's some behavior other than throwing an error that would be preferable here, let me know.

### DIFF
--- a/repo.sh
+++ b/repo.sh
@@ -83,6 +83,10 @@ _clone ()
         # If a directory exists and it is nonempty, assume the repo has been checked out
         # and only make sure it's on the required branch
         if [ -d "$name" -a -n "$(ls -A "$name" 2>/dev/null)" ]; then
+            if [ ! -d "$name/.git" ]; then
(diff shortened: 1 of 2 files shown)


macdiesel (approved):

repo.sh
```diff
@@ -83,6 +83,10 @@ _clone ()
         # If a directory exists and it is nonempty, assume the repo has been checked out
         # and only make sure it's on the required branch
         if [ -d "$name" -a -n "$(ls -A "$name" 2>/dev/null)" ]; then
+            if [ ! -d "$name/.git" ]; then
+              printf "ERROR: [%s] exists but is not a git repo.\n" $name
```
nedbat, line 87: Can we try to keep the indentation consistent?

inventhouse: D'oh!  I was too aggressive about merging this; addressing the indentation in https://github.com/edx/devstack/pull/391

PR #391, opened 2019-03-01
Fix indentation of new .git-check

inventhouse:
Too aggressive on my previous merge, fixing the indentation issue called out in https://github.com/edx/devstack/pull/390


nedbat (approved):
Thanks! Sorry for being nit-picky.  There's a .editorconfig file in this repo that should help get these things right if your editor supports it.

PR #392, opened 2019-03-13
Fix Registrar requirements recipe

kdmccormick:
This PR fixes two issues with `make registrar-requirements`:
* The recipe was not activating registrar_env
* The recipe was also installing `production-requirements` for devstack development. With plain old `pip` this wouldn't be a problem, but `pip-sync` makes the installation profile exactly match the requirements file. So, when `production-requirements` is made, all non-production packages (e.g., `pytest`) are uninstalled.

I also piggy-backed two unrelated Registrar tweaks:
* `make registrar-clone` now clones Registrar into a sibling of the `devstack` directory instead of into a child of it.
* Removed `registrar-static` and `registrar-update-db`, which are redundant with the generic recipes in `Makefile`

@edx/masters-neem

### DIFF
--- a/registrar.mk
+++ b/registrar.mk
@@ -3,13 +3,13 @@ help-registrar: ## Display this help message
 	@perl -nle'print $& if m{^[\.a-zA-Z_-]+:.*?## .*$$}' $(MAKEFILE_LIST) | grep registrar | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m  %-25s\033[0m %s\n", $$1, $$2}'
 
 registrar-clone:  ## Clone registrar repository
-	git clone https://github.com/edx/registrar
+	git clone https://github.com/edx/registrar ../registrar
 
 registrar-pull:  ## Pulls latest version of all Docker images including Registrar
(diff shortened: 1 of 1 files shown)


schenedx (approved):
👍

registrar.mk
```diff
@@ -29,10 +29,13 @@ up-registrar-sync:  ## Bring up all services (including the registrar site) with
 clean-registrar-sync:   ## Remove the docker-sync containers for all services (including the registrar site)
 	docker-sync-stack clean -c docker-sync-registrar.yml
 
+registrar-attach:  ## Attach to registrar docker container
```
zacharis278: I don't think we need this since the generic version produces the same output.
https://github.com/edx/devstack/blob/master/Makefile#L187

zacharis278 (changes requested):

registrar.mk
kdmccormick: Great point. Might as well remove `registrar-static` and `registrar-update-db` too.

zacharis278 (approved):

PR #394, opened 2019-03-28
ARCH-603: add required scope user_id for oauth+sso

robrap:
DOT applications that use the oAuth+SSO flow with the EdXOAuth2 backend
from auth-backends will require the user_id scope in an oauth_dispatch
ApplicationAccess. This addition will request this scope.

ARCH-603


douglashall (approved):

robrap:
@douglashall: I decided to switch the scopes from `email,profile,user_id` to `user_id`, to make the non-default scope more obvious.  I'll let the defaults handle themselves.  Seems to work on devstack.  Make sense?

robrap:
FYI: @iloveagent57

PR #399, opened 2019-04-25
Add Jenkinsfile to build snapshot installers

jmbowman:
Add a Jenkins pipeline for building offline installers for devstack snapshots, to use at conferences and meetup events.  The generated directory hierarchy of tarballs and scripts can be downloaded in one click from the job artifacts.


michaelyoungstrom (approved):
Not sure why the travis build had trouble pulling images, but your code change looks good to me.

estute (approved):

PR #402, opened 2019-05-01
Move Registrar into main docker-compose files and Makefile

kdmccormick:
@edx/masters-neem 

Having Registrar in its own set of docker-compose/docker-sync files is confusing, messy, and unnecessary. With the new `dev.up.<service>` Makefile target, one can start only a service and its dependencies, so developers who don't want to deal with the overhead of starting the Registrar container no longer have to. (Yes, they still have to pull down the image and provision it, but the slowness of that is a separate issue).

This PR removes all Registrar-specific YAML files as well as registrar.mk, and merges them into the corresponding YAML file / Makefile. It also adds some information to README.rst on using `dev.up.<service>`

~Additionally, this PR has a commit that fixes `edxapp.sql`. Currently, the database file references an app name (`django_comment_common`) that was made invalid by https://github.com/edx/edx-platform/pull/20355. This PR updates the app name to `discussion_common` (FYI @nasthagiri )~ (Said edx-platform change was reverted)

TODO after merge: update Registrar README.


README.rst
```diff
@@ -668,6 +680,9 @@ or a manual Docker command to bring down the container:
 
    docker kill $(docker ps -a -q --filter="name=edx.devstack.<container name>")
 
+Alternatively, some terminals allow detachment from a running container with the
```
iloveagent57, line 683: _that's_ what I've been looking for!

docker-compose.yml
```diff
@@ -183,6 +189,68 @@ services:
...
+      - "18734:18734"
+    volumes:
+      - /edx/var/registrar/
+
+  registrar-worker:
+    command: bash -c 'source /edx/app/registrar/registrar_env && cd /edx/app/registrar/registrar && make requirements && celery -A registrar worker -l debug -c 2'
+    container_name: edx.devstack.registrar-worker
+    hostname: registrar-worker.devstack.edx
+    depends_on:
+      - lms
```
iloveagent57, line 230: does this really depend on LMS?  Also, should it depend on registrar?

iloveagent57 (commented):
Looks good to me, just the one question in the docker-compose section for registrar-worker.

docker-compose.yml
kdmccormick: * registrar-worker depends on LMS, because tasks like get-enrollments need to hit LMS APIs
* registrar depends on registrar-worker, so doing the converse too would be a circular dependency. Do you think the relationship should be reversed? My thinking for the direction I chose was "If someone starts Registrar, they probably want the worker started too".

iloveagent57: hmm, yeah, maybe it should be reversed - it would give you more options:
* `make dev.up.registrar` could give you just registrar, with no worker.
* `make dev.up.registrar-worker` gives you the webapp and the workers.

On the other hand, a user would then have to remember to type the extra "-worker" if they typically want the workers, too (which I think they will, as you've pointed out).  So it's probably a toss-up.  

I say leave it as you currently have it.

docker-compose-host.yml
```diff
@@ -37,7 +37,13 @@ services:
     volumes:
       - ${DEVSTACK_WORKSPACE}/gradebook:/edx/app/gradebook:cached
       - gradebook_node_modules:/edx/app/gradebook/node_modules
-      
+  registrar:
+    volumes:
+      - ../registrar:/edx/app/registrar/registrar
```
jmbowman: These entries should use `${DEVSTACK_WORKSPACE}` instead of `..` (the latter is the default value of the former, but that behavior can be overridden).

docker-compose-sync.yml
```diff
@@ -19,6 +19,9 @@ services:
   forum:
     volumes:
       - forum-sync:/edx/app/forum/cs_comments_service:nocopy
+  registrar:
+    volumes:
+      - registrar-sync
```
jmbowman: I think this entry should have `:/edx/app/registrar/registrar:nocopy` (or whatever the appropriate path in the container is) tacked onto the end.  I suspect we'll rip out docker-sync support pretty soon, but might as well get this right in the meantime.

docker-compose.yml
```diff
@@ -183,6 +189,68 @@ services:
...
+      DB_PORT: 3306
+      DB_USER: registrar001
+      DB_PASSWORD: password
+      LMS_HOST: http://localhost:18000
+      MEMCACHE_HOST: edx.devstack.memcached
+      DJANGO_SETTINGS_MODULE: registrar.settings.devstack
+      CELERY_BROKER_TRANSPORT: redis
+      CELERY_BROKER_HOSTNAME: edx.devstack.redis:6379
+      CELERY_BROKER_VHOST: 10
+      CELERY_BROKER_PASSWORD: password
```
jmbowman, line 218: This is probably ok for now, but I suspect we'll want to move some of these environment variables into the `configuration` repo code that builds the container.

docker-compose.yml
```diff
@@ -183,6 +189,68 @@ services:
...
+    image: edxops/registrar:${OPENEDX_RELEASE:-latest}
+    ports:
+      - "18734:18734"
+    volumes:
+      - /edx/var/registrar/
+
+  registrar-worker:
+    command: bash -c 'source /edx/app/registrar/registrar_env && cd /edx/app/registrar/registrar && make requirements && celery -A registrar worker -l debug -c 2'
```
jmbowman: Why the need to run `make requirements` on every container launch?

docker-sync.yml
```diff
@@ -29,3 +29,8 @@ syncs:
     host_disk_mount_mode: 'cached'
     src: '../cs_comments_service/'
     sync_excludes: [ '.git', '.idea' ]
+
+  regsistrar-sync:
+    host_disk_mount_mode: 'cached'
+    src: '../registrar/'
+    sync_excludes: [ '.git', '.idea' Z]
```
jmbowman: Is the Z at the end a typo?

kdmccormick: Yes, thanks.

docker-compose.yml
kdmccormick: Great question. I'm not sure why that's there. I'll remove it.

docker-compose.yml
kdmccormick: I agree. I'll make a note of that for later.

jmbowman (approved):

PR #408, opened 2019-05-15
Declare OEP-7 compliance.

adzuci:


kdmccormick:
While you're cleaning up `openedx.yaml`, you can remove `track-pulls`, too. It's an obsolete key per OEP-2

PR #409, opened 2019-05-16
WIP: change container names so that dns works on local machine and in docker

emma-green:


docker-compose.yml
```diff
@@ -13,7 +13,11 @@ version: "2.1"
 services:
   # Third-party services
   chrome:
-    container_name: edx.devstack.chrome
+    container_name: chrome.local.edx.org
+    network:
```
cpennington: We're using compose format 2.1 (see above in this file), which appears to want `networks`, not `network`: https://docs.docker.com/compose/compose-file/compose-file-v2/#aliases

cpennington:
Looking at the latest failure, you might just need to declare the network that all of these hosts should be in, rather than relying on the named default.

pwnage101:
Have you considered the devstack URLs in other repositories, such as devstack.py settings in IDAs? e.g. https://github.com/edx/credentials/blob/master/credentials/settings/devstack.py#L52-L53

emma-green:
> Have you considered the devstack URLs in other repositories, such as devstack.py settings in IDAs? e.g. https://github.com/edx/credentials/blob/master/credentials/settings/devstack.py#L52-L53

Oooh. Good point! I have not. I aliased the container names on the network though, so it should continue to work. Any other repos that you would suggest I look at?

pwnage101:
Firsthand, I've seen such URLs in devstack.py files in edx/ecommerce and edx/edx-notes-api, but I'm sure there are more.

emma-green:
I was really hoping that there weren't too many places where this was hardcoded...

staubina:
These URLs are used through configurations in other services.  I need to dig up where they are, but you might also need to make updates to the provision scripts.  Like @pwnage101 mentioned these URLs are also set in the devstack.py settings in probably every Dockerized IDA.

docker-compose-marketing-site.yml
```diff
@@ -7,17 +7,21 @@ services:
       - MARKETING_SITE_ROOT="http://localhost:8080"
 
   marketing:
-    container_name: edx.devstack.marketing
+    container_name: marketing.local.edx.org
+    networks:
+      devstack_default:
+        aliases:
+          - edx.devstack.marketing
```
staubina, line 14: Would it be simpler to add the new names as an Alias?

regisb:
This PR makes a lot of sense. However, it leaves out the instructions for the users to configure DNS records in their /etc/hosts file. As far as I understand, users who want to access services running in local containers need to modify the configuration of their DNS server, right? It would make sense to give more detailed instructions.

cpennington:
This actually banks on the fact that `*.local.edx.org` resolves to 127.0.0.1. If you want to use it w/o an internet connection, you would need local entries in `/etc/hosts`.

regisb:
Interesting, I had not noticed that the *.local.edx.org domain names resolved to 127.0.0.1. Maybe it would make sense to mention this in the README?

emma-green:
> Interesting, I had not noticed that the *.local.edx.org domain names resolved to 127.0.0.1. Maybe it would make sense to mention this in the README?

Hi! This pr is not in active development, as other things were prioritized in front of it (last year). I'm happy to have you take it over as owner!

regisb:
> Hi! This pr is not in active development, as other things were prioritized in front of it (last year). I'm happy to have you take it over as owner!

Unfortunately I don't have time to take this on myself. If you can no longer work on this, can you please close this PR? You should feel free to re-open it at a later time, if necessary.

PR #412, opened 2019-05-21
Add program-manager to devstack

kdmccormick:
@edx/masters-neem 
[EDUCATOR-4331](https://openedx.atlassian.net/browse/EDUCATOR-4331)

This PR:
- Adds `frontend-app-program-manager` to devstack in its own container
- Does a bit of clean-up for the `gradebook` devstack installation

Both frontends' containers are build from the same image (`node:10`)

We discussed adding all microfrontends to their own docker-compose file, so that they are not started by `make dev.up`. I opted against this due to the cumbersomeness of adding `-f docker-compose-frontends.yml` to like 10 different Makefile rules. My thinking is that (1) it does not take long to start the MFE containers, and (2) one can avoid starting the MFE containers by instead running `make dev.up.<service-name>`.

If you folks disagree, however, I can definitely switch back to the separate-docker-compose-file approach.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -300,15 +300,28 @@ services:
       - devpi_data:/data
 
   gradebook:
-    command: bash -c 'npm install && npm run start'
+    extends:
+      file: microfrontend.yml
+      service: microfrontend
     working_dir: '/edx/app/gradebook'
     container_name: edx.devstack.gradebook
-    image: node:10
     ports:
       - "1994:1994"
-    environment:
-      - NODE_ENV=development
-      
+    depends_on:
+      - lms
+
+  program-manager:
+    extends:
+      file: microfrontend.yml
+      service: microfrontend
+    working_dir: '/edx/app/program-manager'
+    container_name: edx.devstack.program-manager
+    ports:
+      - "1976:1976"
+    depends_on:
+      - lms
+      - registrar
+
 volumes:
   discovery_assets:
   edxapp_lms_assets:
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -308,7 +308,22 @@ services:
...
+    command: bash -c 'npm install && npm run start'
+    working_dir: '/edx/app/program-manager'
+    container_name: edx.devstack.program-manager
+    image: node:10
+    ports:
+      - "1976:1976"
+    environment:
+      - NODE_ENV=development
+    depends_on:
+      - lms
```
zacharis278, line 322: since this is basically just a transitive dependency through registrar I'm thinking we should just leave this out.

kdmccormick: it directly uses LMS for authN

kdmccormick:
The build is failing with:

```
+./provision-ecommerce.sh
...
...
 Applying catalogue.0032_journal_product_class...Traceback (most recent call last):
  File "manage.py", line 10, in <module>
...
...
django.db.utils.ProgrammingError: (1146, "Table 'ecommerce.catalogue_historicalproductattribute' doesn't exist")
Makefile:47: recipe for target 'migrate' failed
make: *** [migrate] Error 1
...
...
```

This seems to be unrelated to my changes.. does anyone know what might be causing this?

EDIT: yeah, this is failing on #413 too. It's probably an ecommerce issue.
EDIT 2: It is an ecommerce issue, hopefully fixed by https://github.com/edx/ecommerce/pull/2255

zacharis278 (approved):
don't know anything about the build failure but otherwise this lgtm

PR #416, opened 2019-05-31
add make target for gathering state of feature toggles

estute:
This make target will collect the sql dump data from the waffle table in an IDA (and waffle-utils, in the case of platform). These will later be consumed by the feature toggle reporter and used to generate an HTML report on the overall state of toggles in a deployment (in this case devstack).

### DIFF
--- a/gather-feature-toggle-state.sh
+++ b/gather-feature-toggle-state.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+
+# gather-feature-toggle-state.sh
+
+# as part of our feature toggle reporting utility, this script can be run
+# against a running devstack to collect the state of all of feature toggles
+# (django-waffle, waffle-utils) currently configured for each of the IDAs
+# that use them.
+
+if [ -e feature-toggle-data ]; then
+    rm -Rf feature-toggle-data
+fi
+mkdir feature-toggle-data
+
+docker exec -t edx.devstack.credentials bash -c 'source /edx/app/credentials/credentials_env && python /edx/app/credentials/credentials/manage.py dumpdata waffle --format=yaml > /edx/app/credentials/credentials/credentials_waffle.yml'
+docker cp edx.devstack.credentials:/edx/app/credentials/credentials/credentials_waffle.yml feature-toggle-data
+docker exec -t edx.devstack.credentials bash -c 'rm /edx/app/credentials/credentials/credentials_waffle.yml'
+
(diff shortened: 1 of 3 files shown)


michaelyoungstrom (approved):

gather-feature-toggle-state.sh
```diff
@@ -0,0 +1,37 @@
...
+
+docker exec -t edx.devstack.lms bash -c 'source /edx/app/edxapp/edxapp_env && python /edx/app/edxapp/edx-platform/manage.py lms dumpdata waffle --format=yaml > /edx/app/edxapp/edx-platform/lms_waffle.yml'
+docker exec -t edx.devstack.lms bash -c 'source /edx/app/edxapp/edxapp_env && python /edx/app/edxapp/edx-platform/manage.py lms dumpdata waffle_utils --format=yaml > /edx/app/edxapp/edx-platform/lms_waffle_utils.yml'
+docker cp edx.devstack.lms:/edx/app/edxapp/edx-platform/lms_waffle.yml feature-toggle-data
+docker cp edx.devstack.lms:/edx/app/edxapp/edx-platform/lms_waffle_utils.yml feature-toggle-data
+docker exec -t edx.devstack.lms bash -c 'rm /edx/app/edxapp/edx-platform/lms_waffle*.yml'
+
+docker exec -t edx.devstack.studio bash -c 'source /edx/app/edxapp/edxapp_env && python /edx/app/edxapp/edx-platform/manage.py cms dumpdata waffle --format=yaml > /edx/app/edxapp/edx-platform/studio_waffle.yml'
+docker exec -t edx.devstack.studio bash -c 'source /edx/app/edxapp/edxapp_env && python /edx/app/edxapp/edx-platform/manage.py cms dumpdata waffle_utils --format=yaml > /edx/app/edxapp/edx-platform/studio_waffle_utils.yml'
```
jmbowman: The lms and studio containers both use the same MySQL database; is there any point in dumping the data from these tables twice?

estute: I wanted the output of this to generate artifacts that would be similar to what we would get from running a similar script in production. I was assuming they might use a separate DB in production, but perhaps that is incorrect. Do you happen to know, @jmbowman ?

jmbowman: I'm pretty sure they're using the same database even in production, although we'd like to eventually change that.  That's why switching Studio to log in via LMS was a relatively simple change; they share the same users table, etc.

estute: Ok, then there is no need (for now). I'll simplify it

PR #420, opened 2019-06-05
Add dependencies between more of the devstack services so that they s…

cpennington:
…tart in the right order

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -132,6 +132,8 @@ services:
     depends_on:
       - mysql
       - memcached
+      - lms
+      - discovery
     # Allows attachment to the ecommerce service using 'docker attach <containerID>'.
     stdin_open: true
     tty: true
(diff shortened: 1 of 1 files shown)


crice100 (commented):
Are these changes already included in https://github.com/edx/devstack/pull/419?

cpennington:
Yes.

docker-compose.yml
```diff
@@ -150,6 +152,10 @@ services:
       - mysql
       - memcached
       - mongo
+      - discovery
+      - forum
+      - firefox
```
robrap, line 157: Do we need the browsers?

robrap (approved):
One question.

crice100 (approved):

docker-compose.yml
cpennington: Yeah, if you want to run bok_choy tests. I didn't want to add another barrier to being able to run those tests out of the box.

PR #422, opened 2019-06-11
REVMI-272 Expose db port and update readme

crice100:


docker-compose.yml
```diff
@@ -74,8 +74,8 @@ services:
       MYSQL_ROOT_PASSWORD: ""
       MYSQL_ALLOW_EMPTY_PASSWORD: "yes"
     image: mysql:5.6
-    # ports:
-    #   - "3306:3306"
+    ports:
+      - "3506:3306"
```
robrap, line 78: @feanil @macdiesel: Any idea why we didn't expose this by default to begin with?  Is there some particular concern, or was it just the possible port conflict?

README.rst
```diff
@@ -594,6 +594,22 @@ To rebuild static assets for all service containers:
 
    make static
 
+How do I connect to the databases from an outside editor?
+---------------------------------------------------------
+
+To connect to the databases from an outside editor (such as MySQLWorkbench),
+use the values below. Note that the username and password will vary depending
```
robrap, line 601: FYI: We no longer manually wrap our docs.  I'll leave it to you whether you think it is better to be internally consistent in the file or not. :)

robrap (approved):
Thanks.  Just a question for others that you might want to wait on.

docker-compose.yml
feanil: According to this: https://github.com/edx/devstack/commit/21a2f282f8f4acfa05948e301050aa0a4cfdc1a0

It looks like we did it because we didn't want it to conflict with the vagrant VM.  I think it's safe to turn this on now.

PR #423, opened 2019-06-13
dump feature toggle data as json

estute:
For some reason, course discovery was explicitly tagging the creation & modification dates for rows in the waffle table as timestamps (`!! timestamp`) when being dumped as YAML. The other IDAs do not do this. This inconsistency was causing issues when parsing the resulting yaml files in the feature toggle reporter, because PyYaml automatically converts values typed as timestamps and lops off their timezone values. This PR changes the output format of the gather-feature-toggle-state.sh to be JSON instead of YAML, which does not have this inconsistency between the IDAs.

### DIFF
--- a/gather-feature-toggle-state.sh
+++ b/gather-feature-toggle-state.sh
@@ -12,20 +12,20 @@ if [ -e feature-toggle-data ]; then
 fi
 mkdir feature-toggle-data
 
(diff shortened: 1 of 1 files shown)


michaelyoungstrom (approved):

jmbowman (approved):

regisb:
I believe this PR was superseded by https://github.com/edx/devstack/pull/429. I'm closing this now, please feel free to re-open if you think this is a mistake.

PR #442, opened 2019-09-25
Update DB snapshots

coryleeio:


edxapp_csmh.sql
```diff
@@ -24,61 +24,6 @@
 CREATE DATABASE /*!32312 IF NOT EXISTS*/ `edxapp_csmh` /*!40100 DEFAULT CHARACTER SET utf8 */;
 
 USE `edxapp_csmh`;
-
```
jdmulloy: I think we still need this table and it doesn't look like it's been added to the other table. Is CSMHE turned on in devstack when you generated these dumps?

jdmulloy (changes requested):

edxapp_csmh.sql
coryleeio: That dump went wrong - It has been updated now and should be good.

jdmulloy (approved):

PR #449, opened 2019-10-23
Deprecate 'make pull', add 'make dev.pull.<service>'

kdmccormick:
@nasthagiri 

Running `make pull` is rarely required, as it pulls __all__ images. Few devs need every single devstack Docker image, so this is a big waste of time and disk space.

This PR discourages its use by soft-deprecating it, directing devs to use `make dev.pull.<serivce>` instead, which pulls only a service and its dependencies' images. For example, `make dev.pull.lms` will pull only the images of LMS and its dependencies (`edxops/edxapp`, `mysql`, `chrome`, etc.)

It also adds `make dev.pull`, which does what `make pull` used to do.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ before_install:
 
   - make requirements
   - make dev.clone
-  - if [[ $DEVSTACK == 'lms' ]]; then make pull; fi
+  - if [[ $DEVSTACK == 'lms' ]]; then make dev.pull; fi
   - if [[ $DEVSTACK == 'analytics_pipeline' ]]; then make dev.up.analytics_pipeline; fi
 
 script:
--- a/Makefile
+++ b/Makefile
@@ -140,12 +140,21 @@ xqueue-logs: ## View logs from containers running in detached mode
 xqueue_consumer-logs: ## View logs from containers running in detached mode
 	docker-compose -f docker-compose-xqueue.yml logs -f xqueue_consumer
 
-pull:
-	@echo "This command is deprecated."
+RED="\033[0;31m"
+YELLOW="\033[0;33m"
+GREY="\033[1;90m"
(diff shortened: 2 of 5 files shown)


nasthagiri (approved):
As a user, `dev.pull.<service>` works for me!

feanil (approved):

cpennington:
Do we need to prefix everything with `dev.`? Could we instead look to deprecate that prefix and move things to the top level (since everything is part of devstack anyway?)

kdmccormick:
@cpennington I agree with you, but there doesn't seem to be any consistency behind the naming scheme currently, so I went with something that would make sense but also purposefully break `make pull`

kdmccormick:
An alternative could be:
* `make pull` is deprecated
* `make pull-all` pulls all
* `make pull.<service>` pulls a single service

kdmccormick:
On second thought @cpennington , maybe this is chance to put `pull` in line with the pattern of `<service>-<command>`, like `make lms-shell` and `make lms-attach`.

The pulling-all version could be `make all-pull`

cpennington:
Yeah, I don't know what scheme we actually want.

That said, as far as breaking/deprecating goes: You can always make the old-bad command call the new-good command as a dependency, and then print out a screaming warning message after the command runs.

cpennington:
FYI: if you name your `fixup! Some message` as `fixup! Commit message to be squashed into`, then you can use `git rebase --interactive --autosquash` at it'll auto-reorder them for you. (https://thoughtbot.com/blog/autosquashing-git-commits has more details)

kdmccormick:
Cool tip; thanks @cpennington

kdmccormick:
Tests are broken (hope to fix this soon), merging anyway :crossed_fingers:

PR #454, opened 2019-11-21
Change notes container name to edxnotesapi.

azarembok:
API calls from LMS to notes were failing due to the fact that the container name is used to identify the target of the call.  The underscores in edx_notes_api caused an illegal hostname exception on the notes side, which caused the api calls to return errors.

This PR removes the underscores from the container name.  It also enhances the provision_ida.sh script to allow an optional container name parameter to be passed in, allowing the container name to be different from the app name.


fysheets (approved):

fysheets:
Pending tests, of course :)

azarembok:
Tests are currently failing due to a timeout error.  Tracking that issue in https://openedx.atlassian.net/browse/PROD-1027

PR #456, opened 2019-12-02
Use node 12 for the program-manager devstack container

matthugs:
JIRA:EDUCATOR-4755

Will also require a change of travis configuration in the repo itself, i.e. https://github.com/edx/frontend-app-program-manager/pull/14

Tested over several weeks of using the service in devstack without realizing it 😛

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -325,6 +325,7 @@ services:
       file: microfrontend.yml
       service: microfrontend
     working_dir: '/edx/app/program-manager'
+    image: node:12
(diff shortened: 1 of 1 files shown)


kdmccormick (approved):
Looks good! Any reason we wouldn't want to do this with Gradebook as well?

matthugs:
@kdmccormick purely that this is what's been sitting around being tested on my devstack. I can do both at once if we know someone else will be touching gradebook & we'd like to get it onto their devstack to test. Lmk if there's someone in that situation to motivate that

PR #459, opened 2019-12-06
Update microfrontend.yml image to be Node:12 version

schenedx:
Update the gradebook container node version from 10 to 12

@edx/masters-devs Please review.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -325,7 +325,6 @@ services:
       file: microfrontend.yml
       service: microfrontend
     working_dir: '/edx/app/program-manager'
-    image: node:12
     container_name: edx.devstack.program-manager
     ports:
       - "1976:1976"
--- a/microfrontend.yml
+++ b/microfrontend.yml
@@ -5,6 +5,6 @@ version: "2.1"
 services:
   microfrontend:
     command: bash -c 'npm install && npm run start'
-    image: node:10
+    image: node:12
     environment:
       - NODE_ENV=development


iloveagent57 (approved):
LGTM (assuming tests pass)

abutterworth:
Looks like the PR travis build took 50 mins to run and was canceled due to exceeding a build time limit. Is this normal for devstack? (I just restarted the travis build)

kdmccormick:
@abutterworth Unfortunately, it is normal for devstack. I don't know of a workaround; I ended just merging my last devstack PR with tests failing.

@jmbowman mentioned that the fix would likely be making provisioning faster.

kdmccormick:
Looks like @jinder1s is/was working on the issue: https://github.com/edx/devstack/pull/458

matthugs (approved):

docker-compose.yml
```diff
@@ -314,6 +314,7 @@ services:
       file: microfrontend.yml
```
matthugs: We should change this upstream in this file, which is shared between this and the programs learner portal iirc.

matthugs (changes requested):
Oh wait actually I almost forgot

docker-compose.yml
schenedx: @matthugs What should we change here? The content in `microfrontend.yml` has node at v10. That's not going out of support by end of year. @abutterworth because we can leave that in for now.

kdmccormick: I support updating it in `microfrontend.yml`, because only Gradebook and ProgramManager use that file.

schenedx: OK, I upgraded the node version on `microfrontend.yml`. This also means we can remove the `image` override we have in `docker-compose.yml`. @matthugs and @kdmccormick Please review again

matthugs (approved):

PR #461, opened 2019-12-14
Parameterize provisioning script by service

kdmccormick:
The provision script now takes a parameter: `./provision.sh lms`

This allows the creation of a new Make target: `make dev.provision.<single_service>`

Hopefully this should save time when just one of your services gets messed up, but the rest are fine :)

todo:
* local testing
* fix analytics makefile targets


kdmccormick:
Merged along with #463

PR #463, opened 2019-12-17
Parallelizing travis test 

jinder1s:
Hi, I'm close to the end of making changes here. Before I go further, I would like you guys to look through the changes and suggest any improvements.

Writeup for PR:
Problem: Devstack travis tests were taking too long, to the point where devstack was not being tested correctly. 

The changes made in this PR try change the test flow to decrease test time in each travis shard. 

Some changes made:
-Testings things in parallel rather than just as one big test
-adding ability to provision things one at a time rather than all as one(thank you kyle for the wonderful work here)
-adding option to not cache after each dev.up.% command by adding flag

Some things found during changes and possible improvements in the future:
-very difficult to separate forum and course-discovery from lms. They need lms to run correctly.
-course-discovery provisioning and program cache flow can probably be improved. 
- currently, if you only want to provision few services(vs all or just one), you need to call provision time one at a time. It might be a good idea to add command that takes additional inputs for each of the provision things.  Such as: dev.provision service1 service2 ...


healthchecks.sh
```diff
@@ -1,36 +1,73 @@
 set -x
 
-echo "Checking LMS heartbeat:"
-curl http://localhost:18000/heartbeat
-if [ $? -ne 0 ]; then
-    docker-compose logs
-    exit 2
+service=$1
```
jmbowman, line 3: It looks like this script lost the ability to run all the healthchecks, without specifying a service.  That's still used in `make healthchecks`.

jmbowman (commented):
Looks good other than the one issue noted.

jmbowman (approved):

PR #466, opened 2019-12-27
Make provision.sh handle arbitrary # of arguments

kdmccormick:
This allows the passing of multiple arguments to `provision.sh`, and creates a Makefile rule that allows provisioning multiple services at once. For example:
```
make dev.provision.services.lms+ecommerce+forum
```

@jinder1s this is ready for review


jinder1s (approved):
everything looks good

PR #470, opened 2020-01-22
Add all targets to PHONIES

kdmccormick:
```
'make dev.provision' was trying to provision
the 'o' service, due strangeness related to the fact
that Makefiles were designed for compiling C
programs (which have .o intermediate files).

Fix the issue by adding all rules to PHONIES,
which tells Make that none of the targets are
real files.

Co-authored-by: Tim McCormack <tmccormack@edx.org>
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,26 @@
 #
 ########################################################################################################################
 .DEFAULT_GOAL := help
-.PHONY: requirements
+
+.PHONY: analytics-pipeline-devstack-test analytics-pipeline-shell backup \
+        build-courses check-memory create-test-course credentials-shell \
+        destroy dev.checkout dev.clone devpi-password dev.provision \
+        dev.provision.analytics_pipeline dev.provision.analytics_pipeline.run \
+        dev.provision.xqueue dev.provision.xqueue.run dev.pull dev.repo.reset \
+        dev.reset dev.status dev.sync.daemon.start dev.sync.provision \
+        dev.sync.requirements dev.sync.up dev.up dev.up.all \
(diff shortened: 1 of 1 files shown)


timmc-edx:
Thoughts on including other targets as well, such as `update-db` and `lms-shell`? I'm not sure which ones are most likely to cause problems.

Also, apparently one common Makefile style guide recommends putting `.PHONY: foo` above each target `foo:`, but I'm not sure I can stomach that. :-P

timmc-edx:
(I see a total of 77 targets we could include, not counting pattern rules.)

kdmccormick:
@timmc-edx Whoops, good catch. My script wasn't grabbing targets with dashes.

I've seen that recommendation before as well, and I lean towards just maintaining a big blob of phonies at the top instead of peppering the noise throughout the file. I can definitely polish my PHONIES-generation script and push it to repo-tools if that's something people would find useful.

timmc-edx:
You'll want numbers as well, for `e2e-shell` and `e2e-tests`.

This is what I ended up with, for reference: `grep '^[^:%]+(?=:)' -Po ~/openedx/devstack/Makefile | grep -v '\t|#' -P | grep '[a-z]'` ("Any static pattern that isn't all uppercase.")

kdmccormick:
How about now?

timmc-edx (approved):

kdmccormick:
[The Travis failure](https://travis-ci.org/edx/devstack/jobs/640876409?utm_medium=notification&utm_source=github_status) seems to be unrelated and is also occurring on Master.

PR #471, opened 2020-01-28
ADR 2: Development direction for devstack

kdmccormick:
These points came up at an Arch lunch some time in late 2019. @nasthagiri mentioned it'd be good to capture them in an ADR, which I've done my best with here.

### DIFF
--- a/docs/decisions/0002-devstack-direction.md
+++ b/docs/decisions/0002-devstack-direction.md
@@ -1,0 +1,82 @@
+# 2. Development direction for devstack
+
+## Status
+
+Accepted, circa January 2020.
+
+## Context
+
+The edX Docker Development Stack (henceforce just "Devstack")
+that exists in this repository has
+served as the canonical edX development environment since
+we moved off of
+[Vagrant Devstack](https://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/open-release-eucalyptus.master/installation/devstack/install_devstack.html)
+circa 2017.
+This switch seems to have been made in order to aid
+developers who needed to develop across edX microservices
+in a single development environment,
+a task for which Docker Compose (upon which Devstack is built)
+seemed well suited.
+
+Based on the number of devstack-related questions in the edX-internal
+and Open edX Slack workspaces,
+it is clear that Devstack is depended upon by a number of developers.
+Exactly how many developers use Devstack and
+exactly which of its features they
+they depend upon is less clear;
+this may be a future good discovery topic.
+
+However, since the edX Tools team has been dissolved,
+with its members dispersing to the Architecture and Data Engineering Squads,
+Devstack has no clear stewarding team.
+A handful of developers continue to contribute
+to the repository in the time that they have
+to address issues that arise. These include / have included:
+* Engineers waste chunks of time by "nuking" and re-provisioning
+  their Devstacks when problems arise,
+  instead of debugging and then committing solutions back to the repository.
+  This is likely due to both Devstack documentation/usability issues
+  and a general lack of edX engineering proficiency with Docker and Docker Compose.
+* Pulling images for all Devstack services takes a while.
+* Starting Devstack takes a while.
+* Devstack PR builds were timing out in Travis because provisioning took over an hour.
+
+Engineers have discussed different strategies to improving Devstack.
+However, two obstacles stand in the way of a concerted effort for improvement:
+1. As mentioned, nobody "owns" Devstack, so it is unclear who would lead such
+   an initiative and where the developer resources  would come from.
+2. The DevOps team is currently moving us towards Kubernetes for deployment.
+   We know that this may open up new and hopefully-improved possibilities
+   for local development,
+   making us hesitant to seriously invest in the current devstack
+   before we are further along on the Kubernetes transition.
+
+Given all of that, it seems that a balance needs to be struck
+between the extremes of "completely ignore Devstack"
+and "seriously invest in Devstack".
+
+## Decision
+
+We feel that it makes the most sense to take a *medium-term*
+approach to Devstack development,
+investing in quick-wins where we can,
+but making sure not to sink significant amounts of developer time into
+refactoring or remaking Devstack.
+
+Here are examples of some successful "quick wins" we have implemented in the past few months:
+* [#398](https://github.com/edx/devstack/pull/398): Bring up a single service instead of all of them.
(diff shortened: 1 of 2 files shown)


cpennington:
I'm not sure if this belongs in the devstack repo (because it's really a question on how the docker images get built, not a question of how they are used in devstack), but it might be worth capturing an intention to move docker image creation into the app repos, in order to speed up image builds, slim down image sizes, and co-locate the code and deployment configuration more.

kdmccormick:
@cpennington Are you commenting on a specific line? The ADR is meant to touch on a bunch of stuff, most of which is unrelated to image size.

If you're referencing the "images take a long time to pull", a more this-repo-specific way of saying that could be "we pulled too many images."

cpennington:
No, not a specific line, more of a general question about devstack (which runs a collection of images), and how it should work.

That said "images take a long time to pull" is also a result of "images are too damn large".

kdmccormick:
@cpennington Right, I misread your comment at first; posting a better response now.

kdmccormick:
@cpennington I'm cool with those points going in this repo because I don't know of any better place to put decisions related to them.

However, those changes you're describing seem more in the category of "significant improvements that we shouldn't spend time on right now". That could just be a result of my own lack of expertise, and maybe you or someone else could do that work fairly quickly.

I'm not going to add those points to this ADR, but definitely feel free to make a PR against this one or even make a new ADR along the lines of "How we want our devstack to work in the future".

docs/decisions/0002-devstack-direction.md
```diff
@@ -0,0 +1,82 @@
+# 2. Development direction for devstack
```
jmbowman, line 1: OEP-19 says we won't use Markdown for ADRs: https://open-edx-proposals.readthedocs.io/en/latest/oep-0019-bp-developer-documentation.html#format

docs/decisions/0002-devstack-direction.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Context
+
+The edX Docker Development Stack (henceforce just "Devstack")
+that exists in this repository has
+served as the canonical edX development environment since
+we moved off of
+[Vagrant Devstack](https://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/open-release-eucalyptus.master/installation/devstack/install_devstack.html)
+circa 2017.
+This switch seems to have been made in order to aid
```
jmbowman, line 15: This wording seems awfully uncertain regarding a decision made largely by people who are still here...

docs/decisions/0002-devstack-direction.md
```diff
@@ -0,0 +1,82 @@
...
+and Open edX Slack workspaces,
+it is clear that Devstack is depended upon by a number of developers.
+Exactly how many developers use Devstack and
+exactly which of its features they
+they depend upon is less clear;
+this may be a future good discovery topic.
+
+However, since the edX Tools team has been dissolved,
+with its members dispersing to the Architecture and Data Engineering Squads,
+Devstack has no clear stewarding team.
```
jmbowman, line 31: It never really had one; I was the only person on the Tools team who did any significant work on devstack, and I was already doing that while I was still on the Platform team.  From the beginning, Docker devstack was a collaborative effort between multiple individuals not belonging to any single team or even working group.

docs/decisions/0002-devstack-direction.md
```diff
@@ -0,0 +1,82 @@
...
+* [#398](https://github.com/edx/devstack/pull/398): Bring up a single service instead of all of them.
+* [#431](https://github.com/edx/devstack/pull/431): Add a shortcut for running a script inside one containers.
+* [#449](https://github.com/edx/devstack/pull/449): Allow pulling of images for single service and its dependencies.
+* [#463](https://github.com/edx/devstack/pull/463): Allow provisioning of individual services and parallelize Travis tests.
+* [#466](https://github.com/edx/devstack/pull/466): Allow provisioning of set of services.
+
+Here is some future work that may be worth investing in:
+* Smarter program caching... we don't need to do it on every `dev.up.<service>` call.
+* Survey of how Devstack is used, which may be used in (a) improving its current form and (b) designing its successor after the Kubernetes transition.
+* Removing cruft from Devstack - there are *a lot* of rules in `Makefile`, many of which are probably unused.
```
jmbowman, line 76: My suspicion is that almost all of them are used, but the subsets that each developer ever uses vary greatly.  Would be nice to get some data to verify that, though.

Ok, maybe the `sync` commands are unused; that proved too unreliable to use, and we just never ripped it out.

docs/decisions/0002-devstack-direction.md
```diff
@@ -0,0 +1,82 @@
...
+* *If we are sticking with Docker and/or Docker Compose for the forseeable future* - Developer education on those tools.
+* __Increased clarity on ownership of edX's devstack, as it exists now and as it will exist in the future.__
```
jmbowman, line 78: I'm curious to see how that pans out; so far I've seen much stronger agreement that no team wants to own it than that it should even have an owner.  We discussed having a working group for it, maybe we should go ahead with that even if it doesn't translate into any concrete development resources, just to have a forum for discussion on the topic.

docs/decisions/0002-devstack-direction.md
```diff
@@ -0,0 +1,82 @@
...
+with its members dispersing to the Architecture and Data Engineering Squads,
+Devstack has no clear stewarding team.
+A handful of developers continue to contribute
+to the repository in the time that they have
+to address issues that arise. These include / have included:
+* Engineers waste chunks of time by "nuking" and re-provisioning
+  their Devstacks when problems arise,
+  instead of debugging and then committing solutions back to the repository.
+  This is likely due to both Devstack documentation/usability issues
+  and a general lack of edX engineering proficiency with Docker and Docker Compose.
```
abutterworth, line 39: Speaking for me, I don't debug problems in devstack because python, django, etc are not my core strengths. There's a lot of context to build up (both technically and in each project) when usually all I need is to make some minor change or simply have the lms apis running.

kdmccormick:
Thanks all for the review comments.

@jmbowman it looks like we're not aligned on what would be the best direction forward for Devstack (or if there should be a "direction" forward at all). It also looks like I'm missing some historical context here. 

Given that, I'd rather not try to push forward an architectural decision, so I'm going to close this.

A Devstack WG sounds like a cool idea. I wouldn't be able to initiate any such effort right now, but I'd try to participate if it did come into existence.

PR #473, opened 2020-02-03
Add new rule to clone repos using SSH

awaisdar001:
Issue
Need to change `.git/config.url` manually after each provision.

Description
I think it is due to 2FA that I need to change that value from HTTPS to SSH manually. 

Solution
Adding a new rule `make dev.clone.ssh` to use ssh links by default for new installations.

### DIFF
--- a/repo.sh
+++ b/repo.sh
@@ -23,6 +23,7 @@ else
     OPENEDX_GIT_BRANCH=master
 fi
 
+# When you add new services should add them to both repos and ssh_repos
 repos=(
     "https://github.com/edx/course-discovery.git"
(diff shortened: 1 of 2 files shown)


repo.sh
```diff
@@ -39,6 +39,22 @@ repos=(
     "https://github.com/edx/frontend-app-publisher.git"
 )
 
+ssh_repos=(
```
kdmccormick, line 43: Can you add a comment in this file saying that people who add new services should add them to both `repos` and `ssh_repos`?

kdmccormick (approved):
Nice change. Looks good to me.

I have one small comment (no need for re-review).

PR #482, opened 2020-03-03
Add a new Elasticsearch 5.6 container for testing upgrades.

dianakhuang:
This is done as part of [ES Upgrade](https://openedx.atlassian.net/wiki/spaces/AC/pages/1073742677/Elasticsearch+Upgrade+Task+Force) work.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,6 +35,12 @@ services:
       - elasticsearch_data:/usr/share/elasticsearch/data
       - elasticsearch_data:/usr/share/elasticsearch/logs
 
+  # This is meant to be used to test ES upgrades so that we do not have to upgrade all of our services to ES5 at once.
+  elasticsearch-5:
+    container_name: edx.devstack.elasticsearch-5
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -35,6 +35,11 @@ services:
       - elasticsearch_data:/usr/share/elasticsearch/data
       - elasticsearch_data:/usr/share/elasticsearch/logs
 
+  elasticsearch-upgrade:
```
nasthagiri: Since we're thinking of having multiple elastic search containers - one per version, what if the container name included the elastic search version #?

dianakhuang: Our current plan is to only have two containers at max. I don't mind changing the name to elasticsearch-5 or something like that, though.

nasthagiri: I understand the reason for having this duplicative elastic search container (based on our lunch convo). For those who weren't there (or when we forget), can you add a comment to explain its purpose?  Thanks!

nasthagiri (approved):

PR #483, opened 2020-03-04
Updated devstack nfs make targets

UsamaSadiq:
**Issue:** [BOM-1194](https://openedx.atlassian.net/browse/BOM-1194)

### Description
- Updated **make.dev.nfs.up** targets.
- Updated **docker-compose-host-nfs.yml** and **docker-compose-themes-nfs.yml** files.

#### Changed Commands
- Renamed **dev.provision.nfs.run** to **dev.nfs.provision.services**.
- Added **dev.nfs.provision.services.%** command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@
         destroy dev.checkout dev.clone dev.clone.ssh dev.nfs.provision dev.nfs.setup \
         dev.nfs.up dev.nfs.up.all dev.nfs.up.watchers devpi-password \
         dev.provision dev.provision.analytics_pipeline \
-        dev.provision.analytics_pipeline.run dev.provision.nfs.run \
+        dev.provision.analytics_pipeline.run dev.nfs.provision.services \
         dev.provision.services dev.provision.xqueue dev.provision.xqueue.run \
         dev.pull dev.repo.reset dev.reset dev.status dev.sync.daemon.start \
         dev.sync.provision dev.sync.requirements dev.sync.up dev.up dev.up.all \
@@ -53,7 +53,7 @@ export DEVSTACK_WORKSPACE
 export COMPOSE_PROJECT_NAME
 
 STANDARD_COMPOSE_FILES=-f docker-compose.yml -f docker-compose-host.yml -f docker-compose-themes.yml
-
+NFS_COMPOSE_FILES=-f docker-compose.yml -f docker-compose-host-nfs.yml -f docker-compose-themes-nfs.yml
 include *.mk
 
 # Generates a help message. Borrowed from https://github.com/pydanny/cookiecutter-djangopackage.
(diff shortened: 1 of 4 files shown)


docker-compose-themes-nfs.yml
```diff
@@ -0,0 +1,23 @@
...
+      - edx-nfs:/edx/app/edx-themes:cached
+  lms:
+      volumes:
+      - edx-nfs:/edx/app/edx-themes:cached
+  studio:
+      volumes:
+      - edx-nfs:/edx/app/edx-themes:cached
+
+volumes:
+  edx-nfs:
```
jmbowman: I'm not sure it's safe to use the same volume name as in `docker-compose-host-nfs.yml` with a different directory; maybe use something like `themes-nfs` instead?

Makefile
```diff
@@ -130,7 +130,7 @@ dev.nfs.up.watchers: | check-memory ## Bring up asset watcher containers
 	docker-compose -f docker-compose-watchers-nfs.yml up -d
 
 dev.nfs.up: | check-memory ## Bring up all services with host volumes
-	docker-compose -f docker-compose.yml -f docker-compose-host-nfs.yml up -d
+	docker-compose -f docker-compose.yml -f docker-compose-host-nfs.yml -f docker-compose-themes-nfs.yml up -d
```
jmbowman: Could you create a new variable like `DOCKER_COMPOSE_FILES_NFS` to capture these (similar to the existing `DOCKER_COMPOSE_FILES`) so we don't need to keep copying this list in different targets?

Makefile
```diff
@@ -139,7 +139,7 @@ dev.nfs.up.all: | dev.nfs.up dev.nfs.up.watchers ## Bring up all services with h
 dev.nfs.provision: | check-memory dev.clone dev.provision.nfs.run stop ## Provision dev environment with all services stopped
 
 dev.provision.nfs.run: ## Provision all services with local mounted directories
```
jmbowman: The original target this was copied from has since been renamed from `dev.provision.run` to `dev.provision.services`, this should be updated to match, maybe to `dev.nfs.provision.services`.  There should also be a `dev.nfs.provision.services.%` equivalent for `dev.provision.services.%` which limits which services are provisioned.

docker-compose-themes-nfs.yml
UsamaSadiq: okay, I'll update `edx-nfs` to `themes-nfs` for safety.

Makefile
UsamaSadiq: Sure, I was also thinking the same. I'll create another variable for the nfs arguments.

Makefile
UsamaSadiq: Renamed `dev.provision.nfs.run` to `dev.nfs.provision.services`. 
Added `dev.nfs.provision.services.*`.

Makefile
```diff
@@ -130,17 +130,19 @@ dev.nfs.up.watchers: | check-memory ## Bring up asset watcher containers
...
 
 dev.nfs.up: | check-memory ## Bring up all services with host volumes
-	docker-compose -f docker-compose.yml -f docker-compose-host-nfs.yml up -d
+	docker-compose $(NFS_COMPOSED_FILES) up -d
 	@# Comment out this next line if you want to save some time and don't care about catalog programs
 	#./programs/provision.sh cache >/dev/null
 
 dev.nfs.up.all: | dev.nfs.up dev.nfs.up.watchers ## Bring up all services with host volumes, including watchers
 
 dev.nfs.provision: | check-memory dev.clone dev.provision.nfs.run stop ## Provision dev environment with all services stopped
```
jmbowman: Need to match the rename of the target below in the dependencies here.

Makefile
```diff
@@ -53,7 +53,7 @@ export DEVSTACK_WORKSPACE
 export COMPOSE_PROJECT_NAME
 
 STANDARD_COMPOSE_FILES=-f docker-compose.yml -f docker-compose-host.yml -f docker-compose-themes.yml
-
+NFS_COMPOSED_FILES=-f docker-compose.yml -f docker-compose-host-nfs.yml -f docker-compose-themes-nfs.yml
```
jmbowman: Minor suggestion: could this be `NFS_COMPOSE_FILES` instead for consistency with the other variable?

Makefile
```diff
@@ -130,17 +130,19 @@ dev.nfs.up.watchers: | check-memory ## Bring up asset watcher containers
...
 dev.nfs.up.all: | dev.nfs.up dev.nfs.up.watchers ## Bring up all services with host volumes, including watchers
 
 dev.nfs.provision: | check-memory dev.clone dev.provision.nfs.run stop ## Provision dev environment with all services stopped
 
-dev.provision.nfs.run: ## Provision all services with local mounted directories
-	DOCKER_COMPOSE_FILES="-f docker-compose.yml -f docker-compose-host-nfs.yml" ./provision.sh
+dev.nfs.provision.services: ## Provision all services with local mounted directories
+	DOCKER_COMPOSE_FILES=$(NFS_COMPOSED_FILES) ./provision.sh
 
+dev.nfs.provision.services.*: ## Provision specified services with local mounted directories, separated by plus signs
```
jmbowman: I think this needs to be `.%`, not `.*`

Makefile
UsamaSadiq: renamed to `NFS_COMPOSE_FILES`.

Makefile
UsamaSadiq: Renamed to `NFS_COMPOSE_FILES`.

Makefile
UsamaSadiq: renamed to `dev.nfs.provision.services.%`.

Makefile
jmbowman: What I meant was that this line includes "dev.provision.nfs.run", but that target has been renamed to "dev.nfs.provision.services"; it needs to be renamed here also.

jmbowman (approved):

PR #484, opened 2020-03-04
Add learning MFE to devstack

stvstnfrd:
@abutterworth @davidjoy Any reason we don't yet want to include this?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -386,6 +386,17 @@ services:
       - lms
       - registrar
 
+  frontend-app-learning:
+    extends:
+      file: microfrontend.yml
+      service: microfrontend
+    working_dir: '/edx/app/frontend-app-learning'
+    container_name: edx.devstack.frontend-app-learning
+    ports:
+      - "2000:2000"
+    depends_on:
+      - lms
+
 volumes:
   discovery_assets:
   edxapp_lms_assets:
--- a/repo.sh
+++ b/repo.sh
@@ -35,6 +35,7 @@ repos=(
 )
 
 non_release_repos=(
+    "https://github.com/edx/frontend-app-learning.git"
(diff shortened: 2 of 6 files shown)


docker-compose.yml
```diff
@@ -343,6 +343,17 @@ services:
...
       - lms
 
+  frontend-app-learning:
+    extends:
+      file: microfrontend.yml
+      service: microfrontend
+    working_dir: '/edx/app/frontend-app-learning'
+    container_name: edx.devstack.frontend-app-learning
+    ports:
+      - "2000:2000"
```
ormsbee, line 396: @abutterworth, @davidjoy: Do we have any target convention for these port numbers?

ormsbee (approved):
This seems fine to me, after all the builds pass. It looks like a transient issue with edx-when?

docker-compose.yml
abutterworth: We've been counting up for each new MFE. If I recall correctly we started at @reillz10 's birth year (1992). So I guess that's our convention...

abutterworth (approved):

davidjoy (approved):

repo.sh
```diff
@@ -31,6 +31,8 @@ repos=(
     "https://github.com/edx/xqueue.git"
     "https://github.com/edx/edx-analytics-pipeline.git"
     "https://github.com/edx/frontend-app-gradebook.git"
+    "https://github.com/edx/frontend-app-learning.git"
+    "https://github.com/edx/frontend-app-program-console.git"
```
kdmccormick: Since this isn't in Juniper (afaik), I think you'll want to put this in the `non_release_repos` and `non_release_ssh_repos` lists.

stvstnfrd: Thanks for the tip! Done.

PR #485, opened 2020-03-04
Make no_cache option consistent

jdmulloy:


Makefile
```diff
@@ -131,8 +135,12 @@ dev.nfs.up.watchers: | check-memory ## Bring up asset watcher containers
 
 dev.nfs.up: | check-memory ## Bring up all services with host volumes
 	docker-compose -f docker-compose.yml -f docker-compose-host-nfs.yml up -d
-	@# Comment out this next line if you want to save some time and don't care about catalog programs
-	#./programs/provision.sh cache >/dev/null
+	@# If you want to save some time and don't care about catalog programs run "make no_cache=True dev.nfs.up"
+ifdef no_cache
+	echo "no_cache set: Not running %./programs/provision.sh cache"
```
jmbowman: Why the "%" only in this version of the echo statement?

jmbowman (approved):
Looks good other than that one question.

kdmccormick (approved):
Great change, thanks!

Makefile
jdmulloy: I think it's a typo. I'll remove it.

Makefile
```diff
@@ -108,14 +108,18 @@ dev.pull.%: ## Pull latest Docker images for a given service and all its depende
 
 dev.up: | check-memory ## Bring up all services with host volumes
 	bash -c 'docker-compose $(STANDARD_COMPOSE_FILES) up -d'
-	@# Comment out this next line if you want to save some time and don't care about catalog programs
+	@# If you want to save some time and don't care about catalog programs run "make no_cache=True dev.up"
+ifdef no_cache
```
awaisdar001, line 112: what about making it a little readable?

```
ifdef cache_programs
	$(WINPTY) bash ./programs/provision.sh cache
else
	echo "cache_programs set: Not running bash ./programs/provision.sh cache"
endif
```

Now if we run `make dev.up.lms` I won't run the caching. `make cache_programs=True dev.up.lms` will do the caching. 

Is the expected behavior is to run caching by default? I am not sure at this moment. 
Just a nit, but let me know what you think.
Thanks.

kdmccormick: A lot of devs depend on programs being cached by default, such as those of us working on Master's programs. On the other hand, a lot of devs probably don't need programs, and the by-default caching is slowing them down. I don't know right now who is in the majority.

I do like the suggestion of making it a positive flag (`cache_programs=True/False`) instead of a negative one.

My feeling is that we should leave `cache_programs=True` as default for this PR to avoid breaking anyone's flow. Then, if we want to, propose making `False` the default in another PR, but also tag some of the people who will be impacted for review.

kdmccormick: A nice middle ground would beg providing an easy way for devs to toggle caching On or Off without having to include it in each command. For example, having a git-ignored `.env.local` file that gets loaded when running each `make` target where one could define their environment variables.

regisb:
I just realised this PR exists. As suggested in #491 (which I just opened) I strongly suggest *not* caching programs by default. This command makes zero sense for 90% of Open edX developers. Instead of relying on `no_cache=True`, I think it would be better if developers explicitely ran `make dev.up dev.provision.programs` instead of `make dev.up`.

kdmccormick (changes requested):
I recommend closing this, as it's been achieved as part of https://github.com/edx/devstack/pull/491

PR #492, opened 2020-03-10
Fix upgrading by bumping pip-tools; use python 3.5 for CI

timmc-edx:


timmc-edx:
I'm not totally sure if downgrading Python from 3.6 to 3.5 is correct, since this isn't deployed per se, but it would put it in line with our other repos.

robrap:
@timmc-edx: I'd have the same question about Python 3.5 vs 3.6.  It would be great if we could have a documented decision around this for people to refer to.

jmbowman (approved):
Switching to 3.5 for consistency seems like a good idea.  We're likely to go to 3.8 next rather than 3.6 anyway, so not sure we gain much from 3.6 testing.

PR #491, opened 2020-03-10
Accelerate dev.up by disable program caching

regisb:
By default, program caching is executed every time a user runs `make
dev.up`. This is prohibitive and annoying for most developers who are
not working on programs. This change makes programs caching opt-out by
default. To launch the platform with programs caching, run:

    make dev.up.<service> dev.cache-programs

or:

    make dev.up.with-programs.<service>

or:

    make dev.up.with-programs

PR #493, opened 2020-03-11
Add ALWAYS_CACHE_PROGRAMS configuration option

kdmccormick:
By request, a follow-up to @regisb 's PR, which disables default program caching: https://github.com/edx/devstack/pull/491

This will allow developers to revert back to behavior where programs are copied to LMS cache upon startup by default by creating an `options.local.mk` file. This system can also be used in the future for any Devstack configuration options.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,6 @@ local.mk
 
 # Data used in feature toggle reporter
 feature-toggle-data/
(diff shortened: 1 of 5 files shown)


iloveagent57 (approved):
👍 I like it.

PR #497, opened 2020-03-18
Add mysql 5.7 container.

adzuci:


bderusha (approved):

regisb:
This commit just broke my devstack :cry: 

Could we please not use the devstack master branch for experimenting with database migrations? I do not want to run multiple copies of mysql next to one another, and I suspect no one else does.

A comment on PR #502 says "Might not help people that have upgraded in meantime". INDEED.

Mistakes happen, I get it. At the very minimum, could we have prominent instructions in the readme that explain that mysql57 should be disabled in docker-compose.yml? And warning that if we upgraded then we should pull new changes? There are many people out there relying on the devstack to work on the master branch of Open edX and it's important that it remains both stable and well documented.

I'm trying to get my devstack back up since yesterday and this is the _third_ undocumented, blocking issue which I have encountered. Reverse-engineering issues across the edx-platform, devstack and configuration repos takes a very long time and it's distressing.

adzuci:
Hey @regisb I'm looking into this now.  The original issue was that I added the separate mysql 5.7 container and pointed it at the mysql5.6 volume which resulted in the new mysql upgrading the data in a one way fashion.  I had believed that the issue had been resolved but clearly I was mistaken.  

@kdmccormick my instict was to merge https://github.com/edx/devstack/pull/502 to quickly resolve the issue, but I'm doing some more digging to understand the timeline and current status.

adzuci:
@regisb Thanks for raising this issue it looks like there is definitely more work to do.  The original issue was identified by @mikix and others when it cause them to accidentally upgrade their 5.6 container to 5.7.  We at least partially fixed this for future upgrades with https://github.com/edx/devstack/pull/504, however since some users are now using the 5.7 container to prepare for the future upgrade we don't want to remove the container altogether.

We are suggesting that folks who built outor upgraded devstack between `4e8f` (3/21) to `4a3240e27` (3/26) should:

1: Reprovision your mysql volume.
2: Force the upgrade by replacing the 5.6 container with the 5.7 container (if you do this know that your devstack will not catch 5.6 issues that may be found in pre 5.7 production environments.)
3: Figure out how to downgrade volumes: https://dev.mysql.com/doc/refman/5.7/en/downgrading-to-previous-series.html

One thing that may help with this issue and the future upgrade would be to add a config option to switch back and forth between 5.6 and 5.7 which we're now looking into.

kdmccormick:
Sorry that this screwed up your devstack database @regisb . Know that the backwards-incompatibility of this was unintentional; the container was meant to point at its own volume, which would have made it a no-op on a clean Devstack.

My DB volume also got messed up by this. The imperfect fix I went with is @adzuci 's approach # 2, which is done by going to the `mysql` section of docker-compose.yml and changing `image: mysql:5.6` to `image: mysql:5.7`. Your old data should still be intact, but you'll be running MySQL 5.7 instead of 5.6. The "config option" that @adzuci mentions would make it so that you could do this without putting your devstack checkout in a dirty state.

regisb:
What I did was simply to trash the mysql volume, but that's ok: there wasn't any precious data there. As I said to @adzuci in private conversation, it's perfectly fine to make mistakes. I am not concerned about the bug or the loss of data. My problem is that I wasted a lot of time finding the fix for this **undocumented** issue. 

I think this is a great time to add a CHANGELOG.md file to the Devstack. A changelog is the best place for developers to communicate breaking changes made to the devstack.

PR #502, opened 2020-03-26
WIP: Revert mysql57 container if needed

adzuci:


mikix (approved):
Might not help people that have upgraded in meantime, but stops any bleeding.

PR #505, opened 2020-03-27
Make provisioning, CI, and Makefile more consistent

kdmccormick:
## Overview
This PR does a bunch of clean-up and simplification. The most notable changes, from a developer's perspective, are:
* Marketing is tested in its own shard, instead of alongside LMS, Discovery and Forums.
* XQueue is now provisioned and health-checked as part of CI.
* xqueue, analyticspipeline, and marketing can be pulled and started just like any other service (`make dev.pull.analyticspipeline dev.up.analyticspipline`, etc.)
* `make $service-restart-devserver` can be used to restart the Django dev server inside a service's container.

I've made an effort to include no breaking changes in this PR, from a developer's point of view. Hence, I haven't updated any docs yet. I have some more changes planned, and I'd like to make them before circling back and updating all the documentation.

## Changes, commit-by-commit

1. Make registrar-shell match the other IDAs
    * `registrar-shell` had a custom implementation that was unnecessarily complex but did the exact same thing as the other IDAs' shell targets.
3. Generalize `xqueue-restart` into `<service>-restart-devserver`
    * `xqueue-restart` had a cool implementation that restarts the Django server. This seemed really useful, so I generalized it to work with Studio, LMS, Discovery, Registrar, etc. 
    * The name is ambiguous with `docker-compose restart`, so by @alanoe 's suggestion I added the `-devserver` suffix.
3. Require DB services to be up for 'make backup/restore'
    * The `make backup` and `make restore` targets need all three databases to be up in order to work, so we may as well make that a requirement of them.
4. Add more flexibility to 'make dev.pull'
    * Riffing on the pattern I introduced in `make dev.provision.services` of passing in multiple services using `+` as a separator, `make dev.pull.A+B+C` will pull images for services A, B, and C as well as all their dependencies.
      * This consistent interface allows me to simplify provisioning scripts in a later commit.
    * I also added the `make dev.onlypull.<services>` rule for pulling images *without* also their dependent images.
5. Let 'make dev.up.<service>' take multiple +-separated services
    * Same rationale as previous commit.
6. Use Makefile variables to clarfiy compose file scheme
    * This is mostly reverted in commit 12.
7.  Rename healthchecks.sh to check.sh
    *  The logic for the rename is that the name "healthchecks" may imply that all the script is doing is querying the /healthcheck endpoint on each service.
     * While this used to be the case, the script will soon be used for more sophisticated checks, such
    as the analytics pipeline acceptance tests, hence the generalized name for the script.
8. Clean up check.sh and allow it to take multiple args
9. Simplify handling of Analytics Pipeline and Marketing
    * Analytics and Marketing now follow the same Makefile target scheme for provisioning, pulling, and running that most other services do.
    * This allowed me to simplify `.travis.yml` a lot and delete `.travis/run.sh` entirely.
    * The old inconsistent targets are still in place, and are aliased to the new targets. I may pull all "backwards-compatibility" targets into their own sub-Makefile at some point.
    * Marketing is tested by itself in its own shard now.
10. Simplify handling of XQueue and add it to CI
    * XQueue now follows the same Makefile target scheme for provisioning, pulling, and running that all other services do.
    * It is provisioned and health-checked (via its `/status` endpoint) as part of CI.
11. Set configuration options in a way that allows them to be overriden via the command line.
12. Simplify Makefile handling of docker-compose files
     * Add new option SYNC_STRATEGY, which can be set to local-mounts, nfs, or docker-sync
    (defaults to local-mounts).
     *  This allows us to conditionally choose a value for DOCKER_COMPOSE_FILES in one place.
      * Also adds DEFAULT_SERVICES option which can be toggled to change the set of services that the generic targets (dev.up, dev.pull, etc.) act upon.
13.  Exit early in CI for cleaner error logs
      * If pulling or provisioning fails, then we shouldn't do healthchecks. That just makes the logs noisier.
14. Add `dev.up.without-deps.$` target. 
      * Thanks to @alanoe for the idea for this.


Makefile
```diff
@@ -102,8 +102,11 @@ dev.repo.reset: ## Attempts to reset the local repo checkouts to the master work
 dev.pull: ## Pull *all* required Docker images. Consider `make dev.pull.<service>` instead.
 	docker-compose pull
 
-dev.pull.%: ## Pull latest Docker images for a given service and all its dependencies
-	docker-compose pull --include-deps $*
+dev.onlypull.%: ## Pull latest Docker images for services (separated by plus-signs).
```
alanoe: IMHO dev.onlypull is not a good target name, as it is not obvious what it means. I'd recommend to pick a name which describes better what it will be done, eg dev.pull-without-deps

Makefile
```diff
@@ -236,12 +236,12 @@ pull.xqueue: ## Update XQueue Docker images
 validate: ## Validate the devstack configuration
 	docker-compose config
 
-backup: ## Write all data volumes to the host.
+backup: dev.up.mysql+mongo+elasticsearch ## Write all data volumes to the host.
```
alanoe, line 299: If I understood it right, a devstack must have been created with "make dev.provision" before using this. It'd be nice to mention it somewhere

Makefile
```diff
@@ -301,8 +301,8 @@ studio-restart: ## Kill the LMS Django development server. The watcher process w
 xqueue-shell: ## Run a shell on the XQueue container
 	docker exec -it edx.devstack.xqueue env TERM=$(TERM) /edx/app/xqueue/devstack.sh open
 
-xqueue-restart: ## Kill the XQueue development server. The watcher process will restart it.
-	docker exec -t edx.devstack.xqueue bash -c 'kill $$(ps aux | grep "manage.py runserver" | egrep -v "while|grep" | awk "{print \$$2}")'
+%-restart: ## Kill a service's Django development server. The watcher process should restart it.
```
alanoe: - it'd be better to create a new restart_django_web_server_in_container shell script for restarting Docker containers which run a Django development web server. Its input parameter would be docker_name and django_web_server_command and it could be reused for restarting CMS and LMS. Then, here you would only call it passing xqueue and runserver as parameters.

- Do all components which do not have an explicit <component>-restart make target (eg all except LMS and CMS) can be restarted with this command? One that comes to my mind is cs_comments_service, which is not a Django app, but rather a Sinatra (Ruby) app. With this new command, "make forum-restart" will print an error saying that "manage.py" was not found, which will confuse the user.

check.sh
```diff
@@ -0,0 +1,122 @@
...
+#
+# Exists 0 if successful; non-zero otherwise.
+#
+# Fails if no services specified.
+#
+# Note that passing in a non-existent service will not fail if there are
+# other successful checks.
+
+set -e
+set -o pipefail
```
alanoe, line 20: add "set -u", which checks undefined variables

alanoe (commented):
Could you move the file renaming to a separate patch so that we could see the cleanup diff and explain the reason of the file renaming?

Makefile
```diff
@@ -28,6 +28,40 @@
...
+MAIN_COMPOSE_FILE=-f docker-compose.yml
+XQUEUE_COMPOSE_FILE=-f docker-compose-xqueue.yml
+WATCHERS_COMPOSE_FILE=-f docker-compose-watchers.yml
+WATCHERS_COMPOSE_FILE_FOR_NFS=-f docker-compose-watchers-nfs.yml
+ANALYTICS_COMPOSE_FILE=-f docker-compose-analytics-pipeline.yml
+MARKETING_COMPOSE_FILE=-f docker-compose-marketing-site.yml
+
+# Supporting Docker Compose files to enable volume mounting (which allows us to use
+# source code from our host machine) as well as theming.
+SUPPORTING_COMPOSE_FILES=-f docker-compose-host.yml -f docker-compose-themes.yml
```
alanoe, line 40: minor: this var name does not make clear that is used for volume mounting. I recommend using COMPOSE_FILES_FOR_LOCAL_MOUNTS and removing SUPPORTING from the names of the vars below too.

Makefile
```diff
@@ -28,6 +28,40 @@
...
+# sets of services.
+STANDARD_COMPOSE_SUITE=$(MAIN_COMPOSE_FILE) $(SUPPORTING_COMPOSE_FILES)
+STANDARD_COMPOSE_SUITE_FOR_NFS=$(MAIN_COMPOSE_FILE) $(SUPPORTING_COMPOSE_FILES_FOR_NFS)
+STANDARD_COMPOSE_SUITE_FOR_SYNC=$(MAIN_COMPOSE_FILE)  $(SUPPORTING_COMPOSE_FILES_FOR_SYNC)
+XQUEUE_COMPOSE_SUITE=$(MAIN_COMPOSE_FILE) $(XQUEUE_COMPOSE_FILE) $(SUPPORTING_COMPOSE_FILES)
+ANALYTICS_COMPOSE_SUITE=$(STANDARD_COMPOSE_SUITE) $(ANALYTICS_COMPOSE_FILE)
+
+# All Docker Compose YAML files that contain service definitions.
+# Useful for Makefile targets like `dev.pull` and `down`
+ALL_SERVICE_COMPOSE_FILES=\
```
alanoe, line 53: minor: SERVICES_COMPOSE_FILES is shorter and conveys the same meaning

Makefile
```diff
@@ -28,6 +28,40 @@
...
+$(MAIN_COMPOSE_FILE) \
+$(XQUEUE_COMPOSE_FILE) \
+$(WATCHERS_COMPOSE_FILE) \
+$(ANALYTICS_COMPOSE_FILE) \
+$(MARKETING_COMPOSE_FILE)
+
+# All Docker Compose files (except those for NFS and Docker Sync).
+FULL_COMPOSE_SUITE=$(ALL_SERVICE_COMPOSE_FILES) $(SUPPORTING_COMPOSE_FILES)
```
alanoe, line 61: minor: this name is a little confusing, as a suite with NFS would be full too. I recommend FULL_COMPOSE_SUITE_WITH_LOCAL_MOUNTS instead to make it clearer it uses local mounts instead of NFS.

Makefile
```diff
@@ -28,6 +28,40 @@
...
+SUPPORTING_COMPOSE_FILES_FOR_NFS=-f docker-compose-host-nfs.yml -f docker-compose-themes-nfs.yml
+SUPPORTING_COMPOSE_FILES_FOR_SYNC=-f docker-compose-sync.yml
+
+# "Suites" of Docker Compose files that can be used together to run different
+# sets of services.
+STANDARD_COMPOSE_SUITE=$(MAIN_COMPOSE_FILE) $(SUPPORTING_COMPOSE_FILES)
+STANDARD_COMPOSE_SUITE_FOR_NFS=$(MAIN_COMPOSE_FILE) $(SUPPORTING_COMPOSE_FILES_FOR_NFS)
+STANDARD_COMPOSE_SUITE_FOR_SYNC=$(MAIN_COMPOSE_FILE)  $(SUPPORTING_COMPOSE_FILES_FOR_SYNC)
+XQUEUE_COMPOSE_SUITE=$(MAIN_COMPOSE_FILE) $(XQUEUE_COMPOSE_FILE) $(SUPPORTING_COMPOSE_FILES)
+ANALYTICS_COMPOSE_SUITE=$(STANDARD_COMPOSE_SUITE) $(ANALYTICS_COMPOSE_FILE)
```
alanoe, line 49: minor: I'd make it clearer that the first suite uses local mounts. Suggestions: STD_COMPOSE_SUITE_WITH_LOCAL_MOUNTS, STD_COMPOSE_SUITE_WITH_NFS, STD_COMPOSE_SUITE_WITH_DOCKER_SYNC

Makefile
```diff
@@ -28,6 +28,40 @@
...
+WATCHERS_COMPOSE_FILE=-f docker-compose-watchers.yml
+WATCHERS_COMPOSE_FILE_FOR_NFS=-f docker-compose-watchers-nfs.yml
+ANALYTICS_COMPOSE_FILE=-f docker-compose-analytics-pipeline.yml
+MARKETING_COMPOSE_FILE=-f docker-compose-marketing-site.yml
+
+# Supporting Docker Compose files to enable volume mounting (which allows us to use
+# source code from our host machine) as well as theming.
+SUPPORTING_COMPOSE_FILES=-f docker-compose-host.yml -f docker-compose-themes.yml
+SUPPORTING_COMPOSE_FILES_FOR_NFS=-f docker-compose-host-nfs.yml -f docker-compose-themes-nfs.yml
+SUPPORTING_COMPOSE_FILES_FOR_SYNC=-f docker-compose-sync.yml
```
alanoe, line 42: minor: I'd make it clearer that you're referring to docker-sync here (ie. SYNC => DOCKER_SYNC)

alanoe (commented):
clarfiy => clarify in commit msg

kdmccormick:
Thanks for the review @alanoe . I'm swamped with other work this week, but I'll hopefully be able to address your comments soon.

Makefile
kdmccormick: Not necessarily. This will just bring up the databases and dump them to a tar file, whether or not there's any provisioned data.

Makefile
kdmccormick: I agree that a solution that works across all containers would be nice. There is already a `devstack.sh` file in every container, with `start`, `exec`, and `open` (i.e. `shell`) commands -- `restart` could be added. Unfortunately, that's out of the scope of my work here.

The Makefile target as-is does not generalize to all the web services in Devstack. It only works for the services that are Python/Django and don't do the strange two-apps-in-one thing that edx-platform does.

I see two approaches I could take:

* Make it clear that this is Python/Django specific by naming the target `%-restart-django`. Then, alias existing `[lms|studio|xqueue]-restart` commands to `$xyz-restart-django` for backwards compatibility.

* Keep the target name generic (`%-restart`). For all services that do not support the standard implementation, either write a custom implementation like you are doing in #512 with `forum-restart`, or implement the target with `docker-compose restart $service`

Do either of those sound good?

Makefile
kdmccormick: As I thought about the comments you left regarding the names of the Docker Compile file Makefile variables, I realized that the whole set-up I was going for with the variable names was overly complicated.

I've pushed a new commit "Simplify Makefile handling of docker-compose files" that takes a new approach. It defines two new options--`SYNC_STRATEGY` and `DEFAULT_SERVICES`--that, respectively, define which file synchronization strategy (local-mount/nfs/docker-sync) and set of services are used by generic commands like `make dev.up` and `make dev.provision`. With those options in place, we can conditionally set a value for `DOCKER_COMPOSE_FILES` and use that variable everywhere in the Makefile.

Furthermore, I think this would enable paring down the default set of services in the future, like you suggested in your forum post.

Of course, these new options should be documented in RST, but I would like to get your agreement and merge before putting work into the formal documentation.

kdmccormick:
Hi @alanoe - Ready for another round of review when you have the chance.
* For all of your review comments, I've either addressed them with `--fixup` commits or replied in-thread.
* I split the health check changes into `Rename healthchecks.sh to check.sh` and `Clean up check.sh and allow it to take multiple args`. 
* I went ahead and used `shellcheck` on all the scripts I changed this PR.

Makefile
alanoe: I like the approach of using a restart-django-devserver Make target (notice I suggested a "devserver" suffix to make it clear what you're restarting).

Makefile
```diff
@@ -27,44 +26,49 @@
...
+-f docker-compose-watchers-nfs.yml
+endif
+ifeq ($(SYNC_STRATEGY),docker-sync)
+DOCKER_COMPOSE_FILES := \
+-f docker-compose.yml \
+-f docker-compose-sync.yml
+endif
+ifndef DOCKER_COMPOSE_FILES
+$(error SYNC_STRATEGY is set to $(SYNC_STRATEGY). Must be one of: local-mounts, nfs, docker-sync)
+endif
```
alanoe, line 102: if SYNC_STRATEGY must be one of local-mounts, nfs or docker-sync and docker-compose.yml is used by all of them, I suggest you initialize DOCKER_COMPOSE FILES with "-f docker-compose.yml" and add other yml files depending on the value of SYNC_STRATEGY. This will reduce code duplication.

Makefile
```diff
@@ -108,88 +112,73 @@ dev.clone: ## Clone service repos using HTTPS method to the parent directory
 dev.clone.ssh: ## Clone service repos using SSH method to the parent directory
 	./repo.sh clone_ssh
 
-dev.provision.services: ## Provision all services with local mounted directories
-	DOCKER_COMPOSE_FILES="$(STANDARD_COMPOSE_SUITE)" $(WINPTY) bash ./provision.sh
+dev.provision.services: ## Provision default services with local mounted directories
+	$(WINPTY) bash ./provision.sh $(DEFAULT_SERVICES)+e2e
```
alanoe, line 158: Why are you adding e2e here? It seems to me that either it should be added to DEFAULT_SERVICES (best option) or you must update documentation explaining why it is added specifically in this target.

options.mk
```diff
@@ -1,8 +1,8 @@
 # DEFAULT DEVSTACK OPTIONS
 # Included into Makefile and exported to command environment.
 # Defaults are listed in this file.
-# For configuring your local Devstack, create a file called
-# 'options.local.mk' and add your preferred overrides.
+# Local git-ignored overrides can be configured by creating `options.local.mk`.
+# Variables are set here with ?= to allow for overriding them on the command line.
```
alanoe, line 5: I think that adding support to a options.local.mk file with overrides should be done in a separate patch (before this one if it is a requirement).

alanoe (changes requested):
minor: I suggest FS_SYNC_STRATEGY instead of SYNC_STRATEGY

Makefile
alanoe: This looks better than the previous implementation, +1

alanoe:
@kdmccormick Thanks for the changes! I agree with almost everything in the new PR, I have just added a few comments.

Makefile
kdmccormick: @alanoe `make $service-restart-django-devserver` is a lot of typing, especially without target completion.

Would you be open to `$service-restart-devserver`? Your `make forums-restart` implementation in #512 could become `make forums-restart-devserver`.

I've pushed an new commit with my suggestion.

Makefile
kdmccormick: Yeah, `e2e` is a bit of a special snowflake, in that it's a provisioning target but it's not a service (essentially, it just gets some test data ready so that e2e tests can be run). I know this isn't the most elegant solution.

For this PR, I'll take your suggestion to add a comment of why it's being added here.

options.mk
kdmccormick: ~True. Reverted for now.~

kdmccormick: I remembering that I did this last week so that I could implement the old NFS targets as
```
dev.nfs.%:
	FS_SYNC_STRATEGY=nfs make dev.$*
```
I'll put the change in a separate commit.

Makefile
kdmccormick: Fixed this in `Simplify Makefile handling of docker-compose files`.

I prepended the `-f docker-compose.yml` after all the `if` branches so that the `ifndef DOCKER_COMPOSE_FILES` still works.

alanoe:
As soon as you do the small changes we have discussed, you already have my +1

kdmccormick:
Thanks for the thorough and helpful review @alanoe !

I pushed a final commit to make the CI log output cleaner (I didn't realize I needed `set -e` in TravisCI to exit early).

I'm going to ping the folks responsible for Discovery and eCommerce about these test failures in order to unblock our PRs.

kdmccormick:
I'm going to merge this with the eCommerce shard failing. We're seeing the exact same failure on `master`, and the failure occurs after the code paths that I feel need to be tested for this PR.

The eComm failure known of and is being looked into currently. It is likely an issue with https://github.com/edx/configuration associated with the ongoing configuration transition.

.travis.yml
```diff
@@ -10,12 +10,18 @@ branches:
     - master
 
 env:
-  - DEVSTACK_WORKSPACE=/tmp SHALLOW_CLONE=1 DEVSTACK='lms'
-  - DEVSTACK_WORKSPACE=/tmp SHALLOW_CLONE=1 DEVSTACK='registrar'
-  - DEVSTACK_WORKSPACE=/tmp SHALLOW_CLONE=1 DEVSTACK='ecommerce'
-  - DEVSTACK_WORKSPACE=/tmp SHALLOW_CLONE=1 DEVSTACK='edx_notes_api'
-  - DEVSTACK_WORKSPACE=/tmp SHALLOW_CLONE=1 DEVSTACK='credentials'
-  - DEVSTACK_WORKSPACE=/tmp SHALLOW_CLONE=1 DEVSTACK='analytics_pipeline'
```
feanil, line 18: There is a typo where this underscore was removed and we're seeing a breakage on master because a container with the name `analyticspipeline` doesn't exist.  https://travis-ci.org/github/edx/devstack/jobs/732970204  I'll make a pr to fix forward.

kdmccormick: @feanil I'm not sure if that's the issue. This was an intentional change, as `analyticspipeline` is the name of the service: https://github.com/edx/devstack/blob/master/docker-compose-analytics-pipeline.yml#L124

The CI build passed for analytics months after this merged.

PR #509, opened 2020-04-03
Fix bash syntax issue in 'make dev.nfs.provision.services'

kdmccormick:
Due to incorrect variable quoting.

Issue manifested as "/bin/sh: docker-compose.yml: command not found"


marcotuts:
This fixes the issue I was seeing, I can continue provisioning. (Separately I'm running into memory issues but that's for another day.)

marcotuts:
I actually don't know that I am allowed to review this / items in this repository.

matthugs (approved):

kdmccormick:
Discretionary merge without checks complete because they don't check the NFS rules anyway.

PR #511, opened 2020-04-07
Reset and clean git repository before checking out another branch

alanoe:
If an Open edX git repository already exists in your workstation, "sudo make dev.provision"
automatically downloads any new changes (git fetch) and changes the local branch to the
target one if needed (git checkout). However, a simple "git checkout" may fail if there
are files in the current branch which will conflict with the target branch. An even worse
problem is that some untracked files (eg migrations) may not conflict with the target
branch, but they may make the devstack creation fail. The solution is to delete all
uncommitted and untracked files in the git repo directory and this is implemented in
this patch using "git reset" and "git clean", respectively.

### DIFF
--- a/repo.sh
+++ b/repo.sh
@@ -140,6 +140,7 @@ _checkout_and_update_branch ()
         git pull origin ${OPENEDX_GIT_BRANCH}
     else
         git fetch origin ${OPENEDX_GIT_BRANCH}:${OPENEDX_GIT_BRANCH}
+        git reset --hard HEAD; git clean -dfx
         git checkout ${OPENEDX_GIT_BRANCH}
     fi
     find . -name '*.pyc' -not -path './.git/*' -delete
@@ -195,19 +196,23 @@ status ()
     cd - &> /dev/null
 }
 
-if [ "$1" == "checkout" ]; then
-    checkout
-elif [ "$1" == "clone" ]; then
-    clone
-elif [ "$1" == "clone_ssh" ]; then
-    clone_ssh
-elif [ "$1" == "whitelabel" ]; then
-    clone_private
-elif [ "$1" == "reset" ]; then
-    read -p "This will override any uncommited changes in your local git checkouts. Would you like to proceed? [y/n] " -r
-    if [[ $REPLY =~ ^[Yy]$ ]]; then
+if [ "$1" == "status" ]; then
+    status
+else
+    read -p "This will delete any uncommited changes in your local git repositories. Would you like to proceed? [y/n] " -r
+    if ! [[ $REPLY =~ ^[Yy]$ ]]; then
+        exit 0
+    fi
+
+    if [ "$1" == "checkout" ]; then
+        checkout
+    elif [ "$1" == "clone" ]; then
+        clone
+    elif [ "$1" == "clone_ssh" ]; then
+        clone_ssh
+    elif [ "$1" == "whitelabel" ]; then
+        clone_private
+    elif [ "$1" == "reset" ]; then
         reset
     fi
-elif [ "$1" == "status" ]; then
-    status
 fi


openedx-webhooks:
Thanks for the pull request, @alanoe! I've created [OSPR-4362](https://openedx.atlassian.net/browse/OSPR-4362) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

All technical communication about the code itself will still be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).

natabene:
@alanoe Thank you for your contribution. Are you submitting as an individual contributor or on behalf of your employer?

alanoe:
@natabene You're welcome. As an individual contributor.

natabene:
@alanoe Got it. In this case please sign our Individual Contributor Agreement (https://open.edx.org/wp-content/uploads/2019/01/individual-contributor-agreement.pdf) and send to legal@edx.org. Once they process it and you are added to our contributor list, we can review and merge your code again.

alanoe:
@natabene I had already signed it and sent it by email and was just waiting the answer from Legal. I got it today, so this blocker is out of the way.

natabene:
@alanoe Thank you for joining the Open edX community! We have added your user to our records and can now review your code going forward.

natabene:
@jmbowman Can you give this a look when you have a chance?

jmbowman (commented):
Could you also update the end of the file to make sure that the `checkout` and `clone` commands now prompt the user with a warning similar to the `reset` one?  Be sure to specifically mention that uncommitted files will be deleted, since even the `reset` command doesn't do that.  I currently keep a few useful files for local development use in some of my local checkouts, and I'd be pretty annoyed if those were deleted without warning.

alanoe:
@jmbowman Done

jmbowman:
Hi, sorry for delayed response.  Thanks for adding this, although `clone` needs the same warning (and now that I look closer, so do `clone_ssh` and `whitelabel`).

I looked at the test failures, and they're unrelated; there's work in progress elsewhere to fix those.

alanoe:
> Thanks for adding this, although `clone` needs the same warning (and now that I look closer, so do `clone_ssh` and `whitelabel`).

@jmbowman OK. I have refactored the code and resubmitted the PR. Could you please review it?

natabene:
@jmbowman Please have a look when you have a moment.

jmbowman:
Asking some other developers if they have any objections, I'll merge soon if none arise.  In the meantime, could you rebase from master?  I think the test failures on master should be fixed now.

repo.sh
```diff
@@ -130,6 +130,7 @@ _checkout_and_update_branch ()
         git pull origin ${OPENEDX_GIT_BRANCH}
     else
         git fetch origin ${OPENEDX_GIT_BRANCH}:${OPENEDX_GIT_BRANCH}
+        git reset --hard HEAD; git clean -dfx
```
feanil, line 143: Not sure I agree with `-x` option here.  If items are in `.gitignore` I don't think we need to delete them as they won't cause issues with checkout.  Some IDEs generate project files that are ignored by the repo but needed for the individual developers that I wouldn't want to necessarily remove.

alanoe: @feanil At least for me, the devstack creation creates *.pyc files owned by root which cannot be deleted when a later devstack creation tries to checkout another branch. I guess that these files are owned by root because they are created inside Docker containers. Example:

$ export OPENEDX_RELEASE=ironwood.master; make dev.provision
(completes OK)

$ unset OPENEDX_RELEASE; make dev.provision
./repo.sh clone_ssh
The [course-discovery] repo is already checked out. Checking for updates.
remote: Enumerating objects: 172, done.
remote: Counting objects: 100% (172/172), done.
remote: Compressing objects: 100% (23/23), done.
remote: Total 221 (delta 162), reused 150 (delta 149), pack-reused 49
Receiving objects: 100% (221/221), 151.60 KiB | 542.00 KiB/s, done.
Resolving deltas: 100% (178/178), completed with 76 local objects.
From https://github.com/edx/course-discovery
   604a29ba..725fdcb6  master     -> master
Switched to branch 'master'
find: cannot delete ‘./course_discovery/__pycache__/urls.cpython-35.pyc’: Permission denied
find: cannot delete ‘./course_discovery/__pycache__/wsgi.cpython-35.pyc’: Permission denied
find: cannot delete ‘./course_discovery/__pycache__/__init__.cpython-35.pyc’: Permission denied
(...)

The .pyc files could be deleted manually instead of using the -x parameter in "git clean", but I'm not sure if it is enough or if other build files must be deleted manually too. If we want to follow that approach, it'll require additional testing effort.

feanil: Won't this fix also run into the same error where, your git clean will get a `permission denied` for the same reason?

stvstnfrd: ### Actually :nerd_face: 
I agree the `-x` seems a bit harsh; this has the possibility of unexpectedly wiping out changes.
Even the `reset --hard` feels heavy-handed to me.

Personally, I look at warnings like "Local modifications on disk" to be a helpful heads-up, things to be handled before resetting code.
Just like I need to `stash`, `commit`, or otherwise resolve local changes prior to a `git rebase`, I'd expect the system from keeping me from blowing away my local changes.

Hopefully we're checking code in constantly (right?!), but losing uncommitted code is a worst-case scenario as a developer.

### But...

There is definitely a problem here.
In fact, I have a helper target in a `private.mk` locally that uses a form of this `sudo find [...]` [2] to wipe out all `.pyc` files from my `${DEVSTACK_WORKSPACE}`, prior to updating the repos.

So if the underlying issue is that the call to `find` [1] fails due to permissions, what's the best way for us to fix that?

Naive fix seems to just prepend `sudo ` to the `find [...] -delete` call, no?
@feanil Foresee issues with that?

The proper fix might be to ensure the files are created and made accessible to the real `${USER}`, but that's probably more involved than I care to get ;)


- [1] `find . -name '*.pyc' -not -path './.git/*' -delete`
- [2] `sudo find ../ -name '*.pyc' -not -path './.git/*' -delete`

xitij2000: I definitely think this is a pretty destructive change and shouldn't be applied automatically. I completely agree with this just being a warning to the developer. 

I'm never in such a hurry when running `dev.provision` that I'd risk losing potentially important changes and data in the edx-platform repository just to get the devstack process started. I don't know enough about bash or makefiles to know if it's possible but ideally I'd like to see it just record each error and move on and fail at the end instead of immediately.

feanil:
@alanoe thanks for the PR but I think we want to be very careful about when we clean out a repository this aggressively and doing it automatically as part of the standard flow seems risky to me.  I think if there were a new target to do this, that people could call explicitly it would be less controversial but I don't think we can accept the change as it stands today.

Since it's been stagnant for a little bit and is not a flow we want to pursue, I'm gonna close this PR for now.  Thanks again for the contribution, I'm sorry we couldn't land it as is.

openedx-webhooks:
@alanoe Even though your pull request wasn’t merged, please take a moment to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/devstack&entry.867055334=https://github.com/edx/devstack/pull/511&entry.1484655318=https://github.com/alanoe&entry.752974735=2020-04-07+18:42&entry.1917517419=2020-06-29+15:30&entry.2133058324=No) so we can improve your experience in the future.

PR #513, opened 2020-04-13
Remove direct assignment to the forward side of a many-to-many field

jmyatt:
this is not allowed in django v2.2


robrap (approved):
Assuming we get a green build. :)

robrap (approved):
I continue to approve all changes that get you to green.

PR #516, opened 2020-04-17
set a workable watchman timeout.

iloveagent57:


cpennington:
Why aren't we installing `pywatchman` via the `configuration` repo and the docker container definitions?

iloveagent57:
Well, because this was a hack :). I'll take a look at the config repo, since tests are still failing here, anyway.  @cpennington Is the docker-compose.yml file in here the right place to add that `DJANGO_PYWATCHMAN_TIMEOUT` env var, though?

cpennington:
I think configuration may also allow you to override environment variables for the docker containers, but I'm less sure on that.

iloveagent57:
So, while I agree that the configuration repo is a better place to make this change, I don't have faith in my abilities to turn around a change there quickly, and this PR can provide some real relief to developers right now.  Furthermore, I consider this PR to be generally low-risk.

Travis tests are currently failing for reasons unrelated to this PR.  Unless there's strong objection, I'd like to merge this tomorrow (2020-04-24).

jmbowman:
A cleaner fix for pywatchman installation is to update the dev requirements in each service to add pywatchman (we need to remove pyinotify from at least some of them anyway).  But this is faster to get merged and applied everywhere, we can come back and clean up the hack later.

For the environment variables, we could set them via `ENV` in each service's Dockerfile in the configuration repo, but that doesn't really seem any cleaner than what's here (the images are currently really only useful in combination with the docker-compose files here).

kdmccormick:
I respect wanting to give developers relief ASAP. At the same time, one reason Devstack is so messy is because it's full of fix-things-quickly-now style hacks. That being said, it seems fine to merge this and then come back and move it to config later. I think you'd just want to add it to [the xenial-common Dockerfile](https://github.com/edx/configuration/blob/master/docker/build/xenial-common/Dockerfile) as well as the bionic-common one. (Or maybe @jmbowman 's solution would cleaner)

My one request before merging: Can you have someone who runs Ubuntu confirm that this, at least, doesn't make things any worse for them? I would, but I'm OOO for most of today.

iloveagent57:
Alright, I've modified this PR to only set the timeout env. var.  I'll cook up some PRs for edx-platform and discovery to get pywatchman installed in dev settings, then I'll reach out to teams to make PRs for other repos (i.e. "fine, I'll do it the right way").

docker-compose.yml
```diff
@@ -244,7 +249,7 @@ services:
       - edxapp_lms_assets:/edx/var/edxapp/staticfiles/
 
   registrar:
-    command: bash -c 'source /edx/app/registrar/registrar_env && while true; do python /edx/app/registrar/registrar/manage.py runserver 0.0.0.0:18734; sleep 2; done'
+    command: bash -c 'source /edx/app/registrar/registrar_env  && while true; do python /edx/app/registrar/registrar/manage.py runserver 0.0.0.0:18734; sleep 2; done'
```
feanil: NIT: Looks like an extra space got left behind.

feanil (approved):

PR #525, opened 2020-04-28
Fix checkout for OPENEDX_RELEASE (CRI-193)

nedbat:
Some repos are not part of the Open edX releases, and so should not be
checked out.  If OPENEDX_RELEASE is set, then don't try to get those
repos.

### DIFF
--- a/repo.sh
+++ b/repo.sh
@@ -17,13 +17,9 @@ else
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


kdmccormick (approved):

schenedx (approved):
👍

PR #526, opened 2020-04-28
Make Registrar a not-by-default service

kdmccormick:
```
Registrar is not part of the current Open edX release.
Specifically, it does not have a `juniper.rc1` branch.
Thus, it cannot be part of the default Open edX devstack.
```
https://openedx.atlassian.net/browse/CRI-193
@edx/masters-devs 

The impact of this to Master's devs is that new devs would need to specifically check out Registrar and Program Manager, as well as adding the following to `options.local.mk`:
```Makefile
DEFAULT_SERVICES = $(DEFAULT_SERVICES)+registrar
```
If we merge this, I can add those instructions to our wiki or the Registrar README.


kdmccormick:
Closing in favor of #525 , for now at least.

PR #529, opened 2020-05-05
Remove unused Makefile variables from #505

kdmccormick:
When I implemented #505, I waffled between a couple different strategies for bundling together the different sets of docker-compose files.

While squashing commits, I accidentally left in the variable definitions for one of the old strategies. This PR removes them. All the variables are unused.

PR #530, opened 2020-05-05
Don't fail when MFEs are passed to provisioning script

kdmccormick:
PR #528 includes MFEs in the list of `DEFAULT_SERVICES`.

However, that will cause `make dev.provision` to fail, because `gradebook`, `program-console`, and `frontend-app-publisher` are not provisionable. The user would see:
```
Service 'gradebook' either doesn't exist or isn't provisionable. Exiting.
```

CI did not catch this, because it provisions specific services, instead of all default services.

This PR prevents those MFE service names from breaking the provisioning script.

### DIFF
--- a/provision.sh
+++ b/provision.sh
@@ -3,7 +3,7 @@
 # This script will provision the services specified in the argument list,
 # or all services if no arguments are provided.
 #
-# Specifying invalid services will cause the script to exit early.
+# Non-existant services will be ignored.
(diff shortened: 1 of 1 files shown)


feanil (approved):

kdmccormick:
@feanil Thanks for the quick review.

Ecommerce builds are timing out ~60% of the time. Do you feel okay with me merging this and https://github.com/edx/devstack/pull/529 without the ecom shard passing?

PR #536, opened 2020-05-15
Only update database if service is "installed"

stvstnfrd:
by virtue of being set inside `DEFAULT_SERVICES`.

This is needed to prevent errors when running with a custom
`options.local.mk`, specifically where a service is _omitted_ from
`DEFAULT_SERVICES`, but `make update-db` attempts to run the migrations
in a stopped container.

eg: set `DEFAULT_SERVICES=lms+studio+discovery+credentials+forum` and
see `make update-db` (without this patch) break trying to migrate ecommerce.

Now, we maintain a list of updatable database services and only migrate
them if the service is also present in `DEFAULT_SERVICES`.

### DIFF
--- a/options.mk
+++ b/options.mk
@@ -35,3 +35,16 @@ FS_SYNC_STRATEGY ?= local-mounts
 #       The current value was chosen such that it would not change the existing
 #       Devstack behavior.
 DEFAULT_SERVICES ?= lms+studio+ecommerce+discovery+credentials+forum+edx_notes_api+registrar+gradebook+program-console+frontend-app-publisher
+
+# List of all services with database migrations.
+# Services must provide a Makefile target named: $(service)-update-db
+# Note: This list should contain _all_ db-backed services, even if not
+# configured to run; the list will be filtered later against $(DEFAULT_SERVICES)
+DB_SERVICES ?= credentials discovery ecommerce lms registrar studio
+
+# List of Makefile targets to run database migrations, in the form $(service)-update-db
+# Services will only have their migrations added here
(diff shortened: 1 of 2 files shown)


options.mk
```diff
@@ -35,3 +35,8 @@ FS_SYNC_STRATEGY ?= local-mounts
 #       The current value was chosen such that it would not change the existing
 #       Devstack behavior.
 DEFAULT_SERVICES ?= lms+studio+ecommerce+discovery+credentials+forum+edx_notes_api+registrar+gradebook+program-console+frontend-app-publisher
+APP_SERVICES = $(subst +, ,$(DEFAULT_SERVICES))
+DB_SERVICES ?= credentials discovery ecommerce lms studio
+DB_MIGRATION_TARGETS = $(foreach db_service,$(DB_SERVICES),\
+	$(if $(filter $(db_service),$(APP_SERVICES)),\
+		$(db_service)-update-db))
```
kdmccormick: Nice Makefile hackery.

Can you add a comment block above these explaining what they're for? I'm hoping to keep `options.mk` as an approachable catalog of devstack's settings.

Lastly, can you add `registrar` to `DB_SERVICES`? It was originally omitted from `update-db` by mistake.

kdmccormick (approved):
Nice! LGTM, pending green build.

kdmccormick:
(Looks like the shards that failed are unrelated -- Good with me if you want to merge)

stvstnfrd:
@kdmccormick Thanks for the quick eyes on this!

After squashing, only two failing shards are also failing on master (`xqueue` and `analyticspipeline`), so I'll proceed to merge.

PR #544, opened 2020-06-09
Flesh out & reorganize Makefile iterface and README

kdmccormick:
_I plan to merge this on **Monday 7/13** unless changes are requested before that._

## Overview

The general purpose of this PR is to bring together a lot of changes that I and others have made in over the past year or so to make Devstack's interface more powerful, flexible and consistent. Essentially, it organizes the Makefile to make a "semi-official" (by nature of it being more documented and consistent) command interface. It also updates the README to describe that interface along with several other new features that have been added in the past year.

This PR is intended to have **no backwards-incompatible changes**.

Because of the large diff linecount, I recommend reviewing this PR commit-by-commit.

### 1: Fix alphabetization in docker-compose.yml 

Simple housekeeping change; nothing interesting.

### 2: Organize and rename Makefile targets 

Organize Makefile into sections for different types of commands (repo management, image management,
db management, container management, etc.).

Rename many targets to conform to a consistent naming scheme.

The old target names are either:
* Preserved in `compatability.mk`, which is included into the main Makefile, or
* Supported using new "prefix-form" trick, in which commands in the form ``dev.ACTION.SERVICE`` can be written as ``SERVICE-ACTION``. For example, ``make dev.logs.studio`` can still be spelled as ``studio-logs``.

Also, improve documentation in options.mk and standardize how options are assigned and overriden.

This commit should involve no breaking changes or new functionality.

### 3: Add some new ways to start services, manage databases 

New targets:
* dev.drop-db.%
* dev.up.attach.%
* dev.up.shell.%
* dev.up.with-watchers.%
* dev.dbshell
* More general dev.dbshell.%
* Corresponding prefix-form targets for some of the above.

### 4: Reorganize, de-dupe, and update README 

* Pare down number of top-level headings.
* Adds table-of-contents.
* Adds instructions for some new Makefile targets.
* Clarify some sections.
* Remove out-of-date info.
* Split not-generally-applicable sections out
  into their own RST files in the docs/ folder, with
  links out to them from the main README.
* Add section on options.mk
* Reference all tagets in 'dev.ACTION.SERVICE' form,
  but add section on 'SERVICE-ACTION' form.


Makefile
```diff
@@ -1,41 +1,62 @@
...
+        validate-lms-volume vnc-passwords
+
+# Colors for Make messages.
+RED="\033[0;31m"
+YELLOW="\033[0;33m"
+GREY="\033[1;90m"
+NO_COLOR="\033[0m"
 
-# Include local makefile with additional targets.
--include local.mk  # Prefix with hyphen to tolerate absence of file.
```
xitij2000, line 38: I see this disappearing again! Just tagging in case it's accidental :-)

kdmccormick: Actually just moving it this time (check out line ~140), but thanks for keeping an eye on it :wink:

xitij2000: Ah! Sorry for the false alarm.

cpennington:
Why, at this point, do we have the `dev.` prefix, when basically every make target uses it? Is there some meaningful non `dev.` target?

kdmccormick:
@cpennington Good question.

* Not breaking backwards-compatibility. This is the main reason in my mind.
* It can be ambiguous to users whether they should commands in the host shell vs. a container's shell. The `dev.` prefix may make it clearer that something should be run in the host shell in the devstack folder.

inventhouse:
> * It can be ambiguous to users whether they should commands in the host shell vs. a container's shell. The `dev.` prefix may make it clearer that something should be run in the host shell in the devstack folder.

I would rather solve this problem by standardizing to always run _all_ our `make` commands host-side and any that need to do something in a container just `docker-compose exec` whatever they need to do

kdmccormick:
@inventhouse In general, Devstack can be operated entirely by using host-side Make commands. I believe that the updated README does not suggest using any container-side Make commands.

However, nearly every edX repository has a Makefile, and much of our tooling relies on them having some standard targets (`requirements`, `upgrade`, `static`, et al). I'm not going to go around proposing to delete all of these Makefiles, especially when most of those repos are services and libraries that aren't in Devstack at all.

So, given that most repos are going to have their own Makefile for the forseeable future, I think there might be some merit to having the `dev.` prefix for targets in the Devstack Makefile. But that's not my main reason for standardizing on the `dev.` prefix in the PR...

Essentially, I want the Makefile targets to be consistent. I see three ways of doing this:

### Approach 1 (this PR)
* Change all remaining targets to `dev.ACTION.SERVICES` form.
* Alias all `SERVICE-ACTION` targets to `dev.ACTION.SERVICE`.
* Alias things like `make down` to `make dev.down`.

### Approach 2
* Change all targets to `ACTION.SERVICES`.
* Alias all `dev.ACTION.SERVICES` targets to `ACTION.SERVICES`.
* Alias all `SERVICE-ACTION` targets to `ACTION.SERVICE`.

### Approach 3
* Change all targets to `ACTION.SERVICES`.
* Don't alias anything, and send out a notice of the breaking changes.

### Approach null
* Don't do anything.

I chose Approach 1 because it seemed like the fewest number of changes in order to achieve consistency and not break backwards-compatibility. @cpennington @inventhouse do you have a preference between these?

inventhouse:
@kdmccormick Gotcha; yes, make devstack's targets consistent, then maybe we can push the others to adopt better consistency too.  I'm fine with the `dev.` prefix.

kdmccormick:
* [All shards passed](https://travis-ci.org/github/edx/devstack/builds/706938596) except ecommerce, which is failing due to a [known issue](https://openedx.atlassian.net/browse/CR-2138).
* #dev was notified about review period -- will be merging this shortly.

PR #549, opened 2020-06-24
Adding note to readme to suggest using NFS with mac

jinder1s:
Creating this to note that edx has mostly decided against docker-sync. Eventually, we should modify the docs to include more info on nfs setup. For now, I wanted to add the quick note so people might look at nfs option if they want to increase performance in devstack.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1087,10 +1087,10 @@ package versions installed.
 Performance
 -----------
 
-Improve Mac OSX Performace using nfs
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Improve Mac OSX Performance using nfs
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The option to use docker with nfs on mac was added recently. This can potentially increase preformance in mac osx. However, this option is still in testing phase. If you find any corrections that should be made, please start a PR with corrections.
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -1087,6 +1087,11 @@ package versions installed.
 Performance
 -----------
 
+**NOTE:**
+
+docker-sync is no longer actively supported. We now recommend using nfs on MacOS
+to greatly increase performance.
```
robrap: 1. Is this the recommendation?  Has someone run stats?  Does it greatly increase performance?
2. Should we remove docker-sync?

README.rst
```diff
@@ -1087,6 +1087,11 @@ package versions installed.
...
+
 Improve Mac OSX Performance with docker-sync
```
robrap, line 1096: Actually, rather than the note above, why not change this to `Improve Mac OSX Performance with nfs`?

robrap: @jinder1s: I should note that I don't like mucking with my devstack, and I would like to know when and if it is worth it to use `nfs`.  It is unclear if you are making that statement.  If you are just noting options, you can update this section to be `Improve Mac OSX Performance` and include a subsection for `docker-sync`, where you could note that it isn't supported, and a section (above that) for `nfs`, and note that people are experimenting with it, but we don't have stats on whether or not it actually improves performance (unless we do).

README.rst
jinder1s: We don't currently have stats on this. I will change this to not sound like a recommendation.

robrap (approved):

PR #550, opened 2020-06-24
dev.clone should use ssh by default

inventhouse:
Long-time annoyance for me, but both of the most recent engineering new-hires on my team have tripped over devstack cloning with http recently; no more!

I've done an org-wide search and the only places that needed updating were in devstack: https://github.com/search?q=org%3Aedx+%22dev.clone%22&type=Code

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@
 .PHONY: analytics-pipeline-devstack-test analytics-pipeline-shell \
         analyticspipeline-shell backup build-courses check-memory \
         create-test-course credentials-shell destroy dev.cache-programs \
-        dev.check dev.checkout dev.clone dev.clone.ssh dev.down dev.kill \
+        dev.check dev.checkout dev.clone dev.clone.https dev.down dev.kill \
         dev.nfs.setup devpi-password dev.provision \
         dev.provision.analytics_pipeline dev.provision.services \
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -122,10 +122,10 @@ dev.ps: ## View list of created services and their statuses.
 dev.checkout: ## Check out "openedx-release/$OPENEDX_RELEASE" in each repo if set, "master" otherwise
 	./repo.sh checkout
 
-dev.clone: ## Clone service repos using HTTPS method to the parent directory
+dev.clone.https: ## Clone service repos using HTTPS method to the parent directory
 	./repo.sh clone
 
-dev.clone.ssh: ## Clone service repos using SSH method to the parent directory
```
kdmccormick, line 128: How about keeping `dev.clone.ssh` as is, and then making `dev.clone` be an empty recipe with `dev.clone.ssh` as a prereq?

It's functionally the same thing, but would prevent any breakage for anyone who's now used to `make dev.clone.ssh`. Plus it feels like a more symmetrical interface.

kdmccormick (commented):
I like the change. To inform Open edX devs of the change, do you mind making a post on https://discuss.edx.org once the change is merged?

Makefile
inventhouse: I can also make the other uses explicit one way or the other

kdmccormick (approved):
:rocket:

PR #558, opened 2020-07-14
Re-establish the service provisioning order in provision.sh

kdmccormick:
Changing DEFAULT_SERVICES in the previous commit ended up
changing the order in which services are provisioned.
This broke provisioning for Discovery (at least) which depended
on LMS and Ecommerce being started and provisioned.

This commit mandates the order in provision.sh without relying
on that order being defined in DEFAULT_SERVICES.

@Dillon-Dumesnil

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,8 @@ env:
     - DEVSTACK_WORKSPACE=/tmp
     - SHALLOW_CLONE=1
   matrix:
-    - SERVICES=lms+discovery+forum
-    - SERVICES=lms+registrar
+    - SERVICES=discovery+lms+forum  # provision.sh should ensure LMS provisions first.
+    - SERVICES=registrar+lms
     - SERVICES=ecommerce
     - SERVICES=edx_notes_api
     - SERVICES=credentials
--- a/options.mk
+++ b/options.mk
@@ -59,6 +59,11 @@ FS_SYNC_STRATEGY ?= local-mounts
 # when no services are specified manually.
 # Should be a subset of $(EDX_SERVICES).
 # Separated by plus signs. Listed in alphabetical order for clarity.
+# WARNING: You may remove services from this list in order to make Devstack lighter,
+#          but beware that some services have implicit, undocumented dependencies on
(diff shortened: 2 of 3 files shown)


Dillon-Dumesnil (commented):
So I think this only captures part of the problem when looking at the discovery service (at least). It actually needs the LMS service to be provisioned (so it contains the default courses) in order for it to populate its own data. I imagine (but unconfirmed) that ecommerce could be similar.

So I think we want to update this to ensure LMS  (and maybe Studio? I always see them hand in hand) is provisioned first.

.travis.yml
```diff
@@ -15,8 +15,8 @@ env:
     - DEVSTACK_WORKSPACE=/tmp
     - SHALLOW_CLONE=1
   matrix:
-    - SERVICES=lms+discovery+forum
-    - SERVICES=lms+registrar
+    - SERVICES=forum+discovery+lms
+    - SERVICES=registrar+lms
```
kdmccormick, line 19: Will implicitly test that `provision.sh` fixes the provisioning order.

Dillon-Dumesnil (approved):
Assuming tests pass, looks good to me. Thanks for getting on top of this!

kdmccormick:
@Dillon-Dumesnil No problem! Thanks again for spotting it :100:

kdmccormick:
Both the test failures are known and [observed on master from last week](https://travis-ci.org/github/edx/devstack/builds/707469713). Once we diagnose [Dillon's new finding](https://edx-internal.slack.com/archives/C012DCDDUNR/p1594751745312400?thread_ts=1594647121.283600&cid=C012DCDDUNR), then we should either amend or merge this.

kdmccormick:
@Dillon-Dumesnil  and I are still seeing an error along the lines of:
```
Got an error creating the test database: (1044, "Access denied for user 'ecomm001'@'%' to database 'test_ecommerce'")
```
when running tests for ecom and discovery.

It needs to be addressed, but seems unrelated to this PR, so I'm going to merge this now.

PR #532, opened 2020-05-05
Multiple devstack releases with one machine

ztraboo:
This is the start of changing devstack to accommodate multiple releases (master, open-release-name) as mentioned here by @kdmccormick 
https://discuss.openedx.org/t/docker-devstack-multiple-releases-one-machine/1902/4

Replace all the hard-coded container references like:
`docker [exec|run|...] edx.devstack.<service>`

with calls to docker-compose:
`docker-compose [exec|run|...] <service>`

PR #377, opened 2018-12-18
Initialize comprehensive theming volumes for devstack_docker services…

ztraboo:
Setup comprehensive theming volumes for devstack_docker services to also include the watcher containers.

PR #376, opened 2018-12-18
FEDX-507 Fix issue with watcher assets for LMS and Studio services.

ztraboo:
Would like to make sure that the watchers for LMS and Studio services for docker_devstack are working properly. Error was occurring since /edx/app/edxapp/edx-platform/node_modules volume mapping was missing.
https://openedx.atlassian.net/browse/FEDX-507?focusedCommentId=353646&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-353646

Ref: https://github.com/edx/devstack/pull/375 (Closed - Todo: cherry pick to hawthorn).

PR #564, opened 2020-07-24
[Backport] Multiple devstack releases with one machine

ztraboo:
Committing this update for `juniper` release since it didn't exist yet.

@kdmccormick, @jmbowman, @arbrandes, @regisb

https://discuss.openedx.org/t/docker-devstack-multiple-releases-one-machine/1902
Multiple devstack releases with one machine.

This is in regards to this pull request
https://github.com/edx/devstack/pull/532
Updating volumes for `themes` to work with the `ironwood` release.

Applying these changes to get local devstack working for ironwood release.

https://github.com/edx/devstack/pull/376
https://github.com/edx/devstack/pull/377
Moved `COMPOSE_PROJECT_NAME` update to `options.mk` file.

Additional changes to get `devstack_docker` for Ironwood working multisite environment setup.


openedx-webhooks:
Thanks for the pull request, @ztraboo! I've created [OSPR-4854](https://openedx.atlassian.net/browse/OSPR-4854) to keep track of it in JIRA, where we prioritize reviews. Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

natabene:
@ztraboo Thank you for your contribution. Is this ready for edX to review?

ztraboo:
@natabene Yes ready for edX review.

Makefile
```diff
@@ -265,94 +265,100 @@ pull.xqueue: dev.pull.without-deps.xqueue+xqueue_consumer
...
-	docker exec -it edx.devstack.lms env TERM=$(TERM) /edx/app/edxapp/devstack.sh open
+	docker-compose exec lms env TERM=$(TERM) /edx/app/edxapp/devstack.sh open
 
 lms-watcher-shell: ## Run a shell on the LMS watcher container
-	docker exec -it edx.devstack.lms_watcher env TERM=$(TERM) /edx/app/edxapp/devstack.sh open
+	docker-compose exec lms_watcher env TERM=$(TERM) /edx/app/edxapp/devstack.sh open
 
 %-attach: ## Attach to the specified service container process to use the debugger & see logs.
-	docker attach edx.devstack.$*
+	docker-compose attach $*
```
regisb: As far as I know, `attach` is not a docker-compose command: https://docs.docker.com/compose/reference/

Makefile
```diff
@@ -265,94 +265,100 @@ pull.xqueue: dev.pull.without-deps.xqueue+xqueue_consumer
...
 xqueue_consumer-restart: ## Kill the XQueue development server. The watcher process will restart it.
-	docker exec -t edx.devstack.xqueue_consumer bash -c 'kill $$(ps aux | grep "manage.py run_consumer" | egrep -v "while|grep" | awk "{print \$$2}")'
+	docker-compose exec xqueue_consumer bash -c 'kill $$(ps aux | grep "manage.py run_consumer" | egrep -v "while|grep" | awk "{print \$$2}")'
 
 %-static: ## Rebuild static assets for the specified service container
-	docker exec -t edx.devstack.$* bash -c 'source /edx/app/$*/$*_env && cd /edx/app/$*/$*/ && make static'
+	docker-compose exec $* bash -c 'source /edx/app/$*/$*_env && cd /edx/app/$*/$*/ && make static'
+
+lms-sass: ## Rebuild sass assets for the LMS container
+	docker-compose exec lms bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform/ && paver compile_sass --debug --system=lms --theme-dirs /edx/app/edx-themes/edx-platform/ --themes my-theme'
```
regisb: Are we quite sure that we want to compile only the "my-theme" theme? This theme seems to be undocumented and unused elsewhere. (same below)

Makefile
```diff
@@ -265,94 +265,100 @@ pull.xqueue: dev.pull.without-deps.xqueue+xqueue_consumer
 validate: ## Validate the devstack configuration
 	docker-compose config
 
-backup: dev.up.mysql+mongo+elasticsearch ## Write all data volumes to the host.
-	docker run --rm --volumes-from edx.devstack.mysql -v $$(pwd)/.dev/backups:/backup debian:jessie tar zcvf /backup/mysql.tar.gz /var/lib/mysql
-	docker run --rm --volumes-from edx.devstack.mongo -v $$(pwd)/.dev/backups:/backup debian:jessie tar zcvf /backup/mongo.tar.gz /data/db
-	docker run --rm --volumes-from edx.devstack.elasticsearch -v $$(pwd)/.dev/backups:/backup debian:jessie tar zcvf /backup/elasticsearch.tar.gz /usr/share/elasticsearch/data
+backup: ## Write all data volumes to the host.
+	docker-compose run --rm --volumes-from -T mysql -v $$(pwd)/.dev/backups:/backup debian:jessie tar zcvf /backup/mysql.tar.gz /var/lib/mysql
```
regisb: There is something I don't quite understand. Some of the commands below use `docker-compose $(DOCKER_COMPOSE_FILES)` instead of just `docker-compose`. Can we avoid passing the list of docker-compose files to this command here?
(same below for the exec and run commands)

options.mk
```diff
@@ -8,6 +8,15 @@
 # Defaults to parent.
 DEVSTACK_WORKSPACE ?= $(shell pwd)/..
 
+# Name of Docker Compose project.
+# See https://docs.docker.com/compose/reference/envvars/#compose_project_name	# See https://docs.docker.com/compose/reference/envvars/#compose_project_name
+# Defaults to 'devstack'.	# Defaults to 'devstack-master' should OPENEDX_RELEASE not be defined.
+ifdef OPENEDX_RELEASE
+	COMPOSE_PROJECT_NAME ?= devstack-${OPENEDX_RELEASE}
+else
+	COMPOSE_PROJECT_NAME ?= devstack-master
```
regisb: In all docker-compose*.yml files above, we assume that the default for COMPOSE_PROJECT_NAME is "devstack", and not "devstack-master".

1. Which one is it?
2. Do we really need to specify `${COMPOSE_PROJECT_NAME:-devstack}` everywhere, since the default is defined here?

docker-compose-analytics-pipeline.yml
```diff
@@ -2,34 +2,42 @@ version: "2.1"
 
 services:
   namenode:
+    container_name: "edx.${COMPOSE_PROJECT_NAME:-devstack}.analytics_pipeline.namenode"
```
regisb, line 5: I don't understand why we need to modify the container name. Instead, we could simply pass the `--project-name` option to the docker-compose command https://docs.docker.com/compose/reference/overview/

regisb (commented):
I really like the fact that we are replacing most docker commands by docker-compose. I'd just like to clarify some questions, as I didn't take the time to run this PR locally myself.

Makefile
ztraboo: @regisb I'm ok with changing this. For the moment this was a placeholder for anyone's theme.

Makefile
ztraboo: @regisb Thanks for finding this. I'll revert back to `docker attach` command here.

Makefile
ztraboo: @kdmccormick Any advice on this?

Makefile
ztraboo: @regisb I think this is ok because where just talking to one service here (e.g. `mysql`). If we need to run a command where more than one service is involved then we need to use `docker-compose ${DOCKER_COMPOSE_FILES}` instead.

cc: @kdmccormick Any advice on this?

regisb: Indeed, it is not necessary to specify the docker-compose files as long as the service in argument is declared in docker-compose.yml. But if it's declared in another file, the command will fail. Take for example the `%-update-db`:

    docker-compose exec $* bash -c 'source /edx/app/$*/$*_env && cd /edx/app/$*/$*/ && make migrate'

This will fail if the proposed service is not declared in `docker-compose.yml`.

options.mk
ztraboo: I'll update that to show `devstack` for the `master` release. I forgot to do that. We had defined this base on release and removed the `container_name` for each service and then `COMPOSE_PROJECT_NAME` was used for that instance. We added back the `container_name`, however, I cannot recall why.

cc: @kdmccormick Can you help with the understanding why we needed to specify `COMPOSE_PROJECT_NAME`?

docker-compose-analytics-pipeline.yml
ztraboo: @regisb This is, so that, other services can reach each other on the Docker network I believe.

cc: @kdmccormick

Makefile
ztraboo: Fix this by adding in `dev.print-container.%` Make target back in to find unique container ID then use that to call `docker` command.

This make target was added here.
https://github.com/edx/devstack/pull/532/commits/59f812eed18e33f15c1616a0e2176d0aff079d26

docker-compose-analytics-pipeline.yml
ztraboo: More about that here.
https://github.com/edx/devstack/pull/532#discussion_r425225387

ztraboo: Making the container names backward compatible and more predictable by external services was the reason to name do it like this. I don't know enough to answer reason to use `--project-name` versus specifying the `container_name`.

https://github.com/edx/devstack/pull/532/commits/df73f42bdb4789e86cc62f4d3400475a209f634c

cc: @arbrandes, @kdmccormick

ztraboo:
@regisb Made another update https://github.com/edx/devstack/pull/564/commits/037c229865041edfd05a609b26f1fcaddcc33013 to align with what we did in https://github.com/edx/devstack/pull/532 for the `master` release. This adds a `Makefile > dev.print-container.%` target to find the ID of the container since the containers will be different per named release and therefore we need to identify which container we're using.

Also I provisioned this again and I didn't see any immediate errors.

options.mk
ztraboo: We added these additional comments too.

```
# Name of Docker Compose project.
# Volumes and network are namespaced based on this value,
# so changing it will give you a separate set of databases.
# See https://docs.docker.com/compose/reference/envvars/#compose_project_name
# If OPENEDX_RELAESE is defined, defaults to `devstack-${OPENEDX_RELEASE}`;
# otherwise, it defaults to `devstack`.
# Be sure to bring down services before changing the value of `COMPOSE_PROJECT_NAME`.`
```

ztraboo:
@regisb @kdmccormick @arbrandes Finally got around to adding documentation on how to use this multiple devstack same host here hhttps://github.com/edx/devstack/pull/564/commits/261bc6cefbfff33774cd08dabd1f1aab7a813fe5. I did this for the `master` branch here https://github.com/edx/devstack/pull/569.

Makefile
ztraboo: @regisb Probably the best person to ask about this is @kdmccormick. He made this change https://github.com/edx/devstack/commit/ec1e4ceb2051b90b156547fbea7b3710fe5a187b to add in `DOCKER_COMPOSE_FILES` conditionally.

Makefile
ztraboo: @regisb Any recommendations on what theme to use default from `./edx-platform/themes`? Maybe the `red-theme`?
- dark-theme
- edge.edx.org
- edx.org
- open-edx
- red-theme
- stanford-style

Makefile
ztraboo: @regisb Some of the commands that got executed with `docker exec` where talking to one container so it's using that `dev.print-container.%` to find the container id. Let me know if this makes sense or not.

Makefile
regisb: Why not compile all themes by omitting the `--themes` option?

Makefile
regisb: My point is that if you want to replace `docker exec` by `docker-compose exec`, you will need to pass the full list of docker-compose files to the `docker-compose` commands. I might be wrong -- a simple way to find out is for you to try out running `make xqueue-shell`. IMHO this command will not work anymore.

ztraboo:
@regisb Updated `README.rst` for documentation related to Juniper https://github.com/CUCWD/devstack/tree/ztraboo.juniper/multiple-releases-one-machine#how-do-i-run-multiple-named-open-edx-releases-on-same-machine. Please review when you have time.

docker-compose-analytics-pipeline.yml
regisb: As far as I understand, it should be sufficient to keep the container name as it is (`container_name: edx.devstack.analytics_pipeline.namenode`) and not add a network alias.

ztraboo: @regisb If we don't update the `container_name` per named release then it will always use `devstack` master. That's why we do that.

regisb: Why is that a problem? If the issue is name collision, then that's the problem addressed by setting `--project-name`. Maybe I am missing the underlying issue here.

ztraboo: @regisb You’re right in that if you set the —project-name that the container_name isn’t really necessary because it should name the containers according with the project name. If you read one of the comments above it talks about backward compatibility with external edX services that @arbrandes was concerned about. We added it back in for that reason only.

arbrandes: @regisb, @ztraboo, that's correct.  We need the network alias, otherwise reaching edX services from the container network would be very cumbersome.

Makefile
ztraboo: @regisb Makes sense. I removed the `--themes` option.

Makefile
ztraboo: @regisb Not sure that we need to specify`$(DOCKER_COMPOSE_FILES)` for this particular `docker-compose run` call on line 269 above since we're just running this on the `mysql` container only. @kdmccormick Can you comment on this?

ztraboo: @regisb @kdmccormick Thanks for mentioning testing out `make xqueue-shell`. I just realized that `xqueue` is not apart of the `$(DEFAULT_SERVICES)` https://github.com/edx/devstack/blob/open-release/juniper.master/options.mk#L32-L37 (same for `master` branch) and therefore it doesn't get provisioned initially with a `make dev.provision` here https://github.com/edx/devstack/blob/open-release/juniper.master/Makefile#L122-L127. Shouldn't `xqueue` be apart of the default edX services?

Also when I issue `make xqueue-shell` without defining `$(DOCKER_COMPOSE_FILES)` for the following code I get this error.
```
xqueue-shell: ## Run a shell on the XQueue container
	docker-compose exec xqueue env TERM=$(TERM) /edx/app/xqueue/devstack.sh open
```
Error
```
(venv)ztraboo@112 devstack % make xqueue-shell
docker-compose exec xqueue env TERM=xterm-256color /edx/app/xqueue/devstack.sh open
ERROR: No such service: xqueue
make: *** [xqueue-shell] Error 1
```

After adding `$(DOCKER_COMPOSE_FILES)` back in to this Makefile target I get this message now. It does appear that we need to include this `$(DOCKER_COMPOSE_FILES)` for the `make xqueue-shell` target since the `xqueue` Dockerfile service is defined in the `./docker-compose-xqueue.yml` file which is included in the `$(DOCKER_COMPOSE_FILES)` here https://github.com/edx/devstack/blob/open-release/juniper.master/Makefile#L52. 
```
(venv)ztraboo@112 devstack % make xqueue-shell
docker-compose -f docker-compose.yml -f docker-compose-host.yml -f docker-compose-themes.yml -f docker-compose-watchers.yml -f docker-compose-xqueue.yml -f docker-compose-analytics-pipeline.yml -f docker-compose-marketing-site.yml exec xqueue env TERM=xterm-256color /edx/app/xqueue/devstack.sh open
root@51c2e89ab2bd:/edx/app/xqueue/xqueue# 
```

I'll update the code to reflect this needed change. @kdmccormick I'll need you to review this update. This also needs to be updated for the `master` branch too for `devstack`.

ztraboo: @regisb @kdmccormick It also appears that the `make lms-shell` command works without specifying the `$(DOCKER_COMPOSE_FILES)` variable because the Docker service is defined in the `docker-compose.yml` file while `xqueue` service is not defined in that file.
```
# Shows docker command without specifying the `$(DOCKER_COMPOSE_FILES)` argument works because the Docker `lms` service is defined in the `docker-compose.yml` file.

lms-shell: ## Run a shell on the LMS container
```
https://github.com/edx/devstack/blob/open-release/juniper.master/Makefile#L315-L316

ztraboo: @regisb @kdmccormick I like the idea that we add `$(DOCKER_COMPOSE_FILES)` to all `docker-compose` calls to ensure that all Docker services that are not specified within the `docker-compose.yml` file get referenced. Let me know if that's a bad idea or not. For example, adding this argument to the `make lms-shell` didn't cause any issues that I can see initially.

ztraboo: @regisb @kdmccormick Yep I updated the `dev print-container.%` target to include the `$(DOCKER_COMPOSE_FILES)` argument and printing out the container id for the Docker `xqueue` service worked.
```
dev.print-container.%: ## Get the ID of the running container for a given service.
	@# Can be run as ``make --silent dev.print-container.$service`` for just ID.
	@echo $$(docker-compose $(DOCKER_COMPOSE_FILES) ps --quiet $*)
```
Works
```
(venv)ztraboo@112 devstack % make dev.print-container.xqueue
51c2e89ab2bd53be57e598906b176441f49f1432ad07645ca38b0757b0082e3f
```
Without the `$(DOCKER_COMPOSE_FILES)` argument results with
```
(venv)ztraboo@112 devstack % make dev.print-container.xqueue
ERROR: No such service: xqueue
```

ztraboo: @regisb @kdmccormick This fixes the `docker-compose exec` commands by passing in the `DOCKER_COMPOSE_FILES` argument for services not defined in `docker-compose.yml`. Please review this when you have time. I performed a `make down` then `make dev.provision` and received no errors.
https://github.com/edx/devstack/pull/564/commits/9056bfcab016e4eb3bd9ef468a2077dbe08e3f73

ztraboo: @kdmccormick With this commit for the `master` branch https://github.com/edx/devstack/pull/532/commits/4f2253814c4cf1b3249e7af822960634c7370130 for making a multiple devstack release on one machine you removed a lot of the `$(DOCKER_COMPOSE_FILES)` from several of the `docker-compose` commands. Can you explain why that occurred because with my recent commit with `juniper` I added this back in https://github.com/edx/devstack/commit/9056bfcab016e4eb3bd9ef468a2077dbe08e3f73. You also mentioned in your commit comment that "should fix the xqueue-shell and analyticspipeline-shell commands". It appears that you updated the provision scripts for those, however, I believe we need those `DOCKER_COMPOSE_FILES` added back so that the `docker-compose` command gets the Docker service configuration. Please advise if we should add `DOCKER_COMPOSE_FILES` back in for the `docker-compose exec` calls specifically. I release there are other `docker-compose` commands than `exec` so adding the compose files there too would also need an explanation. I'm trying to figure out if I should revert by recent juniper commit https://github.com/edx/devstack/commit/9056bfcab016e4eb3bd9ef468a2077dbe08e3f73 or create a new PR against `master` branch to add back in `COMPOSE_FILES` that was removed with your https://github.com/edx/devstack/pull/532/commits/4f2253814c4cf1b3249e7af822960634c7370130 commit.

cc: @regisb

docker-compose-analytics-pipeline.yml
regisb: @ztraboo @arbrandes I think there is a misunderstanding.

> if you set the —project-name that the container_name isn’t really necessary because it should name the containers according with the project name.

This not what I said. AFAIU the project name is only necessary to avoid docker container name conflicts.

> We need the network alias, otherwise reaching edX services from the container network would be very cumbersome.

If we keep the container name as it is (`container_name: edx.devstack.analytics_pipeline.namenode`), my theory is that containers will *not* need aliases.

Makefile
```diff
@@ -177,6 +177,10 @@ dev.nfs.%:
 
 dev.up.all: dev.up dev.up.watchers ## Bring up all services with host volumes, including watchers
 
+dev.print-container.%: ## Get the ID of the running container for a given service.
+	@# Can be run as ``make --silent dev.print-container.$service`` for just ID.
+	@echo $$(docker-compose $(DOCKER_COMPOSE_FILES) ps --quiet $*)
```
regisb, line 182: In the future we could even replace `docker-compose $(DOCKER_COMPOSE_FILES)` by `$(DOCKER_COMPOSE)`, but this is probably outside the scope of this PR.

Makefile
```diff
@@ -265,94 +269,100 @@ pull.xqueue: dev.pull.without-deps.xqueue+xqueue_consumer
 validate: ## Validate the devstack configuration
 	docker-compose config
```
regisb: Shouldn't this be `docker-compose $(DOCKER_COMPOSE_FILES) config`?

regisb:
@ztraboo I just need to understand the `container_name` and `alias` thing and then this will be ready to merge.

docker-compose-analytics-pipeline.yml
arbrandes: @regisb 

> If we keep the container name as it is (`container_name: edx.devstack.analytics_pipeline.namenode`), my theory is that containers will _not_ need aliases.

This might (_might_ - I'm not entirely sure) work as far as avoiding collisions between containers on devstacks with different project names, but it makes it difficult for external scripts to reach containers of running devstacks.  I forget the details, but I do remember trying to run different devstacks without network aliases, and running into all sorts of issues - including some internal to the Open edX devstack itself.

This would be simple to test, though, if anybody has the time.  Just try to fire up `blockstore` without network aliases into a devstack that has a project name different than the default.  I'm pretty sure it won't work.

Makefile
kdmccormick: Sorry for the delay @ztraboo @regisb; I was out-of-office last week.

Specifying the service list isn't necessary or even desirable, as [this part of the Makefile](https://github.com/edx/devstack/blob/open-release/juniper.master/Makefile#L39-L74) sets the `DOCKER_COMPOSE_FILES` environment variable for all targets and sub-shells of the Makefile. That variable purposefully inclues ALL relevant Compose files, so that you can run `make dev.up.xqueue` without XQueue being a default service.

When using local mounts, XQueue and Analytics should work without a hitch (if that's not the case, then something is amiss). Are you are using NFS or docker-sync? If so, there is not (and has never been) a Docker Compose declaration for XQueue with NFS/docker-sync.

natabene:
@regisb You are planning to review and merge this as a core committer, correct?

docker-compose-analytics-pipeline.yml
regisb: @ztraboo Can you please verify whether this works without network aliases?

regisb:
@natabene Yes, this is correct. There remain one outstanding issue that concerns network aliases.

provision-ecommerce.sh
```diff
@@ -4,6 +4,6 @@
...
-docker exec -t edx.devstack.ecommerce bash -c 'source /edx/app/ecommerce/ecommerce_env && python /edx/app/ecommerce/ecommerce/manage.py oscar_populate_countries --initial-only'
-docker exec -t edx.devstack.ecommerce bash -c 'source /edx/app/ecommerce/ecommerce_env && python /edx/app/ecommerce/ecommerce/manage.py create_demo_data --partner=edX'
+docker-compose $DOCKER_COMPOSE_FILES exec ecommerce bash -c 'source /edx/app/ecommerce/ecommerce_env && python /edx/app/ecommerce/ecommerce/manage.py create_or_update_site --site-id=1 --site-domain=localhost:18130 --partner-code=edX --partner-name="Open edX" --lms-url-root=http://edx.devstack.lms:18000 --client-side-payment-processor=cybersource --payment-processors=cybersource,paypal --client-id=ecommerce-key --client-secret=ecommerce-secret --from-email staff@example.com --discovery_api_url=http://edx.devstack.discovery:18381/api/v1/ --journals_api_url=http://journals.app:18606/api/v1/'
```
morenol: This is removing these arguments that are needed in Juniper:

`--backend-service-client-id=ecommerce-backend-service-key --backend-service-client-secret=ecommerce-backend-service-secret`

morenol:
```suggestion
docker-compose $DOCKER_COMPOSE_FILES exec ecommerce bash -c 'source /edx/app/ecommerce/ecommerce_env && python /edx/app/ecommerce/ecommerce/manage.py create_or_update_site --site-id=1 --site-domain=localhost:18130 --partner-code=edX --partner-name="Open edX" --lms-url-root=http://edx.devstack.lms:18000 --lms-public-url-root=http://localhost:18000 --client-side-payment-processor=cybersource --payment-processors=cybersource,paypal --sso-client-id=ecommerce-sso-key --sso-client-secret=ecommerce-sso-secret --backend-service-client-id=ecommerce-backend-service-key --backend-service-client-secret=ecommerce-backend-service-secret --from-email staff@example.com --discovery_api_url=http://edx.devstack.discovery:18381/api/v1/'
```

Makefile
ztraboo: @regisb Yep I think that was done with the `master` release. I agree with you to wait for this PR. `$(COMPOSE_FILE)` I believe.

Looking at this https://docs.docker.com/compose/reference/envvars/#compose_file it appears that you can specify this `COMPOSE_FILE` environment variable and it will `"If not provided, Compose looks for a file named docker-compose.yml in the current directory and then each parent directory in succession until a file by that name is found."` It looks like @kdmccormick made that change here https://github.com/edx/devstack/pull/532/commits/4f2253814c4cf1b3249e7af822960634c7370130, however, I think this didn't get updated with Juniper release so that is why we need to specify `${DOCKER_COMPOSE_FILES)` for the Juniper release.

Makefile
ztraboo: @regisb Yes. It appears that leaving out `$(DOCKER_COMPOSE_FILES)` caused the `docker-compose config` to not include the **volume** information for services within the `docker-compose.yml` file, as well as, excluding additional services like `docker-compose-{analytics-pipeline,watchers}.yml` to name a few. 

Fixed this with https://github.com/edx/devstack/pull/564/commits/3808fa3b3adb870a623a14c3f7a76552406e5bcc

docker-compose-analytics-pipeline.yml
ztraboo: @regisb I'll re-provision Juniper now to test removal of network aliases, however, I think that I did this one time before and it was successful without them. @arbrandes just had a concern mentioned ^ with external services that I was not aware of at the time I made this multi-release changes.

It does appear that you can specify the `--project-name` to the `docker-compose` command or specify the `COMPOSE_PROJECT_NAME` as indicated in this pull request.
https://github.com/edx/devstack/pull/564/commits/9a55c5bf1717753960601627962c5a6cb046fd87#diff-86df7081aaade96499ca968414b07e48R14-R17

Indication of the above is here https://docs.docker.com/compose/#multiple-isolated-environments-on-a-single-host.

I'm going to comment out network aliases and set the container_names back to default without `.${COMPOSE_PROJECT_NAME:-devstack}.` to see what re-provisioning does. We may need to update `master` if this works and everyone agrees for this change.

cc: @arbrandes @kdmccormick

ztraboo: @regisb So far in provisioning I've had to remove the `container_name` from the Docker services to account for the container automatic naming using the `$(COMPOSE_PROJECT_NAME)` as indicated https://docs.docker.com/compose/#multiple-isolated-environments-on-a-single-host

Here is the error that I get from the LMS container saying that it cannot find the `django.db.utils.OperationalError: (2003, "Can't connect to MySQL server on 'edx.devstack.mysql' (111)")` container. This is why we added the network alias in.
https://gist.github.com/ztraboo/ed88ae663dab47a33d43a252afbb4e3f#file-dev-provision-sh-L414

It appears that we could manually define the container_names with `.${COMPOSE_PROJECT_NAME:-devstack}.` or have `docker-compose` create this name automatically for us using the `$(COMPOSE_PROJECT_NAME)` environment variable. I remember talking with @kdmccormick about this and I think we decided to leave it in. It seems like with the `--project-name` option to `docker-compose` or defining `$(COMPOSE_PROJECT_NAME)` environment variable will do this for us automatically.

I think we could just leave this as is unless you see otherwise. The network aliases are a must to allow services like the LMS to get to other network containers like the MySQL as shown above in the provisioning log error. We could remove the explicit container_name `.${COMPOSE_PROJECT_NAME:-devstack}.`, however, I think `docker-compose` is doing this automatically for us with the same name. Please advise.

cc: @arbrandes

provision-ecommerce.sh
ztraboo: @morenol I accepted your recommended change. Thanks for letting me know.

Change: https://github.com/edx/devstack/pull/564/commits/0ae1c8a256b7814e4d2c5aec49fd2cb3ea5cc7b9

docker-compose-analytics-pipeline.yml
ztraboo: @regisb It appears to do the following with the container_names.
- When Docker service `container_name` is explicitly defined (like it is https://github.com/edx/devstack/pull/564/files#diff-4e5e90c6228fd48698d074241c2ba760R281) with `${COMPOSE_PROJECT_NAME:-devstack}` then the name of the container looks like `edx.devstack-juniper.master_lms`. So there is a prefix `edx.` and `.` between `juniper` and `master` with no `_1` postfix applied. This allows the service container to be reachable more easily by name rather than the two alternatives below.

- When defining `COMPOSE_PROJECT_NAME` like so https://github.com/edx/devstack/commit/9a55c5bf1717753960601627962c5a6cb046fd87#diff-86df7081aaade96499ca968414b07e48R14-R17 through the Makefile `options.mk` file with no Docker service `container_name` defined the name of the container looks like `devstack-junipermaster_lms_1`. 

- When defining `export COMPOSE_PROJECT_NAME="juniper-test"` within the `.envrc` file and issuing `direnv allow` the name of the container looks like `juniper-test_lms_1`. This is meant more if you wanted to define a specific container_name from the default edX multiple versions same machine but it can be accomplished by predefining the `COMPOSE_PROJECT_NAME` environment variable ahead of making the `make dev.provision` call.

The above is all about Docker service container names and does prevent overriding existing release containers with another release. However, I still think we need to include the Docker service network alias to not prevent the containers on the Docker network from seeing each other due to some legacy code calls (e.g. `edx.devstack.mysql`) for example.

If we left the container names as is (e.g. `edx.devstack.mysql`) then you're correct then the provisioning script would not have failed with `django.db.utils.OperationalError: (2003, "Can't connect to MySQL server on 'edx.devstack.mysql' (111)")`. However I don't think we can leave the container names as before due to https://docs.docker.com/compose/#multiple-isolated-environments-on-a-single-host over changing the Docker service container_names automatically or explicitly to have new containers per release name.

Due to changing Docker container_names for this multiple isolated environments on single host I think the network aliases are required to prevent errors with provisioning and being able to reach network services from an existing container.

If you read the description for ALIASES here https://docs.docker.com/compose/compose-file/compose-file-v2/#networks it says `Aliases (alternative hostnames) for this service on the network. Other containers on the same network can use either the service name or this alias to connect to one of the service’s containers.` Adding the alias prevents us from having to update the legacy code calls where the original service container name was used (e.g. `edx.devstack.mysql`).

Anyway hope that helps explain why. Sorry for the long discussion.

cc: @kdmccormick @arbrandes

ztraboo:
@regisb Let me know what else needs to be done if anything. I made a comment above about `container_name` and `aliases` here https://github.com/edx/devstack/pull/564#discussion_r468198768.

docker-compose-analytics-pipeline.yml
regisb: Right, my suggestion stemmed from some misunderstanding of the impact of `container_name`. I believed that the name of a docker container was of the form "{project_name}_{container_name}_1" and that a container could be addressed inside the docker network as "{container_name}". The latter is correct, but the former isn't.
As a consequence, I agree that it is necessary to change the container_name to include the compose project name, as well as define a network alias.

Sorry about the prolonged conversation guys, and thanks for bearing with me @ztraboo  @arbrandes.

regisb (approved):
This LGTM. @ztraboo can you please rebase on top of master? Also, you should replace the commit that updates the README.rst and add the commit with the better phrasing from PR #569. Maybe it would make sense to merge #569 first and then backport that commit in this PR.

ztraboo:
@regisb Just made this commit https://github.com/edx/devstack/pull/564/commits/b73a1479da321e226dabbd202e9a4bea4a6e2965 to update the `README.rst` file for this PR instead of waiting for https://github.com/edx/devstack/pull/569 to be merged and backporting the change to Juniper from `master` branch.

> This LGTM. @ztraboo can you please rebase on top of master? 

I rebased on top of `open-release/juniper.master`.

> Also, you should replace the commit that updates the README.rst and add the commit with the better phrasing from PR #569. Maybe it would make sense to merge #569 first and then backport that commit in this PR.

Instead of removing the `README.rst` commits in this PR I just updated it to reflect what was done with https://github.com/edx/devstack/pull/569 and changed commands for stopping containers to ``make stop.all`` to reflect Juniper commands as indicated here https://github.com/edx/devstack/pull/564/commits/b73a1479da321e226dabbd202e9a4bea4a6e2965. This was updated to ``make dev.stop`` in the `master` release.

regisb:
This looks ready to merge! @ztraboo would you like to squash some of your commits or should I merge this as it is?

ztraboo:
@regisb 

> This looks ready to merge! @ztraboo would you like to squash some of your commits or should I merge this as it is?

I can squash all commits if needed. Would it be helpful to some to see what changes were done rather than all together as one commit? Does squashing commits make it easier to see what this PR changes were done in one commit rather than separated out? Please advise.

regisb:
Personnally, I prefer to squash together the commits that impact the same features or things. For instance, I would squash the commits that affect the README. If one particular commit fixes a bug that is not related to the other features from this PR, then this commit should stand on its own and not be squashed. Commits that all pertain to a single feature, such as "modify implementation according to regisb's bloody nitpickings" (:wink:) should be squashed in a single features.

ztraboo:
@regisb Thanks for that explanation. I squashed some commits. Ready to merge.

openedx-webhooks:
@ztraboo 🎉 Your pull request was merged!

Please take a moment to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/devstack&entry.867055334=https://github.com/edx/devstack/pull/564&entry.1484655318=https://github.com/ztraboo&entry.752974735=2020-07-24+02:36&entry.1917517419=2020-08-17+12:46&entry.2133058324=Yes) so we can improve your experience in the future.

docker-compose-analytics-pipeline.yml
arbrandes: No worries @regisb.  I'm actually very glad that you're taking this amount of care with the release branches. :+1:

PR #565, opened 2020-07-24
[BD-18] Fixes to run e2e tests on devstack

ziafazal:
Fixes to run e2e tests on devstack.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@
         dev.up.with-watchers dev.validate e2e-tests e2e-tests.with-shell \
         feature-toggle-state help requirements selfcheck upgrade upgrade \
         up-marketing-sync validate-lms-volume vnc-passwords
-        
+
 # Load up options (configurable through options.local.mk).
 include options.mk
 
@@ -218,7 +218,7 @@ dev.provision: dev.check-memory ## Provision dev environment with default servic
 	# it's just a way to tell ./provision.sh that the fake data for end-to-end
 	# tests should be prepared.
 	$(WINPTY) bash ./provision.sh $(DEFAULT_SERVICES)+e2e
-	make dev.stop	
+	make dev.stop
 
 dev.provision.%: ## Provision specified services.
 	echo $*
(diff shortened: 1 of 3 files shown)


ziafazal:
@feanil @kdmccormick could you please review this PR?

Makefile
```diff
@@ -101,6 +101,10 @@ ifndef COMPOSE_FILE
 $(error FS_SYNC_STRATEGY is set to $(FS_SYNC_STRATEGY). Must be one of: local-mounts, nfs, docker-sync)
 endif
 
+ifndef COMPOSE_PROJECT_NAME
+COMPOSE_PROJECT_NAME = devstack
+endif
+
```
kdmccormick: Does this have any effect? Ideally this should already be handled in [options.mk](https://github.com/edx/devstack/blob/master/options.mk#L35-L39).

Makefile
```diff
@@ -578,10 +582,11 @@ _expects-database.%:
 ########################################################################################
 
 e2e-tests: ## Run the end-to-end tests against the service containers.
-	docker run -t --network=${COMPOSE_PROJECT_NAME:-devstack}_default -v ${DEVSTACK_WORKSPACE}/edx-e2e-tests:/edx-e2e-tests -v ${DEVSTACK_WORKSPACE}/edx-platform:/edx-e2e-tests/lib/edx-platform --env-file ${DEVSTACK_WORKSPACE}/edx-e2e-tests/devstack_env edxops/e2e env TERM=$(TERM)  bash -c 'paver e2e_test --exclude="whitelabel\|enterprise"'
+	docker run -t --network=$(COMPOSE_PROJECT_NAME)_default -v $(DEVSTACK_WORKSPACE)/edx-e2e-tests:/edx-e2e-tests --env-file $(DEVSTACK_WORKSPACE)/edx-e2e-tests/devstack_env edxops/e2e env TERM=$(TERM) bash -c 'paver e2e_test'
```
kdmccormick, line 581: ~Regarding the `:-devstack` addition, I have the same comment as above about `options.mk`.~
Nevermind this comment; I read the diff backwards.

kdmccormick (commented):
Thanks for the PR. I'm excited to see e2e tests working in Devstack again!

I only have a couple of comments; I haven't tested this. @feanil will you be reviewing this, or did you want me to do a fuller review?

feanil:
@kdmccormick if you can do a further review, that would be great but I'm happy to take it if you're strapped for time.

Makefile
kdmccormick: I was able to get the tests working without this change. I believe that your other fixes are sufficient to get the tests running again.

Makefile
```diff
@@ -578,10 +582,11 @@ _expects-database.%:
 ########################################################################################
 
 e2e-tests: ## Run the end-to-end tests against the service containers.
```
kdmccormick:
```suggestion
e2e-tests: dev.up.lms+studio ## Run the end-to-end tests against the service containers.
```
How about making both this and `e2e-tests.with-shell` targets depend on `dev.up.lms+studio`? They break if those services haven't been brought up, right?

kdmccormick (changes requested):
Generally looks good!

I was able to run tests, but the ones I tried to run have failed. For example:
```
root@fa406f207ae7:/edx-e2e-tests# paver e2e_test common/test_html_component.py
```
gives:
```
bok_choy.page_object.PageLoadError: Message: Timed out waiting to load page 'CourseOutlinePageExtended(<browser>)' at URL 'http://studio.devstack.edx:18010/course/course-v1:edX+E2E-101+course'
```
That being said, at least the tests run now, which is better than before, so I'm happy to give my :+1: on this once my comments are addressed.

ziafazal:
@kdmccormick Thanks for reviewing. I have addressed feedback could you please take another look?

kdmccormick (approved):
Looks good, with the caveat that this just gets test to run for me, but not to pass.

feanil:
Ecom failure looks like it's happening on master as well.

PR #567, opened 2020-07-24
Making edx-mktg slightly easier to set up

matthugs:
was following
https://openedx.atlassian.net/wiki/spaces/ENG/pages/159162183/Marketing+Site
but was surprised to find that repo didn't have any required volumes
mounted.

working on some simple testing of
https://github.com/edx/edx-mktg/pull/3017

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,7 @@ COMPOSE_FILE := $(COMPOSE_FILE):docker-compose-watchers.yml
 COMPOSE_FILE := $(COMPOSE_FILE):docker-compose-xqueue.yml
 COMPOSE_FILE := $(COMPOSE_FILE):docker-compose-analytics-pipeline.yml
 COMPOSE_FILE := $(COMPOSE_FILE):docker-compose-marketing-site.yml
+COMPOSE_FILE := $(COMPOSE_FILE):docker-compose-marketing-site-host.yml
 endif
 
 # Files for use with Network File System -based synchronization.


matthugs:
It looks like the testing setup requires some things that aren't currently automated as part of the setup process to be automated as part of the setup process. I.e. the condition being tested for isn't true in the standard devstack use-case. I'm not going to correct that, but perhaps the owners should. I'll leave this open and flag it to the website team, since they currently own this repo.

kdmccormick (approved):
Sorry; was on PTO. Nice fix.

Approving this, pending a fix for the `marketing` shard tests.

The test failures on the `ecommerce` shard are known and can be ignored for this PR.

morenol:
Could you change the https://github.com/edx/devstack/blob/master/docker-compose-marketing-site.yml file to use the latest image of the docker image and not the thing that is configured in the env var OPENEDX_RELEASE?, The only tag available in docker hub is 'latest' https://hub.docker.com/r/edxops/edx-mktg/tags

kdmccormick:
Hey @morenol . The edX.org marketing site is closed-source, and that image wouldn't be very useful to anyone outside of the company.

kdmccormick:
@morenol I'm realizing that the README doesn't make that fact very clear -- here's a PR to make it more explicit: https://github.com/edx/devstack/pull/595

morenol:
> Hey @morenol . The edX.org marketing site is closed-source, and that image wouldn't be very useful to anyone outside of the company.

Ok, I didnt pretend to use that, I just saw the error when I ran all the test suite in the juniper branch setting the OPENEDX_RELEASE environment variable as juniper.master.

kdmccormick:
@morenol Ohhh, good catch, I didn't even consider that. Honestly, if possible, we shouldn't even be running tests for Marketing (or any other service in the `non_release_repos` array in clone.sh) when `OPENEDX_RELEASE` is set. 

If you can think of a way to make it so non-OpenedX services don't break tests when `OPENEDX_RELEASE` is set, then I would be happy to review that PR.

jmbowman:
In light of https://github.com/edx/devstack/pull/659 , should this just be closed now?

PR #598, opened 2020-08-13
draft: feat: generalize dev.restart-devserver.% to MFEs

kdmccormick:
**Putting this on hold until I can make it work with newer frontend-build versions a well**

`make dev.restart-devserver.%` does not currently work on MFEs. This intends to fix that by writing a general `make dev.restart-devserver.%` that works on all of our services.


kdmccormick:
Closing - will be focusing my time on Tutor instead.

PR #602, opened 2020-08-26
Separate .tox directories into separate mounts to avoid cross-contaim…

cpennington:
…ination with host filesystems


inventhouse (approved):
Confirmed that this fixed the `import Crypto` issue for me

PR #607, opened 2020-09-08
Create an ES7 container for devstack testing.

dianakhuang:


docker-compose.yml
```diff
@@ -61,14 +61,14 @@ services:
...
 
   # This is meant to be used to test ES upgrades so that we do not have to upgrade all of our services to ES5 at once.
-  elasticsearch-5:
-    container_name: "edx.${COMPOSE_PROJECT_NAME:-devstack}.elasticsearch-5"
-    hostname: elasticsearch-5.devstack.edx
-    image: elasticsearch:5.6.16
+  elasticsearch-7:
+    container_name: "edx.${COMPOSE_PROJECT_NAME:-devstack}.elasticsearch-7"
+    hostname: elasticsearch-7.devstack.edx
+    image: elasticsearch:7.9.1
```
stvstnfrd: Just to flag, the `cs_comments_service` PR is using `v7.8.0`; should we keep these in sync?

https://github.com/edx/cs_comments_service/pull/327/files#diff-184d171bba45cc420fe2efaf16c0c975R6

davidjoy: Adding some color here - the gem version is also 7.8.0 in the other PR.  I'm not sure to what extent the gem/image versions need to be kept in sync, though I'd hope the image could move ahead in minor/patch versions.

davidjoy: According to this the latest version of the gem is 7.9.0: https://github.com/elastic/elasticsearch-ruby

dianakhuang: I will roll it back for now. This branch has mostly just been me hacking at things to see if I can get them to work. 😅

PR #612, opened 2020-09-18
Add elasticsearch7 to devstack

stvstnfrd:
This looks to include some seemingly unrelated changes, presumably due to a recent rebase on my branch.

PR #621, opened 2020-09-24
Upgrade notes to elasticsearch7

stvstnfrd:
Note: This depends on https://github.com/edx/devstack/pull/612 .
This branch will be rebased and the PR re-pointed to `master` following #612 merging to `master`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -246,7 +246,7 @@ services:
     hostname: edx_notes_api.devstack.edx
     depends_on:
       - devpi
-      - elasticsearch
+      - elasticsearch7
       - mysql
     image: edxops/notes:${OPENEDX_RELEASE:-latest}
     networks:
@@ -264,7 +264,7 @@ services:
       DB_USER: "notes001"
       DJANGO_WATCHMAN_TIMEOUT: 30
       ENABLE_DJANGO_TOOLBAR: 1
-      ELASTICSEARCH_URL: "http://edx.devstack.elasticsearch:9200"
+      ELASTICSEARCH_URL: "http://edx.devstack.elasticsearch7:9200"
 
   forum:
     command: bash -c 'source /edx/app/forum/ruby_env && source /edx/app/forum/devstack_forum_env && cd /edx/app/forum/cs_comments_service && bundle install && while true; do ruby app.rb -o 0.0.0.0 ; sleep 2; done'


docker-compose.yml
```diff
@@ -264,7 +264,7 @@ services:
       DB_USER: "notes001"
       DJANGO_WATCHMAN_TIMEOUT: 30
       ENABLE_DJANGO_TOOLBAR: 1
-      ELASTICSEARCH_URL: "http://edx.devstack.elasticsearch:9200"
+      ELASTICSEARCH_URL: "http://edx.devstack.elasticsearch7:9200"
```
stvstnfrd, line 267: This is one change that had _not_ already been in the previous WIP PR, but it _seems_ useful?

stvstnfrd (commented):
@dianakhuang Who is best to confirm this works for notes?

I'm averse to merging it myself, as I haven't tested notes at all.

docker-compose.yml
kdmccormick: I believe you'll need this; otherwise, notes will keep pointing at the old ES container.

dianakhuang: It doesn't hurt, but there is a new setting used for ES, which is why I didn't catch it before.

dianakhuang:
I'm having trouble with my docker volumes, but it seems like this works otherwise. I would say we should merge this and https://github.com/edx/edx-notes-api/pull/203 at the same time.

dianakhuang (approved):

docker-compose.yml
regisb: Unless I'm mistaken, you will need to modify the hardcoded `ELASTICSEARCH_DSL['default']['hosts'] = 'edx.devstack.elasticsearch:9200'` in the devstack settings: https://github.com/edx/edx-notes-api/blob/6e339b1bbbb421efe52cb7737be4b61651a594a4/notesserver/settings/devstack.py#L16

regisb (changes requested):

dianakhuang:
@regisb yup, you are correct. I have a PR up for that: https://github.com/edx/edx-notes-api/pull/203

davidjoy (approved):

regisb (changes requested):
Would you mind documenting the switch to ES7 in the README? It should include a "Temporary fixes" section which should also describe the progressive switch to MySQL 5.7. This is necessary to inform devstack users of this unexpected behaviour.

EDIT: also, I'd like to see CI run on this PR once https://github.com/edx/edx-notes-api/pull/203 has been merged.

dianakhuang:
We're seeing `Failed to establish a new connection: [Errno 111] Connection refused` after merging the notes PR, so something is still broken in notes. I created https://github.com/edx/devstack/pull/626 to unblock provisioning until we can get ES7 working in devstack here.

EDIT: It is trying to access the correct host: ` http://edx.devstack.elasticsearch7:9200/notes_index`

dianakhuang:
Closing in favor of: https://github.com/edx/devstack/pull/630

PR #626, opened 2020-09-29
Temporarily remove notes from the list of all services

dianakhuang:
to unblock provisioning.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ env:
     - SERVICES=discovery+lms+forum  # provision.sh should ensure LMS provisions first.
     - SERVICES=registrar+lms
     - SERVICES=ecommerce
-    - SERVICES=edx_notes_api
+      # - SERVICES=edx_notes_api
     - SERVICES=credentials
     - SERVICES=analyticspipeline
     - SERVICES=xqueue
--- a/README.rst
+++ b/README.rst
@@ -1338,6 +1338,14 @@ docker-sync, but this feature hasn't been fully implemented yet (as of
 Docker 17.12.0-ce, "delegated" behaves the same as "cached").  There is a
 GitHub issue which explains the `current status of implementing delegated consistency mode`_.
 
+Known Issues
+------------
+
+The Notes service has been disabled and removed from provisioning due to issues with getting it working with ES7.
+It will be added again after these issues have been resolved, but Notes will have to be provisioned separately once
+the fixes have been merged.
+
+
 Advanced Configuration Options
 ------------------------------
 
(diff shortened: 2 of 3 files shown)


regisb:
Looks like this is failing because of `dev.check.notes`:

```
$ make dev.check."$SERVICES"
bash ./check.sh edx_notes_api
Checking edx_notes_api heartbeat:
> curl --fail -L http://localhost:18120/heartbeat
curl: (56) Recv failure: Connection reset by peer
/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/paramiko/transport.py:33: CryptographyDeprecationWarning: Python 3.5 support will be dropped in the next release ofcryptography. Please upgrade your Python.
  from cryptography.hazmat.backends import default_backend
Attaching to edx.devstack.edxnotesapi
Successful checks: NONE
Failed checks: edx_notes_api_heartbeat
Check result: FAILURE
Makefile:357: recipe for target 'dev.check.edx_notes_api' failed
make: *** [dev.check.edx_notes_api] Error 2
```

Also, could you please comment out the "notes" line instead of deleting it? It will make it easier to insert it again later. A line in the README would be great as well.

dianakhuang:
Unfortunately, I think that bash doesn't let you comment out the middle of strings. I can make the comment clearer and add something to the README, though.

provision.sh
```diff
@@ -42,14 +42,15 @@ NC='\033[0m' # No Color
 # of services are requested.
 # Changing this order may break provisioning.
 # For example, Discovery breaks if LMS is not provisioned first.
+
+# TODO: To resolve issues with notes provisioning with ES7, we are removing it from this list until it can be added again.
```
kdmccormick: This list is just for establishing a provisioning order -- it doesn't actually determine what is provisioned. Instead, you want to modify `DEFAULT_SERVICES` in `options.mk`.

That will remove it from up/migrations/provisioning/etc, but still allow us to do `make dev.provision.notes` and `make dev.up.edx_notes_api`  on our own.

kdmccormick (changes requested):
The idea looks good; my comment is just about an implementation detail.

Also, the **Services** section in the readme might be a good place to note this change.

provision.sh
dianakhuang: Thanks! I was just going to let things run and see what breaks and what doesn't.

provision.sh
```diff
@@ -42,14 +42,15 @@ NC='\033[0m' # No Color
 # of services are requested.
 # Changing this order may break provisioning.
 # For example, Discovery breaks if LMS is not provisioned first.
+
+# TODO: To resolve issues with notes provisioning with ES7, we are removing it from this list until it can be added again between forum and registrar.
```
kdmccormick: @dianakhuang can you revert the change to this file?

dianakhuang: Yup. No problem.

kdmccormick (approved):
LGTM! (haven't tested it myself)

PR #345, opened 2018-10-23
Make sure OPENEDX_GIT_BRANCH is honoured during clone

silviot:
The `repo.sh` script seems to currently ignore the value of the `OPENEDX_GIT_BRANCH` variable in its `_clone` function. It's only honoured in the `_checkout` function.

This PR makes sure that also when invoking `make dev.clone` the version in `OPENEDX_GIT_BRANCH` is cloned.

PR #628, opened 2020-09-30
Don't limit devstack clones to a single branch

timmc-edx:
This option had caused clones to have a `remote.origin.fetch` config value
of `+refs/heads/master:refs/remotes/origin/master`, which resulted in
developers not being able to see other remote branches, but only in repos
cloned by `make clone`.

(The usual value is `+refs/heads/*:refs/remotes/origin/*`)

The behavior was originally introduced in a PR for checking out
a specified branch after clone (https://github.com/edx/devstack/pull/345)
and I don't think the single-branch behavior is required for that goal.

### DIFF
--- a/repo.sh
+++ b/repo.sh
@@ -129,9 +129,9 @@ _clone ()
             cd ..
         else
             if [ "${SHALLOW_CLONE}" == "1" ]; then
-                git clone --single-branch -b ${OPENEDX_GIT_BRANCH} -c core.symlinks=true --depth=1 "${repo}"
+                git clone -b ${OPENEDX_GIT_BRANCH} -c core.symlinks=true --depth=1 "${repo}"
             else
-                git clone --single-branch -b ${OPENEDX_GIT_BRANCH} -c core.symlinks=true "${repo}"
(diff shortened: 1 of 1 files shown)


crice100 (approved):

kdmccormick (approved):
Have not tested, but looks good. Thanks for the change. Consider dropping a note on Discourse.

FYIing @regisb .

repo.sh
```diff
@@ -129,9 +129,9 @@ _clone ()
             cd ..
         else
             if [ "${SHALLOW_CLONE}" == "1" ]; then
-                git clone --single-branch -b ${OPENEDX_GIT_BRANCH} -c core.symlinks=true --depth=1 "${repo}"
+                git clone -b ${OPENEDX_GIT_BRANCH} -c core.symlinks=true --depth=1 "${repo}"
```
matthugs, line 132: nit: want to emphasize that this portion of the diff is no-op, since setting a `--depth` implies `--single-branch`

matthugs (approved):

repo.sh
timmc-edx: Yeah, this should change nothing for shallow clones.

regisb:
Looks very good to me :)

PR #629, opened 2020-09-30
Update notes MySQL version in docker-compose.yml

adzuci:
This is the first step towards getting notes to use mysql57 in production

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -252,7 +252,7 @@ services:
       - devpi
       - elasticsearch7
       - lms
-      - mysql
+      - mysql57
     image: edxops/notes:${OPENEDX_RELEASE:-latest}
     networks:
       default:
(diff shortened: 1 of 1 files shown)


natabene:
@fredsmith or @jdmulloy Could you review please when you have a chance?

jdmulloy (approved):

docker-compose.yml
```diff
@@ -247,7 +247,7 @@ services:
     depends_on:
       - devpi
       - elasticsearch
-      - mysql
+      - mysql57
```
regisb, line 255: You will almost certainly have to change `DB_HOST: "edx.devstack.mysql"` to `DB_HOST: "edx.devstack.mysql57"`, no?

adzuci: Oof, yes, thanks for the catch Regis!  I was rushing and definitely meant to update that part...

adzuci:
Updated the DB_HOST value and rebased the PR.

PR #634, opened 2020-10-05
DISCO-1675: switch discovery to mysql57

mikix:


docker-compose.yml
```diff
@@ -197,13 +197,14 @@ services:
...
     depends_on:
-      - mysql
+      - mysql57
       - elasticsearch
       - memcached
     # Allows attachment to the discovery service using 'docker attach <containerID>'.
     stdin_open: true
     tty: true
     environment:
+      DB_MIGRATION_HOST: edx.devstack.mysql57
```
mikix, line 209: This is an existing override mechanism for the default db host value coming from `edx/configuration`. Once `configuration` is updated for a bit, we can remove this again.

Makefile
```diff
@@ -257,6 +259,7 @@ dev.drop-db: _expects-database.dev.drop-db
 
 dev.drop-db.%: ## Irreversably drop the contents of a MySQL database.
```
adzuci: Nit: could update this comment or make changes to other comments to note that the targets impact both DBs, but I think it's also fine not to.

adzuci (approved):
LGTM

bseverino (approved):

PR #635, opened 2020-10-05
Correct the service name to include underscore.

feanil:
https://hub.docker.com/repository/docker/edxops/analytics_pipeline This exists but `analyticspipeline` does not.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ env:
     - SERVICES=ecommerce
     - SERVICES=edx_notes_api
     - SERVICES=credentials
-    - SERVICES=analyticspipeline
+    - SERVICES=analytics_pipeline
     - SERVICES=xqueue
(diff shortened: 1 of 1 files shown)


kdmccormick (changes requested):
I don't think this will fix it (although if I'm wrong, we'll find that out soon :) My understanding of the flow here, unless it has changed since I merged #505, is that:

* `$SERVICES` gets passed to `make dev.pull.%`
* `make dev.pull.analyticspipeline` does `docker pull --include-deps analyticspipeline`
* `analyticspipeline` is the name of the service, which references the image `edxops/analytics_pipeline`

The confusing part here is that the image and container are called `analytics_pipeline`, but the docker-compose service is called `analyticspipeline`.

feanil:
Ah, yea, I misread the target as the image, I see that `nodemanager` it's failing to pull.

PR #636, opened 2020-10-05
Adding lms as a dependencies of discovery

jinder1s:
lms is a run time dependency of discovery, especially for login. It was missing from docker-compose.yml.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -200,6 +200,7 @@ services:
       - mysql
       - elasticsearch
       - memcached
+      - lms
     # Allows attachment to the discovery service using 'docker attach <containerID>'.
     stdin_open: true
     tty: true


regisb (changes requested):
This introduces a circular dependency between the LMS and the Discovery services:

```
ERROR: Circular dependency between discovery and lms
```

https://travis-ci.com/github/edx/devstack/jobs/395350684

Unfortunately, docker-compose is unable to handle this. You could remove the dependency on the discovery service from the LMS, but then that would break the LMS, which is used by more people than the discovery service. So I suggest you simply add a note to the readme, stating that to start the discovery service you also need to start the LMS.

jinder1s:
Thanks  for the clarification! Glad there is a reason for this omission.

PR #637, opened 2020-10-06
Make `dev.reset-repos` safer by refusing to switch/update unclean branches

timmc-edx:


repo.sh
```diff
@@ -168,19 +168,20 @@ clone_private ()
...
-    currDir=$(pwd)
     for repo in ${repos[*]}
     do
         [[ $repo =~ $name_pattern ]]
         name="${BASH_REMATCH[1]}"
 
         if [ -d "$name" ]; then
-            cd "$name";git reset --hard HEAD;git checkout master;git reset --hard origin/master;git pull;cd "$currDir"
+            (cd "$name"; git checkout -q master && git pull -q --ff-only) || {
+                echo "Failed to checkout and pull master in $name repo. Exiting."
```
jinder1s: Does the git command output error for why it failed to checkout and pull? If not, it might be a good idea to add directions for why it failed/recommended next steps.

timmc-edx: It does, but I suppress that with `-q` because otherwise it also prints information about how it succeeded. I assume if people see this error message they will go in manually and run `git checkout master` and `git pull`.

feanil: Can we tell people to do that as a part of your message?

eg. `... Exiting.  Please go to the repo and cleanup any issues preventing 'git checkout master' from working.`

timmc-edx: Yeah, I'll make it more explicit.

jinder1s (approved):

PR #638, opened 2020-10-07
Add some git tips to the README

kdmccormick:
@jhan217 @MatthewPiatetsky 

as follow-up from the [ecomm repo size RCA](https://openedx.atlassian.net/wiki/spaces/ENG/pages/1894515349/RCA+ecommerce+git+repository+size+spike)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1281,6 +1281,50 @@ what your current code branch expects; you may need to rerun ``pip`` on a
 requirements file or pull new container images that already have the required
 package versions installed.
 
+Missing git branches
+~~~~~~~~~~~~~~~~~~~~
+
+When trying to check out a branch, you may see an error like this::
+
+    git checkout jj/REV-666-implement-evil-feature
+    > error: pathspec 'jj/REV-666-implement-evil-feature' did not match any file(s) known to git
+
+If you are sure you have (i) recently run ``git fetch`` and (ii) didn't misspell the
+branch name, then it is possible your repository is set in "single-branch" mode, meaning
+that it is configured to only fetch ``master``. Although devstack currently clones
+repositories with all their branches, devstacks provisioned before September 2020
+will start out with single-branch repositories.
+
+You can manually reconfigure your repository to pull all branches by running::
+
+    git config remote.origin.fetch "+refs/heads/*:refs/remotes/origin/*"
+    git fetch origin
+    git checkout jj/REV-666-implement-evil-feature
+    > Switched to branch 'jj/REV-666-implement-evil-feature'.
+
+General git troubleshooting
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+``git`` is powerful but complex; you may occasionally find your respository in a
+confusing state. This problem isn't devstack-specific.
+
+If you find yourself stuck, folks in the edX-internal or Open edX Slack workspaces may
+be able to give you a hand.
+
+Alternatively, if you are at a roadblack and
+*don't care about any data that is in your local copy of the repository*
+(i.e., you have pushed or otherwise saved your work elsewhere)
+then you can always delete the repository and start over again::
+
+    rm -r ./<repository>
+    git clone git@github.com:edx/<repository>
+
+Finally, if you regularly find yourself mystified by git, consider reading
+through `Understanding Git Conceptually`_. It explains core Git principles in way
+that makes it easier to use the simpler ``git`` commands more effectively
+and easier to use the more complicated ``git`` commands when you have to.
+
+
 Troubleshooting: Performance
 ----------------------------
 
@@ -1398,3 +1442,4 @@ it's good to bring down containers before changing any settings.
 .. _Marketing site instructions: https://openedx.atlassian.net/wiki/spaces/ENG/pages/159162183/Marketing+Site
 .. _updating relational database dumps: docs/database-dumps.rst
 .. _building images for devstack: docs/building-images.rst
+.. _Understanding Git Conceptually: https://www.sbf5.com/~cduan/technical/git/
--- a/README.rst
+++ b/README.rst
@@ -1291,7 +1291,7 @@ When trying to check out a branch, you may see an error like this::
 
 If you are sure you have (i) recently run ``git fetch`` and (ii) didn't misspell the
 branch name, then it is possible your repository is set in "single-branch" mode, meaning
-that it is configured to only fetch ``master``. Although devstack currently clones IDAs'
+that it is configured to only fetch ``master``. Although devstack currently clones services'
 repositories with all their branches, devstacks provisioned before September 2020
 will start out with single-branch repositories. You check if your repository is in this
 state by running ``git branch -r``. If you only see a couple of entries
--- a/README.rst
+++ b/README.rst
@@ -1320,7 +1320,7 @@ Alternatively, if you are at a roadblock and
 (i.e., you have pushed or otherwise saved your work elsewhere)
 then you can always delete the repository and start over again::
 
-    rm -r ./<repository>
+    rm -rf ./<repository>
     git clone git@github.com:edx/<repository>
 
 Finally, if you regularly find yourself mystified by git, consider reading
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -1281,6 +1281,50 @@ what your current code branch expects; you may need to rerun ``pip`` on a
 requirements file or pull new container images that already have the required
 package versions installed.
 
+Missing git branches
+~~~~~~~~~~~~~~~~~~~~
+
+When trying to check out a branch, you may see an error like this::
```
jhan217, line 1287: Is the double `::` a syntax thing?

README.rst
```diff
@@ -1281,6 +1281,50 @@ what your current code branch expects; you may need to rerun ``pip`` on a
...
+
+    git checkout jj/REV-666-implement-evil-feature
```
jhan217, line 1289: Can we change this to `matthewpiatetsky/REV-666-implement-evil-feature`?

README.rst
```diff
@@ -1281,6 +1281,50 @@ what your current code branch expects; you may need to rerun ``pip`` on a
...
+    > error: pathspec 'jj/REV-666-implement-evil-feature' did not match any file(s) known to git
+
+If you are sure you have (i) recently run ``git fetch`` and (ii) didn't misspell the
+branch name, then it is possible your repository is set in "single-branch" mode, meaning
+that it is configured to only fetch ``master``. Although devstack currently clones
```
jhan217: Maybe add something about running `git branch` to check if you are in this condition?

README.rst
```diff
@@ -1281,6 +1281,50 @@ what your current code branch expects; you may need to rerun ``pip`` on a
...
+repositories with all their branches, devstacks provisioned before September 2020
+will start out with single-branch repositories.
+
+You can manually reconfigure your repository to pull all branches by running::
+
+    git config remote.origin.fetch "+refs/heads/*:refs/remotes/origin/*"
```
jhan217, line 1304: Does this need to be run from the current directory/repo?

README.rst
```diff
@@ -1281,6 +1281,50 @@ what your current code branch expects; you may need to rerun ``pip`` on a
...
+
+If you find yourself stuck, folks in the edX-internal or Open edX Slack workspaces may
+be able to give you a hand.
+
+Alternatively, if you are at a roadblack and
+*don't care about any data that is in your local copy of the repository*
+(i.e., you have pushed or otherwise saved your work elsewhere)
+then you can always delete the repository and start over again::
+
+    rm -r ./<repository>
```
jhan217: `-rf`?

README.rst
```diff
@@ -1281,6 +1281,50 @@ what your current code branch expects; you may need to rerun ``pip`` on a
...
+repositories with all their branches, devstacks provisioned before September 2020
+will start out with single-branch repositories.
+
+You can manually reconfigure your repository to pull all branches by running::
+
+    git config remote.origin.fetch "+refs/heads/*:refs/remotes/origin/*"
+    git fetch origin
+    git checkout jj/REV-666-implement-evil-feature
+    > Switched to branch 'jj/REV-666-implement-evil-feature'.
+
```
jhan217, line 1308: Worth adding a step here to check that running `git remote -v` shows the correct repo as the remote here, to prevent the issue in the RCA?

README.rst
```diff
@@ -1281,6 +1281,50 @@ what your current code branch expects; you may need to rerun ``pip`` on a
...
+General git troubleshooting
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+``git`` is powerful but complex; you may occasionally find your respository in a
+confusing state. This problem isn't devstack-specific.
+
+If you find yourself stuck, folks in the edX-internal or Open edX Slack workspaces may
+be able to give you a hand.
+
+Alternatively, if you are at a roadblack and
```
jhan217: nit: `roadblock` ?

README.rst
```diff
@@ -1281,6 +1281,50 @@ what your current code branch expects; you may need to rerun ``pip`` on a
...
+*don't care about any data that is in your local copy of the repository*
```
jhan217: Should this be `any code changes`? Or does killing off the repo also kill off the data in the database?

README.rst
```diff
@@ -1281,6 +1281,50 @@ what your current code branch expects; you may need to rerun ``pip`` on a
...
+
+When trying to check out a branch, you may see an error like this::
+
+    git checkout jj/REV-666-implement-evil-feature
+    > error: pathspec 'jj/REV-666-implement-evil-feature' did not match any file(s) known to git
+
+If you are sure you have (i) recently run ``git fetch`` and (ii) didn't misspell the
+branch name, then it is possible your repository is set in "single-branch" mode, meaning
+that it is configured to only fetch ``master``. Although devstack currently clones
+repositories with all their branches, devstacks provisioned before September 2020
```
jhan217, line 1295: Does edx-platform also clone with all the branches?

README.rst
```diff
@@ -1281,6 +1281,50 @@ what your current code branch expects; you may need to rerun ``pip`` on a
...
+will start out with single-branch repositories.
+
+You can manually reconfigure your repository to pull all branches by running::
```
MatthewPiatetsky: perfect!

MatthewPiatetsky (commented):
thank you!

README.rst
kdmccormick: yup, it makes the following indented lines into a code block.

you can see the result if you select "view the rich diff" over in Files Changed:
image

README.rst
kdmccormick: `-r` (recursive) without `-f` (force) will remove a directory, but still warn if you are trying to remove write-protected files. So you'll more likely get a warning if you accidentally try to remove the wrong thing.

If someone is using this section in the README to help them with `rm`, then they're probably pretty new to using `rm`, so I didn't want to direct them to use `-f` right off the bat.

README.rst
kdmccormick: `devstack` clones the repositories of all default IDAs with all their branches now. I will change this to "currently clones IDA repositories with all their branches".

README.rst
kdmccormick: throwing your teammate under the bus, huh? seems pretty evil to me @jhan217 .

README.rst
jhan217: TIL

README.rst
kdmccormick: Following this process doesn't put you at any risk of setting the wrong origin, so it seems somewhat tangential to add to this section. My hope is that by having a process to follow, one is less likely to need to fiddle with their `origin` and accidentally set the wrong remote.

While checking `git remote -v` is a good piece of Git advice, there are a lot of worthy pieces of Git advice (e.g. don't force push to master, use good commit messages, etc) that we could list, but I don't think this README is a good place for all of those.

README.rst
jhan217: (what does IDA stand for?)

README.rst
kdmccormick: Yup, good call-out.

README.rst
jhan217: Are there a lot of those in a given repo?

README.rst
kdmccormick: The database is not within the repository, but the reader may not know that. I'll clarify this.

kdmccormick: Ok, I changed it to `*don't care about any changes to your local copy of the repository*`. I didn't mention code vs. databases, because most repositories don't use any databases inside themselves, but some might. Plus, there are other non-code changes you can make to your local repository (text files, config, any random file you happened to add) so I would rather leave it to the reader to determine whether it's safe to delete the folder or not.

README.rst
kdmccormick: good idea :+1:

jhan217 (approved):

README.rst
kdmccormick: edX jargon for Independently Deployable Application.

I'm going to change it to `services'`. Why use made-up edX words when we don't have to?

README.rst
kdmccormick: actually, yeah, I just tried to `rm -r course-discovery/` and there were a lot. I'll change it to `-rf`

PR #642, opened 2020-10-13
Update instructions for multiple devstacks

angonz:
Update the instructions for installing multiple releases in one machine and switching among them.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -499,7 +499,7 @@ You can have multiple isolated Devstacks provisioned on a single computer now. F
 
 #. If you haven't done so, follow the steps in the `Getting Started`_ section, to install the master devstack or any other named release. We recommend that you have at least one devstack on the master branch.
 #. Change directory to your devstack and activate the virtual env.
-#. Stop any running containers by issuing a ``make dev.stop``. Note that in older named releases ``make dev.stop`` may not be available. Use ``make stop`` instead. Make sure that all containers are stopped from the docker dashboard.
+#. Stop any running containers by issuing a ``make dev.stop``.
 #. Follow the steps in `Getting Started`_ section again, setting the additional OPENEDX_RELEASE you want to install in step 2
 
 The implication of this is that you can switch between isolated Devstack databases by changing the value of the ``OPENEDX_RELEASE`` environment variable.
@@ -507,7 +507,7 @@ The implication of this is that you can switch between isolated Devstack databas
 Switch between your Devstack releases by doing the following:
 *************************************************************
 
-#. Stop the containers by issuing a ``make dev.stop`` for the running release. Note that in older named releases ``make dev.stop`` may not be available. Use ``make stop`` instead.
+#. Stop the containers by issuing a ``make dev.stop`` for the running release.
 #. Edit the project name in ``options.local.mk`` or set the ``OPENEDX_RELEASE`` environment variable and let the ``COMPOSE_PROJECT_NAME`` be assigned automatically. 
 #. Check out the appropriate branch in devstack, e.g. ``git checkout open-release/ironwood.master``
 #. Use ``make dev.checkout`` to check out the correct branch in the local
--- a/README.rst
+++ b/README.rst
@@ -108,6 +108,8 @@ the ones installed globally on your system.
 Using the Latest Images
 -----------------------
 
+By default, these instructions will install the master branch. If you want to install a named release instead (e.g. juniper.master), follow the steps in `How do I run the images for a named Open edX release?`_ before pulling the docker images. You can learn more about named releases in the `official documentation <https://edx.readthedocs.io/projects/edx-developer-docs/en/latest/named_releases.html>`_.
+
 New images for our services are published frequently.  Assuming that you've followed the steps in `Getting Started`_
 below, run the following sequence of commands if you want to use the most up-to-date versions of *all* default devstack images.
 
@@ -178,8 +180,6 @@ The default devstack services can be run by following the steps below. For analy
    .. _step 3:
 3. Pull any changes made to the various images on which the devstack depends.
 
-   **NOTE:** These instructions will install the master release. If you want to install a named release instead (e.g. juniper.master), follow the steps in `How do I run the images for a named Open edX release?`_ before pulling the docker images.
-   
    .. code:: sh
 
        make dev.pull
@@ -483,16 +483,9 @@ By default, the steps above will install the devstack using the master branch of
 #. Set the ``OPENEDX_RELEASE`` environment variable to the appropriate image
    tag; "hawthorn.master", "zebrawood.rc1", etc.  Note that unlike a server
    install, ``OPENEDX_RELEASE`` should not have the "open-release/" prefix.
-
-   **NOTE:** The ``COMPOSE_PROJECT_NAME`` variable is used to define Docker namespaced volumes and network based on this value, so changing it will give you a separate set of databases. This is handled for you automatically by setting the ``OPENEDX_RELEASE`` environment variable in ``options.mk`` (e.g. ``COMPOSE_PROJECT_NAME=devstack-juniper.master``. Should you want to manually override this, edit the ``options.local.mk`` in the root of this repo and create the file if it does not exist. Change the devstack project name by adding the following line:
-   ``COMPOSE_PROJECT_NAME=<your-alternate-devstack-name>`` (e.g. ``COMPOSE_PROJECT_NAME=secondarydevstack``)
-
-   As a specific example, if ``OPENEDX_RELEASE`` is set in your environment as ``juniper.master``, then ``COMPOSE_PROJECT_NAME`` will default to ``devstack-juniper.master`` instead of ``devstack``.
-
 #. Check out the appropriate branch in devstack, e.g. ``git checkout open-release/ironwood.master``
 #. Use ``make dev.checkout`` to check out the correct branch in the local
-   checkout of each service repository once you've set the ``OPENEDX_RELEASE``
-   environment variable above.
+   checkout of each service repository
 #. Continue with `step 3`_ in the Getting Started guide to pull the correct docker images.
 
 All ``make`` target and ``docker-compose`` calls should now use the correct
@@ -502,11 +495,11 @@ an empty string.
 
 How do I run multiple named Open edX releases on same machine?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-You can have multiple isolated Devstacks provisioned on a single computer now. Follow these directions **after installing the master release devstack** to switch between the named releases.
+You can have multiple isolated Devstacks provisioned on a single computer now. Follow these directions **after installing at least two devstacks** to switch between them.
 
 #. If you haven't done so, follow the steps in the `Getting Started`_ section, to install the master devstack or any other named release. We recommend that you have at least one devstack on the master branch.
 #. Change directory to your devstack and activate the virtual env.
-#. Stop any running containers by issuing a ``make stop``. Make sure that all containers are stopped from the docker dashboard.
+#. Stop any running containers by issuing a ``make dev.stop``. Note that in older named releases ``make dev.stop`` may not be available. Use ``make stop`` instead. Make sure that all containers are stopped from the docker dashboard.
 #. Follow the steps in `Getting Started`_ section again, setting the additional OPENEDX_RELEASE you want to install in step 2
 
 The implication of this is that you can switch between isolated Devstack databases by changing the value of the ``OPENEDX_RELEASE`` environment variable.
@@ -514,7 +507,7 @@ The implication of this is that you can switch between isolated Devstack databas
 Switch between your Devstack releases by doing the following:
 *************************************************************
 
-#. Stop the containers by issuing a ``make stop`` for the running release.
+#. Stop the containers by issuing a ``make dev.stop`` for the running release. Note that in older named releases ``make dev.stop`` may not be available. Use ``make stop`` instead.
 #. Edit the project name in ``options.local.mk`` or set the ``OPENEDX_RELEASE`` environment variable and let the ``COMPOSE_PROJECT_NAME`` be assigned automatically. 
 #. Check out the appropriate branch in devstack, e.g. ``git checkout open-release/ironwood.master``
 #. Use ``make dev.checkout`` to check out the correct branch in the local
@@ -1427,6 +1420,18 @@ For more detail, refer to the comments in the file itself.
 If you're feeling brave, you can create an git-ignored overrides file called
 ``options.local.mk`` in the same directory and set your own values. In general,
 it's good to bring down containers before changing any settings.
+
+Changing the Docker Compose Project Name
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``COMPOSE_PROJECT_NAME`` variable is used to define Docker namespaced volumes 
+and network based on this value, so changing it will give you a separate set of databases. 
+This is handled for you automatically by setting the ``OPENEDX_RELEASE`` environment variable in ``options.mk`` 
+(e.g. ``COMPOSE_PROJECT_NAME=devstack-juniper.master``. Should you want to manually override this, edit the ``options.local.mk`` in the root of this repo and create the file if it does not exist. Change the devstack project name by adding the following line:
+   ``COMPOSE_PROJECT_NAME=<your-alternate-devstack-name>`` (e.g. ``COMPOSE_PROJECT_NAME=secondarydevstack``)
+
+As a specific example, if ``OPENEDX_RELEASE`` is set in your environment as ``juniper.master``, then ``COMPOSE_PROJECT_NAME`` will default to ``devstack-juniper.master`` instead of ``devstack``.
+
 
 .. _Docker Compose: https://docs.docker.com/compose/
 .. _Docker for Mac: https://docs.docker.com/docker-for-mac/
--- a/README.rst
+++ b/README.rst
@@ -905,8 +905,6 @@ LMS and Studio (a.k.a. CMS) read many configuration settings from the container 
 in the following locations:
 
 - ``/edx/etc/lms.yml``
-- ``/edx/etc/lms.yml``
-- ``/edx/etc/studio.yml``
 - ``/edx/etc/studio.yml``
 
 Changes to these files will *not* persist over a container restart, as they
--- a/README.rst
+++ b/README.rst
@@ -175,6 +175,7 @@ The default devstack services can be run by following the steps below. For analy
    (macOS only) Share the cloned service directories in Docker, using
    **Docker -> Preferences -> File Sharing** in the Docker menu.
 
+   .. _step 3:
 3. Pull any changes made to the various images on which the devstack depends.
 
    **NOTE:** These instructions will install the master release. If you want to install a named release instead (e.g. juniper.master), follow the steps in `How do I run the images for a named Open edX release?`_ before pulling the docker images.
@@ -477,13 +478,13 @@ Frequently Asked Questions
 
 How do I run the images for a named Open edX release?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-By default, the steps above will install the devstack using the master branch of all repos. If you want to install a named release instead, follow these steps before pulling the docker images in step 3 of the Getting Started guide:
+By default, the steps above will install the devstack using the master branch of all repos. If you want to install a named release instead, follow these steps before pulling the docker images in `step 3`_ of the Getting Started guide:
 
 #. Set the ``OPENEDX_RELEASE`` environment variable to the appropriate image
    tag; "hawthorn.master", "zebrawood.rc1", etc.  Note that unlike a server
    install, ``OPENEDX_RELEASE`` should not have the "open-release/" prefix.
 
-   **NOTE:** The ``COMPOSE_PROJECT_NAME`` variable is used to define Docker namespaced volumes and network based on this value, so changing it will give you a separate set of databases. This is handled for you automatically by setting the ``OPENEDX_RELEASE`` environment variable in ``options.mk`` (e.g. ``COMPOSE_PROJECT_NAME=devstack-juniper.master``. Should you want to manually override this edit the ``options.local.mk`` in the root of this repo and create the file if it does not exist. Change the devstack project name by adding the following line:
+   **NOTE:** The ``COMPOSE_PROJECT_NAME`` variable is used to define Docker namespaced volumes and network based on this value, so changing it will give you a separate set of databases. This is handled for you automatically by setting the ``OPENEDX_RELEASE`` environment variable in ``options.mk`` (e.g. ``COMPOSE_PROJECT_NAME=devstack-juniper.master``. Should you want to manually override this, edit the ``options.local.mk`` in the root of this repo and create the file if it does not exist. Change the devstack project name by adding the following line:
    ``COMPOSE_PROJECT_NAME=<your-alternate-devstack-name>`` (e.g. ``COMPOSE_PROJECT_NAME=secondarydevstack``)
 
    As a specific example, if ``OPENEDX_RELEASE`` is set in your environment as ``juniper.master``, then ``COMPOSE_PROJECT_NAME`` will default to ``devstack-juniper.master`` instead of ``devstack``.
@@ -492,7 +493,7 @@ By default, the steps above will install the devstack using the master branch of
 #. Use ``make dev.checkout`` to check out the correct branch in the local
    checkout of each service repository once you've set the ``OPENEDX_RELEASE``
    environment variable above.
-#. Continue with step 3 in the Getting Started guide to pull the correct docker images.
+#. Continue with `step 3`_ in the Getting Started guide to pull the correct docker images.
 
 All ``make`` target and ``docker-compose`` calls should now use the correct
 images until you change or unset ``OPENEDX_RELEASE`` again.  To work on the
@@ -503,10 +504,10 @@ How do I run multiple named Open edX releases on same machine?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 You can have multiple isolated Devstacks provisioned on a single computer now. Follow these directions **after installing the master release devstack** to switch between the named releases.
 
-#. If you haven't done so, follow the `Getting Started`_ guide to install the master devstack or any other named release. We recommend that you have at least one devstack on the master branch.
+#. If you haven't done so, follow the steps in the `Getting Started`_ section, to install the master devstack or any other named release. We recommend that you have at least one devstack on the master branch.
 #. Change directory to your devstack and activate the virtual env.
 #. Stop any running containers by issuing a ``make stop``. Make sure that all containers are stopped from the docker dashboard.
-#. Follow the steps in `Getting Started`_ section, setting the additional OPENEDX_RELEASE you want to install in step 2
+#. Follow the steps in `Getting Started`_ section again, setting the additional OPENEDX_RELEASE you want to install in step 2
 
 The implication of this is that you can switch between isolated Devstack databases by changing the value of the ``OPENEDX_RELEASE`` environment variable.
 
@@ -517,8 +518,7 @@ Switch between your Devstack releases by doing the following:
 #. Edit the project name in ``options.local.mk`` or set the ``OPENEDX_RELEASE`` environment variable and let the ``COMPOSE_PROJECT_NAME`` be assigned automatically. 
 #. Check out the appropriate branch in devstack, e.g. ``git checkout open-release/ironwood.master``
 #. Use ``make dev.checkout`` to check out the correct branch in the local
-   checkout of each service repository once you've set the ``OPENEDX_RELEASE``
-   environment variable above.
+   copy of each service repository
 #. Bring up the containers with ``make dev.up``, ``make dev.nfs.up`` or ``make dev.sync.up``.
 
 **NOTE:** Additional instructions on switching releases using ``direnv`` can be found in `How do I switch releases using 'direnv'?`_ section.
--- a/README.rst
+++ b/README.rst
@@ -177,6 +177,8 @@ The default devstack services can be run by following the steps below. For analy
 
 3. Pull any changes made to the various images on which the devstack depends.
 
+   **NOTE:** These instructions will install the master release. If you want to install a named release instead (e.g. juniper.master), follow the steps in `How do I run the images for a named Open edX release?`_ before pulling the docker images.
+   
    .. code:: sh
 
        make dev.pull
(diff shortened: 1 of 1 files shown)


openedx-webhooks:
Thanks for the pull request, @angonz! I've created [OSPR-5057](https://openedx.atlassian.net/browse/OSPR-5057) to keep track of it in JIRA, where we prioritize reviews. Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

angonz:
Hi @regisb, I had forgotten to push the create PR button... now it's on its way.

README.rst
```diff
@@ -475,15 +477,22 @@ Frequently Asked Questions
 
 How do I run the images for a named Open edX release?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+By default, the steps above will install the devstack using the master branch of all repos. If you want to install a named release instead, follow these steps before pulling the docker images in step 3 of the Getting Started guide:
 
 #. Set the ``OPENEDX_RELEASE`` environment variable to the appropriate image
    tag; "hawthorn.master", "zebrawood.rc1", etc.  Note that unlike a server
    install, ``OPENEDX_RELEASE`` should not have the "open-release/" prefix.
+
+   **NOTE:** The ``COMPOSE_PROJECT_NAME`` variable is used to define Docker namespaced volumes and network based on this value, so changing it will give you a separate set of databases. This is handled for you automatically by setting the ``OPENEDX_RELEASE`` environment variable in ``options.mk`` (e.g. ``COMPOSE_PROJECT_NAME=devstack-juniper.master``. Should you want to manually override this edit the ``options.local.mk`` in the root of this repo and create the file if it does not exist. Change the devstack project name by adding the following line:
```
regisb: Missing comma: "Should you want to manually override this, edit..."

README.rst
```diff
@@ -475,15 +477,22 @@ Frequently Asked Questions
...
+   ``COMPOSE_PROJECT_NAME=<your-alternate-devstack-name>`` (e.g. ``COMPOSE_PROJECT_NAME=secondarydevstack``)
+
+   As a specific example, if ``OPENEDX_RELEASE`` is set in your environment as ``juniper.master``, then ``COMPOSE_PROJECT_NAME`` will default to ``devstack-juniper.master`` instead of ``devstack``.
+
 #. Check out the appropriate branch in devstack, e.g. ``git checkout open-release/ironwood.master``
 #. Use ``make dev.checkout`` to check out the correct branch in the local
    checkout of each service repository once you've set the ``OPENEDX_RELEASE``
    environment variable above.
-#. ``make dev.pull`` to get the correct images.
+#. Continue with step 3 in the Getting Started guide to pull the correct docker images.
```
regisb: Can you add an internal link to step 3? (same above)

README.rst
```diff
@@ -492,25 +501,25 @@ an empty string.
...
-#. Perform steps in `How do I run the images for a named Open edX release?`_ for specific release.
-#. Follow the steps in `Getting Started`_ section to update requirements (e.g. ``make requirements``) and provision (e.g. ``make dev.provision``) the new named release containers.
+You can have multiple isolated Devstacks provisioned on a single computer now. Follow these directions **after installing the master release devstack** to switch between the named releases.
 
-As a specific example, if ``OPENEDX_RELEASE`` is set in your environment as ``juniper.master``, then ``COMPOSE_PROJECT_NAME`` will default to ``devstack-juniper.master`` instead of ``devstack``.
+#. If you haven't done so, follow the `Getting Started`_ guide to install the master devstack or any other named release. We recommend that you have at least one devstack on the master branch.
+#. Change directory to your devstack and activate the virtual env.
+#. Stop any running containers by issuing a ``make stop``. Make sure that all containers are stopped from the docker dashboard.
+#. Follow the steps in `Getting Started`_ section, setting the additional OPENEDX_RELEASE you want to install in step 2
```
regisb: "Follow the steps in the `Getting Started`_ section,"

README.rst
```diff
@@ -492,25 +501,25 @@ an empty string.
...
 *************************************************************
 
-#. Stop the containers by issuing a ``make dev.stop`` for the running release.
-#. Follow the instructions from the `How do I run multiple named Open edX releases on same machine?`_ section.
+#. Stop the containers by issuing a ``make stop`` for the running release.
 #. Edit the project name in ``options.local.mk`` or set the ``OPENEDX_RELEASE`` environment variable and let the ``COMPOSE_PROJECT_NAME`` be assigned automatically. 
-#. Bring up the containers with ``make dev.up``.
+#. Check out the appropriate branch in devstack, e.g. ``git checkout open-release/ironwood.master``
+#. Use ``make dev.checkout`` to check out the correct branch in the local
+   checkout of each service repository once you've set the ``OPENEDX_RELEASE``
```
regisb: I think it's unnecessary to repeat that the ``OPENEDX_RELEASE`` variable should be set.

regisb (changes requested):

regisb:
Maybe you can take a look at this @ztraboo? You're the one who originally designed the multi-devstack feature and wrote the readme instructions.

angonz (commented):
Thanks @regisb for your comments, I have sent a new commit based on your suggestions.

README.rst
```diff
@@ -175,8 +175,11 @@ The default devstack services can be run by following the steps below. For analy
    (macOS only) Share the cloned service directories in Docker, using
    **Docker -> Preferences -> File Sharing** in the Docker menu.
 
+   .. _step 3:
 3. Pull any changes made to the various images on which the devstack depends.
 
+   **NOTE:** These instructions will install the master release. If you want to install a named release instead (e.g. juniper.master), follow the steps in `How do I run the images for a named Open edX release?`_ before pulling the docker images.
```
regisb: This comment concerns not only the `dev.pull` command, but also the repository cloning. Thus, I'd suggest to put this note at the very beginning of the getting started instructions. I don't think we should  add this as a note. Instead, just describe the fact that the following instructions will bootstrap a working Open edX platform from the master branches.

BTW, the term "master release" is incorrect. This is not a release per se.

README.rst
```diff
@@ -475,15 +478,22 @@ Frequently Asked Questions
...
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+By default, the steps above will install the devstack using the master branch of all repos. If you want to install a named release instead, follow these steps before pulling the docker images in `step 3`_ of the Getting Started guide:
 
 #. Set the ``OPENEDX_RELEASE`` environment variable to the appropriate image
    tag; "hawthorn.master", "zebrawood.rc1", etc.  Note that unlike a server
    install, ``OPENEDX_RELEASE`` should not have the "open-release/" prefix.
+
+   **NOTE:** The ``COMPOSE_PROJECT_NAME`` variable is used to define Docker namespaced volumes and network based on this value, so changing it will give you a separate set of databases. This is handled for you automatically by setting the ``OPENEDX_RELEASE`` environment variable in ``options.mk`` (e.g. ``COMPOSE_PROJECT_NAME=devstack-juniper.master``. Should you want to manually override this, edit the ``options.local.mk`` in the root of this repo and create the file if it does not exist. Change the devstack project name by adding the following line:
+   ``COMPOSE_PROJECT_NAME=<your-alternate-devstack-name>`` (e.g. ``COMPOSE_PROJECT_NAME=secondarydevstack``)
```
regisb: I feel like this is the wrong place to add this note. First of all, why should anyone want to override the ``COMPOSE_PROJECT_NAME``? Answering this will help us determine where this note should go.

README.rst
```diff
@@ -475,15 +478,22 @@ Frequently Asked Questions
...
+
+   As a specific example, if ``OPENEDX_RELEASE`` is set in your environment as ``juniper.master``, then ``COMPOSE_PROJECT_NAME`` will default to ``devstack-juniper.master`` instead of ``devstack``.
+
 #. Check out the appropriate branch in devstack, e.g. ``git checkout open-release/ironwood.master``
 #. Use ``make dev.checkout`` to check out the correct branch in the local
```
regisb, line 487: I know that you didn't write this, but you could remove the "once you've set the ``OPENEDX_RELEASE``
   environment variable above" part.

README.rst
```diff
@@ -492,25 +502,24 @@ an empty string.
 
 How do I run multiple named Open edX releases on same machine?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-You can have multiple isolated Devstacks provisioned on a single computer now. Follow these directions to switch between the named releases.
+You can have multiple isolated Devstacks provisioned on a single computer now. Follow these directions **after installing the master release devstack** to switch between the named releases.
```
regisb: Wait, why do we need to install the devstack from the master branch prior to running multiple releases on the same machine? This should not be the case. If it is, then this might be a bug.

README.rst
```diff
@@ -492,25 +502,24 @@ an empty string.
...
+#. Follow the steps in `Getting Started`_ section again, setting the additional OPENEDX_RELEASE you want to install in step 2
 
 The implication of this is that you can switch between isolated Devstack databases by changing the value of the ``OPENEDX_RELEASE`` environment variable.
 
 Switch between your Devstack releases by doing the following:
 *************************************************************
 
-#. Stop the containers by issuing a ``make dev.stop`` for the running release.
-#. Follow the instructions from the `How do I run multiple named Open edX releases on same machine?`_ section.
+#. Stop the containers by issuing a ``make stop`` for the running release.
```
regisb: We are trying to push for `make dev.*` commands, so `make dev.stop` is the recommended command.

regisb (changes requested):
Hey @angonz I have quite a few more comments. Sorry if I didn't add them earlier, but at the time I didn't review this with a clear mind.

README.rst
ztraboo: @angonz I agree with @regisb comments here. I'm sure you wrote this note because you were looking to inquire about setting up a named release but didn't know how to do that until after you ran the `Getting Started` instructions and read further in the FAQ section https://github.com/angonz/devstack/blob/patch-1/README.rst#how-do-i-run-the-images-for-a-named-open-edx-release notably. I'm glad you wrote in this note at this section since that's what the developer would do first most likely. It may be worth adding a link to this documentation page showing what the named releases are for https://edx.readthedocs.io/projects/edx-developer-docs/en/latest/named_releases.html.

README.rst
ztraboo: @regisb We are defining the `COMPOSE_PROJECT_NAME` based on the setting of the `OPENEDX_RELEASE` variable here https://github.com/angonz/devstack/blob/patch-1/options.mk#L28-L39. This is how the containers get there unique name based on Open edX named release since https://github.com/angonz/devstack/blob/patch-1/docker-compose.yml#L8  says that the value of `COMPOSE_PROJECT_NAME` is " prepended along with the service name to the container on start up." We explicitly defined the `container_name` though rather than Docker defining it for us like so (e.g. https://github.com/angonz/devstack/blob/patch-1/docker-compose.yml#L294) because we needed it to be defined a certain way consistently.

The Docker container's network aliases provide a way for other containers on the network to reach each other https://github.com/angonz/devstack/blob/patch-1/docker-compose.yml#L316-L320.

cc: @angonz, @kdmccormick Any more advice on this?

README.rst
```diff
@@ -475,15 +478,22 @@ Frequently Asked Questions
...
 
 #. Set the ``OPENEDX_RELEASE`` environment variable to the appropriate image
    tag; "hawthorn.master", "zebrawood.rc1", etc.  Note that unlike a server
    install, ``OPENEDX_RELEASE`` should not have the "open-release/" prefix.
+
+   **NOTE:** The ``COMPOSE_PROJECT_NAME`` variable is used to define Docker namespaced volumes and network based on this value, so changing it will give you a separate set of databases. This is handled for you automatically by setting the ``OPENEDX_RELEASE`` environment variable in ``options.mk`` (e.g. ``COMPOSE_PROJECT_NAME=devstack-juniper.master``. Should you want to manually override this, edit the ``options.local.mk`` in the root of this repo and create the file if it does not exist. Change the devstack project name by adding the following line:
+   ``COMPOSE_PROJECT_NAME=<your-alternate-devstack-name>`` (e.g. ``COMPOSE_PROJECT_NAME=secondarydevstack``)
+
+   As a specific example, if ``OPENEDX_RELEASE`` is set in your environment as ``juniper.master``, then ``COMPOSE_PROJECT_NAME`` will default to ``devstack-juniper.master`` instead of ``devstack``.
+
```
ztraboo: Looking at https://docs.docker.com/compose/reference/envvars/#compose_project_name and seeing what I have for Juniper it should say something like `edx.devstack-juniper.master.<container>` (e.g. `edx.devstack-juniper.master.lms` for the LMS container).

README.rst
ztraboo: @angonz What you wrote is correct in that based on naming the `COMPOSE_PROJECT_NAME` on Open edX named release it does prefix the network and volumes with the project name. Here are some examples below for the `open-release/juniper.master` named release.

**Docker Network:**
An example of network for `open-release/juniper.master` results in `devstack-junipermaster_default`

**Docker Volumes:**
An example of volumes for `open-release/juniper.master` results in `devstack-junipermaster_edxapp_lms_assets`

cc: @regisb

ztraboo: @angonz Based on my conversations with @kdmccormick here https://discuss.openedx.org/t/docker-devstack-multiple-releases-one-machine/1902/4?u=zachary_trabookis it does seem what you wrote above matches with what was discussed. I think you're good on this comment.

cc: @regisb

README.rst
ztraboo: @angonz I agree with @regisb on his comment. I was able to install `open-release/juniper.master` on my machine without having to install `master` first. Please remove this comment. If you want to test this out before removing it just install Juniper and see if it works by itself.

README.rst
ztraboo: @angonz It appears that you moved the `NOTE` section from what I had done in this section https://github.com/edx/devstack#how-do-i-run-multiple-named-open-edx-releases-on-same-machine to the section above https://github.com/edx/devstack#how-do-i-run-the-images-for-a-named-open-edx-release. Since the `COMPOSE_PROJECT_NAME` is more inline with the multiple release on one machine that comment may be more applicable to that section. Again this is handled in devstack automatically with this code https://github.com/angonz/devstack/blob/patch-1/options.mk#L28-L39 and if you're just working on one release you should be as concerned with defining this project name. 

@regisb What are your additional thoughts on moving the note concerning setting the  `COMPOSE_PROJECT_NAME` to a section that doesn't talk about multiple name releases on same machine?

README.rst
ztraboo: @angonz Please test installing `open-release/juniper.master` by itself. I think what you added here in this section should just be removed. You shouldn't have to install `master` prior to an Open edX named release. This is just going to confuse someone especially if they don't or need to develop against the `master` release.

cc: @regisb

README.rst
ztraboo: @angonz Details about that are here https://github.com/edx/devstack/blob/master/Makefile#L322-L324 for the `master` release. I agree with @regisb  on this. The latest supported release `open-release/juniper.master` does still have the old Makefile target of `make stop` noted here https://github.com/edx/devstack/blob/open-release/juniper.master/Makefile#L207-L209.

README.rst
```diff
@@ -492,25 +502,24 @@ an empty string.
...
 
-#. Stop the containers by issuing a ``make dev.stop`` for the running release.
-#. Follow the instructions from the `How do I run multiple named Open edX releases on same machine?`_ section.
+#. Stop the containers by issuing a ``make stop`` for the running release.
 #. Edit the project name in ``options.local.mk`` or set the ``OPENEDX_RELEASE`` environment variable and let the ``COMPOSE_PROJECT_NAME`` be assigned automatically. 
-#. Bring up the containers with ``make dev.up``.
+#. Check out the appropriate branch in devstack, e.g. ``git checkout open-release/ironwood.master``
+#. Use ``make dev.checkout`` to check out the correct branch in the local
+   copy of each service repository
+#. Bring up the containers with ``make dev.up``, ``make dev.nfs.up`` or ``make dev.sync.up``.
```
ztraboo, line 516: @angonz I think it's fine to make this more explicit. We were trying to reuse some of the sections above like https://github.com/edx/devstack/blob/master/README.rst#how-do-i-run-multiple-named-open-edx-releases-on-same-machine here but yeah I can understand how you may want a clearer approach to switching between releases. What you wrote is fine. The only command that was left off was the `make dev.pull` noted in https://github.com/edx/devstack/blob/master/README.rst#how-do-i-run-the-images-for-a-named-open-edx-release section. I don't consider that a necessity unless you want latest updates from new commits for a release but maybe @kdmccormick @regisb  can look that over and make further suggestions.

ztraboo:
@angonz I made several comments above. Please look those over when you can and let me know if anything I said was unclear.

README.rst
kdmccormick: > @regisb: I feel like this is the wrong place to add this note. First of all, why should anyone want to override the COMPOSE_PROJECT_NAME? Answering this will help us determine where this note should go.

I find it useful to be able to do this occasionally, particularly when testing changes to devstack itself. It's nice to have an isolated set of database volumes to try out destructive migrations on.

> @ztraboo: @regisb What are your additional thoughts on moving the note concerning setting the COMPOSE_PROJECT_NAME to a section that doesn't talk about multiple name releases on same machine?

I think that's a good idea. The target audience for this setting is probably only devstack power users, so briefly touching on in the Advanced Configuration section may good enough.

README.rst
angonz: I've moved the text to "Using the Latest Images" section above. I've also changed "master release" to "master branch", I guess it's more correct.

README.rst
angonz: I'm glad you said it's only for power users, because I had no idea how to use it!
I moved it to the bottom of the file, together with another reference to the `options.mk` file, under Advanced Configuration Options section.

README.rst
angonz: You are right. What you need is to have is at least two devstacks to switch among! I've fixed it.
My confusion comes from something that should be clarified somewhere: to which branch should I submit my changes? When to master, and when to a named release? Please @ztraboo and @regisb your help with this.

README.rst
angonz: Ok, I will revert to make dev.stop but make a not for older versions.

README.rst
angonz: When you are switching between devstacks, I guess it's not necessary to pull the latest images, and it takes a lot of time. I have a question here... do you need to provision again if you pull new images? In which cases a new provision is needed?

README.rst
ztraboo: @angonz In my experience if you think any future releases may benefit from your changes then submit a PR to `master` branch, otherwise, if you're targeting a specific Open edX named release then submit a PR against that.

`master` is current working development of latest changes while Open edX named releases are snapshots in time of that `master` branch.

You don't necessarily need to have more than one releases to switch among to develop. For instance you could just focus on `master` release only and never touch any Open edX named release. What I've done on my machine is use this directory structure to avoid having to always switch branches with git and keep whatever changes I'm working on separately to avoid issues.
```
- open-edx/devstack.juniper
- open-edx/devstack.master
- open-edx/devstack.<next-release-name>
```

README.rst
angonz: @ztraboo, let me know if I'm wrong, but looking at [options.mk](https://github.com/edx/devstack/blob/master/options.mk), ``COMPOSE_PROJECT_NAME ?= devstack-${OPENEDX_RELEASE}``, which is the compose project. ``edx.devstack-juniper.master.<container>`` are the name of the containers.

angonz:
Looks like this caused a good deal of discussion!! I have submitted a 2nd round of changes.

README.rst
ztraboo: @angonz Great question. Maybe @kdmccormick @regisb can answer this. I've always tried to understand why I needed to do this. I think the reason is that whenever a new PR gets merged into a release a new Docker image gets created for that service. Some of the PR may have database changes and therefore may need to have provisioning issued again. I've never re-provisioned again for the same release after updating the Docker images since I believe it reconfigures everything again and destroys anything in your database setting anything you had back to the default release provisioning configuration.

**Makefile provision targets**
https://github.com/edx/devstack/blob/master/Makefile#L215-L225

This https://github.com/edx/devstack/blob/master/provision.sh will get called. Look at that in detail to see what it does. At the end it calls the specific service provisioning scripts. If we look into the https://github.com/edx/devstack/blob/master/provision-lms.sh#L21-L22 script here for the LMS service you will see that it performs a database migration. This script does a lot of other things like perform prerequisite install, create superuser, create static assets, etc. Notice the database load of data to save time https://github.com/edx/devstack/blob/master/provision-lms.sh#L7-L9; it appears that this drops the tables and recreates them.

Long story short I don't think I'd `make dev.provision.lms` when updating the Docker images on the same release name. Looking at the https://github.com/edx/devstack#switching-branches it says to do this only `"... when switching to a branch which differs greatly from the one you're workign on (especially if the new branch is more recent)."`. Since we're isolating these releases with their own containers I don't see the need to do this since the release probably won't change significantly.

README.rst
ztraboo: @angonz Sorry I miss-typed the container name for the LMS which I've corrected above. Even though we've set the `COMPOSE_PROJECT_NAME` with the https://github.com/edx/devstack/blob/master/options.mk#L28-L39 we noticed that Docker liked to name it something else so we explicitly defined it here https://github.com/edx/devstack/blob/master/docker-compose.yml#L294 so it would be consistent.

README.rst
```diff
@@ -108,6 +108,8 @@ the ones installed globally on your system.
 Using the Latest Images
 -----------------------
 
+By default, these instructions will install the master branch. If you want to install a named release instead (e.g. juniper.master), follow the steps in `How do I run the images for a named Open edX release?`_ before pulling the docker images. You can learn more about named releases in the `official documentation <https://edx.readthedocs.io/projects/edx-developer-docs/en/latest/named_releases.html>`_.
+
```
ztraboo, line 112: @angonz Thank you for adding this. Hopefully this will avoid confusion in the future.

README.rst
```diff
@@ -1428,6 +1421,18 @@ If you're feeling brave, you can create an git-ignored overrides file called
...
+The ``COMPOSE_PROJECT_NAME`` variable is used to define Docker namespaced volumes 
+and network based on this value, so changing it will give you a separate set of databases. 
+This is handled for you automatically by setting the ``OPENEDX_RELEASE`` environment variable in ``options.mk`` 
+(e.g. ``COMPOSE_PROJECT_NAME=devstack-juniper.master``. Should you want to manually override this, edit the ``options.local.mk`` in the root of this repo and create the file if it does not exist. Change the devstack project name by adding the following line:
+   ``COMPOSE_PROJECT_NAME=<your-alternate-devstack-name>`` (e.g. ``COMPOSE_PROJECT_NAME=secondarydevstack``)
+
+As a specific example, if ``OPENEDX_RELEASE`` is set in your environment as ``juniper.master``, then ``COMPOSE_PROJECT_NAME`` will default to ``devstack-juniper.master`` instead of ``devstack``.
+
+
```
ztraboo, line 1433: @angonz Yes I think this works with what @kdmccormick was mentioning for advanced users.

README.rst
```diff
@@ -502,19 +495,19 @@ an empty string.
...
+#. Stop any running containers by issuing a ``make dev.stop``. Note that in older named releases ``make dev.stop`` may not be available. Use ``make stop`` instead. Make sure that all containers are stopped from the docker dashboard.
 #. Follow the steps in `Getting Started`_ section again, setting the additional OPENEDX_RELEASE you want to install in step 2
 
 The implication of this is that you can switch between isolated Devstack databases by changing the value of the ``OPENEDX_RELEASE`` environment variable.
 
 Switch between your Devstack releases by doing the following:
 *************************************************************
 
-#. Stop the containers by issuing a ``make stop`` for the running release.
+#. Stop the containers by issuing a ``make dev.stop`` for the running release. Note that in older named releases ``make dev.stop`` may not be available. Use ``make stop`` instead.
```
ztraboo: @angonz I'm not sure that we need to include for older releases since your PR is going into the `master` release. When I created a PR against another Open edX named release (e.g. open-release/juniper.master) I've changed this command to say `make stop` since that was relevant to that Makefile target for that release. I would remove that sentence to avoid having to make additional comments for all the releases that Open edX has and just keep it specific to the release that you're making the PR for.

cc: @regisb Any additional comments on this?

README.rst
```diff
@@ -502,19 +495,19 @@ an empty string.
 
 How do I run multiple named Open edX releases on same machine?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-You can have multiple isolated Devstacks provisioned on a single computer now. Follow these directions **after installing the master release devstack** to switch between the named releases.
+You can have multiple isolated Devstacks provisioned on a single computer now. Follow these directions **after installing at least two devstacks** to switch between them.
```
ztraboo, line 498: @angonz I think this is fine to say at the very least you need to have two devstacks provisioned.

cc: @regisb Any additional comments on this change?

README.rst
```diff
@@ -502,19 +495,19 @@ an empty string.
...
 
 #. If you haven't done so, follow the steps in the `Getting Started`_ section, to install the master devstack or any other named release. We recommend that you have at least one devstack on the master branch.
 #. Change directory to your devstack and activate the virtual env.
-#. Stop any running containers by issuing a ``make stop``. Make sure that all containers are stopped from the docker dashboard.
+#. Stop any running containers by issuing a ``make dev.stop``. Note that in older named releases ``make dev.stop`` may not be available. Use ``make stop`` instead. Make sure that all containers are stopped from the docker dashboard.
```
ztraboo: @angonz Review my comment https://github.com/edx/devstack/pull/642/files/103fdbf440dc9338518d9699b97ceffd9ec46e11..99e4455f659ef7339ab85983f2b297bc3406b9dc#r504948689 and make the same change by removing this additional sentence for handling older releases for Makefile target for stopping the edX services running.

ztraboo:
@angonz Please review my comments on your 2nd round of changes.
> Looks like this caused a good deal of discussion!! I have submitted a 2nd round of changes.

README.rst
angonz: Then I'd leave it as it is. There is a section about [Using the latest images](https://github.com/angonz/devstack/tree/patch-1#using-the-latest-images), where the `make dev.pull` is shown. I guess it's clear enough and there's no need to repeat here.

README.rst
angonz: I take it. At least, when you notice it doesn't work, then you look at the make help and you'll find out.

README.rst
angonz: done

README.rst
ztraboo: @angonz Yeah however that section looks like it indicates a ‘make down’ right before the ‘make dev.pull’ call. That down command removes Docker containers and rebuilds them. I don’t typically have to do this everytime I’m switching between releases.

@kdmccormick Can you explain the need to issue a ‘make dev.pull’ especially if you altered the MySQL database?

kdmccormick: > When you are switching between devstacks, I guess it's not necessary to pull the latest images, and it takes a lot of time. 

For switching between named releases, running `make dev.pull` is usually skippable. This is because `make dev.up` will automatically pull the named release images for you if you don't already have them. Howvever, other than wait time, there's no *harm* in pulling images. So, for the sake of brevity, I would avoid explicitly recommending `make dev.pull` in the multiple/named releases section, but I also wouldn't discourage people from pulling new images. Just leave it out from that section.

> I have a question here... do you need to provision again if you pull new images? In which cases a new provision is needed?

Generally, **no**, pulling new images should not require a re-provision. The primary thing that new images give you is an environment with updated requirements installed. New images do not contain database changes (except in the rare case where we upgrade the mysql, elasticsearch, or mongo version -- but we usually provide migration instructions in those cases so that re-provisioning isn't required).  Re-provisioning should only be necessary when one wants to throw away their database and start from scratch, which having new images should not necessitate.

> Yeah however that section looks like it indicates a ‘make down’ right before the ‘make dev.pull’ call. That down command removes Docker containers and rebuilds them.

Once new images are pulled, they aren't actually *used* until you bring services down and then up again, hence the `make down` recommendation.

>  I don’t typically have to do this everytime I’m switching between releases.

Which don't you have time for: the image pulling, or the down/up cycle? The image pulling should be a no-op, as the Juniper and Ironwood images will rarely ever change. As for the down/up cycle, though, it seems that it would be necessary to do that when switching named releases due to the container and volume isolation.

kdmccormick (commented):
I'm not able to spend much time on Devstack PRs these days and won't be able to do a full review, so I apologize for this drive-by comment.

But I recommend stepping back and reading through all the instructions in this readme related to named releases and multiple isolated devstacks, and thinking about whether they tell a coherent story to someone who is completely to new this. Right now, they are spread across multiple sections and sub-sections (and I know I recommended moving some of it to Advanced Configuration, so I'm contributing to the problem). If I were new, I bet that I'd find it confusing.

Consider pulling all these topics into one top-level heading about "testing different versions of Open edX via Devstack" or something like that. The section could start with info about switching between Open edX named releases (because that's what 90% of the readers will care about), and explain that switching between releases will use an isolated set of databases volumes. Then, in an "advanced" part of that section, you could touch on how `COMPOSE_PROJECT_NAME` can be modified directly, to the same effect. The crux here is that switching between named releases and directly changing `COMPOSE_PROJECT_NAME` will have almost the exact same set of instructions, and describing them together might minimize confusion.

Thanks for your work @angonz . I know this is a lot of discussion for a relatively small diff, but this is probably the most complex part of devstack, and even I still have a hard time explaining how it works to people :)

Feel free to disregard most or all of my comment. I will leave review to @ztraboo and @regisb , as they've already been doing a good job here.

README.rst
ztraboo: @kdmccormick Sounds good that not a lot of database changes occur on these Docker image updates for name releases so issuing the `make dev.pull` seems ok. 

> That down command removes Docker containers and rebuilds them. I don’t typically have to do this everytime I’m switching between releases.

Here is what I meant by this `"down"` command found at top of this section https://github.com/edx/devstack#using-the-latest-image. I typically am working off of the latest supported release and occasionally switch to `master`. When I meant `"down"` I meant `make down` sorry that I wasn't clear. Not the stopping of the current named release containers (`make dev.stop`) and starting up the next named release containers (`make dev.up`). That stop of existing services for the containers on current working release is necessary because we did not specify new Docker container ports, therefore, if we stood up two release together there would be issues with accessing the port addresses amongst the releases. That's why we turn down one release and stand up another on devstack. Work on one release at a time on devstack.

cc: @angonz

kdmccormick: You weren't unclear, I just didn't even consider doing `make dev.stop` instead of `make dev.down`! Good point.

ztraboo:
@kdmccormick Thanks for suggesting updating that README file. I agree with you that it probably does need a second eye on this from a new perspective.

@angonz Try and do that rework if possible. I can help clarify anything if necessary.

natabene:
@angonz Thank you for your contribution.

natabene:
@regisb Can you take this over the finish line?

README.rst
regisb: Wait, I don't understand this anymore. This comment now implies that the devstack repository should be cloned multiple times in different folders. I thought it was possible to use a single `devstack` folder and switch between the various devstack versions simply by checking out the git branch?

regisb (commented):
I have a single remaining question. Sorry about the back-and-forth @angonz, but we are almost ready to merge this.

README.rst
ztraboo: @regisb I've never tried that before by keeping one devstack directory on local machine and switching between the releases. I do think that is possible if the git branch checkout is applied. Typically I just keep separate directories per release since I'm working on something that's in-progress and don't want to have to worry about stashing my changes before switching branches. That way I can pickup where I left off. But yeah I think you're right in saying that you can have one devstack directory and switch between the releases. Maybe @angonz can test that out and adjust these directions.

regisb: This kind of testing is probably outside the scope of this PR. Let's merge this as-is.

regisb (approved):

openedx-webhooks:
@angonz 🎉 Your pull request was merged!

Please take a moment to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/devstack&entry.867055334=https://github.com/edx/devstack/pull/642&entry.1484655318=https://github.com/angonz&entry.752974735=2020-10-13+16:32&entry.1917517419=2020-10-22+13:58&entry.2133058324=Yes) so we can improve your experience in the future.

PR #645, opened 2020-10-23
Make course-discovery use ES7

feanil:


mikix (approved):

dianakhuang (approved):

docker-compose.yml
```diff
@@ -198,6 +201,7 @@ services:
     hostname: discovery.devstack.edx
     depends_on:
       - mysql57
+      - elasticsearch7
       - elasticsearch
```
mikix: Why not remove the es1 dep?

feanil: This was from when I was trying to query both for debugging. will remove.

PR #649, opened 2020-10-29
Remove analytics-pipeline

macdiesel:
The edx-analytics-pipeline repository is being retired. This PR removes these code bits from devstack. For more info see https://openedx.atlassian.net/browse/DEPR-119


jmbowman (approved):
No complaints on the technical side of things, although the PR should have a description.  Should probably also give a little more time for feedback on [DEPR-119](https://openedx.atlassian.net/browse/DEPR-119) before merging.

macdiesel:
I agree. I'll fix that up.

docker-compose-analytics-pipeline.yml
```diff
@@ -1,155 +0,0 @@
-version: "2.1"
-
-services:
-  namenode:
-    image: edxops/analytics_pipeline_hadoop_namenode:${OPENEDX_RELEASE:-latest}
```
pwnage101, line 5: Maybe we should also delete all the logic to build these images? https://github.com/edx/configuration/tree/master/docker/build

pwnage101 (approved):
Disclaimer: I did not do any grepping, but what I see in this PR looks good.

docker-compose-analytics-pipeline.yml
macdiesel: Good idea!

PR #652, opened 2020-11-05
load-db.sh should load the mysql snapshot into the mysql 5.7 container

iloveagent57:
If you attempt to upgrade edxapp to mysql 5.7, one of the steps is `make dev.provision.lms`, which will end up running `./load-db.sh edxapp`.  Without this change, that'll run against the old mysql 5.6 container, blowing away the presumably good state that you'd like to transfer over to your new mysql 5.7 container/volume.

### DIFF
--- a/load-db.sh
+++ b/load-db.sh
@@ -17,6 +17,6 @@ then
 fi
 
 echo "Loading the $1 database..."
-mysql_container=$(make -s dev.print-container.mysql)
+mysql_container=$(make -s dev.print-container.mysql57)
 docker exec -i "$mysql_container" mysql -uroot $1 < $1.sql
 echo "Finished loading the $1 database!"


load-db.sh
```diff
@@ -17,6 +17,6 @@ then
 fi
 
 echo "Loading the $1 database..."
-mysql_container=$(make -s dev.print-container.mysql)
+mysql_container=$(make -s dev.print-container.mysql57)
```
jhan217, line 20: What happens if the user didn't upgrade their mysql container locally?

iloveagent57: Then the provisioning will fail, I guess.  I'm not exactly sure what the right thing is to do here.  I'm also not in love with the idea of running `load-db.sh` at all.

jhan217: Is it easy to check to see what version of mysql people are on + run the right command for either one?

iloveagent57: I'm not sure, because you can be "on" both versions at the same time, or at least have both running at the same time.

jhan217: In that case, how disruptive would it be to force the user to specify which db they want to load the data into?

kdmccormick (approved):
:dolphin:

PR #659, opened 2020-11-16
Drop support for edx-mktg development.

feanil:
The drupal marketing site is no longer in use so drop support for being
able to develop with it.


regisb (approved):
Out of curiosity, what does edX.org now use to power the marketing website?

feanil:
The site is mostly statically generated now using React and some off the shelf static site generation framework.

PR #669, opened 2020-12-02
Fix typo in mongo data backup

stvstnfrd:
else `make dev.backup` fails for me.


stvstnfrd:
@kdmccormick I'm guessing this was a typo? Or am I missing something else?

regisb (approved):

kdmccormick (approved):
Definitely a typo. Thanks for catching it!

PR #679, opened 2021-01-19
Update mongo3.6 to 4 on devstack

syedimranhassan:


nadeemshahzad (approved):

syed-awais-ali (approved):

regisb:
It looks like CI is failing because we are not using an authenticated docker.io account. Who feels like adding docker.io credentials to the CI build?

PR #684, opened 2021-02-10
Improve devstack defaults for discovery Partner Object

Dillon-Dumesnil:
Adds in the lms course mode api url (unsure if this is used, but
can't hurt), studio url, and publisher url.

Relies on https://github.com/edx/course-discovery/pull/2931

Implements partial functionality of https://openedx.atlassian.net/browse/DISCO-1673


mikix (approved):

Dillon-Dumesnil:
Okay, so the only thing failing is the ecommerce checks and it has consistently failed because it receives no response for > 10 min while running ecommerce migrate. I'm going to merge anyway since that's completely unrelated to this change.

PR #690, opened 2021-02-25
feat: adds scripts to start/stop all enterprise services

chavesj:
### Description

* Adds bash scripts to start/stop the following enterprise services (and their dependencies):
  * LMS (edx-enterprise)
  * ecommerce
  * enterprise-catalog
  * license-manager
* Also starts/stops the enterprise learner portal and admin portal MFEs
  * `npm start &` used to start each MFE as a background job
  * Logs for above command are still output to the terminal, it'd be nice for them not to be
* Script requires the `DEVSTACK_WORKSPACE` environment variable to be defined

Note: This script doesn't start any of the enterprise frontend applications - would we like it to?

### Testing instructions

* Stop all docker containers running on your machine
* Run the new script (`enterprise/up.sh`) and observe all enterprise backend services being started

### DIFF
--- a/enterprise/up.sh
+++ b/enterprise/up.sh
@@ -13,16 +13,16 @@ if [[ $(lsof -i tcp:8734 | grep LISTEN | awk '{print $2}') ]]
 then
   echo "INFO: Enterprise Learner Portal already running"
 else
-  echo "INFO: Silently starting Enterprise Learner Portal"
-  cd ../frontend-app-learner-portal-enterprise/ && npm run --quiet start &
+  echo "INFO: Starting Enterprise Learner Portal"
+  cd ../frontend-app-learner-portal-enterprise/ && npm start &
 fi
 
 if [[ $(lsof -i tcp:1991 | grep LISTEN | awk '{print $2}') ]]
 then
   echo "INFO: Enterprise Admin Portal already running"
 else
-  echo "INFO: Silently starting Enterprise Admin Portal"
-  cd ../frontend-app-admin-portal/ && npm run --quiet start &
+  echo "INFO: Starting Enterprise Admin Portal"
+  cd ../frontend-app-admin-portal/ && npm start &
 fi
 
 exit 0
--- a/enterprise/up.sh
+++ b/enterprise/up.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+if [[ -z "${DEVSTACK_WORKSPACE}" ]]; then
+  echo "DEVSTACK_WORKSPACE environment variable undefined. Exiting..."
+  exit 1
+fi
+
+cd $DEVSTACK_WORKSPACE/devstack/ && make dev.up.lms
+cd ../enterprise-catalog/ && make dev.up
+cd ../license-manager/ && make dev.up
+
+exit 0
--- a/enterprise/up.sh
+++ b/enterprise/up.sh
@@ -5,8 +5,25 @@ if [[ -z "${DEVSTACK_WORKSPACE}" ]]; then
   exit 1
 fi
 
-cd $DEVSTACK_WORKSPACE/devstack/ && make dev.up.lms
+cd $DEVSTACK_WORKSPACE/devstack/ && make dev.up.lms+ecommerce
 cd ../enterprise-catalog/ && make dev.up
 cd ../license-manager/ && make dev.up
 
+if [[ $(lsof -i tcp:8734 | grep LISTEN | awk '{print $2}') ]]
+then
+  echo "INFO: Enterprise Learner Portal already running"
+else
+  echo "INFO: Silently starting Enterprise Learner Portal"
+  cd ../frontend-app-learner-portal-enterprise/ && npm run --quiet start &
+fi
+
+if [[ $(lsof -i tcp:1991 | grep LISTEN | awk '{print $2}') ]]
+then
+  echo "INFO: Enterprise Admin Portal already running"
+else
+  echo "INFO: Silently starting Enterprise Admin Portal"
+  cd ../frontend-app-admin-portal/ && npm run --quiet start &
+fi
+
 exit 0
+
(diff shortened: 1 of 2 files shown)


enterprise/up.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+if [[ -z "${DEVSTACK_WORKSPACE}" ]]; then
+  echo "DEVSTACK_WORKSPACE environment variable undefined. Exiting..."
+  exit 1
+fi
+
+cd $DEVSTACK_WORKSPACE/devstack/ && make dev.up.lms
```
iloveagent57: `lms+redis`

kellmarie: Would there be any harm in doing `make dev.up` instead of just lms? I normally bring up all of devstack + redis.

enterprise/up.sh
```diff
@@ -0,0 +1,12 @@
...
+cd ../enterprise-catalog/ && make dev.up
+cd ../license-manager/ && make dev.up
```
kellmarie, line 10: I wonder if we might want to start other enterprise things too like enterprise-catalog and the front end portals?

enterprise/up.sh
adamstankiewicz: `make dev.up.ecommerce+redis` as some of our enterprise services/MFEs rely on ecommerce (i.e., codes)?

adamstankiewicz:
> Note: This script doesn't start any of the enterprise frontend applications - would we like it to?

@chavesj We might need to backlog a ticket to enable the frontend apps as they are not currently dockerized, so require a longstanding webpack-dev-server process to run in a terminal. If we dockerized the MFEs and/or the dev server process isn't actually an issue), then yes including them in this script would be fantastic :)

enterprise/up.sh
chavesj: `redis` will be started when `enterprise-catalog` or `license-manager` are started so this line should be fine without it. `ecommerce`, on the other hand, needs to be added 🙂. This script is focusing on services that are needed specifically for enterprise so I'm going to continue with `make dev.up.lms+ecommerce`

chavesj:
> > Note: This script doesn't start any of the enterprise frontend applications - would we like it to?
> 
> @chavesj We might need to backlog a ticket to enable the frontend apps as they are not currently dockerized, so require a longstanding webpack-dev-server process to run in a terminal. If we dockerized the MFEs and/or the dev server process isn't actually an issue), then yes including them in this script would be fantastic :)

While dockerization is generally more convenient, it's not the only way to serve our MFEs without hijacking the shell job. If run as a background job (e.g. `npm start &`) we should be able to do so, but we won't have the idempotence that we'd get out of the box with Docker. However, this can still be solved from the shell by only running the background job if no running processes are listening on the port used by the given MFE. 

e.g. `lsof -i tcp:8734 | grep LISTEN | awk '{print $2}'` will return the pid(s) for any process(es) listening on port 8734. This can be used in two ways - to determine when an MFE has already been started and to use in killing the MFE processes. Now I'm starting to see that a `down.sh` script should accompany `up.sh` 🙂 

Dockerizing our MFEs still sounds like a good thing to backlog but the described approach should get us what we want for now.

enterprise/up.sh
chavesj: enterprise-catalog is already part of the script, see later PR comments for context on enterprise MFEs 👍

enterprise/up.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+if [[ -z "${DEVSTACK_WORKSPACE}" ]]; then
+  echo "DEVSTACK_WORKSPACE environment variable undefined. Exiting..."
+  exit 1
+fi
+
+cd $DEVSTACK_WORKSPACE/devstack/ && make dev.up.lms+ecommerce
```
binodpant, line 8: redis too? I have often found redis does not run by default on devstack.. not clear if this has to solve that but ...

enterprise/up.sh
```diff
@@ -0,0 +1,29 @@
...
+cd ../enterprise-catalog/ && make dev.up
+cd ../license-manager/ && make dev.up
+
+if [[ $(lsof -i tcp:8734 | grep LISTEN | awk '{print $2}') ]]
+then
+  echo "INFO: Enterprise Learner Portal already running"
+else
+  echo "INFO: Starting Enterprise Learner Portal"
+  cd ../frontend-app-learner-portal-enterprise/ && npm start &
```
binodpant, line 17: if we backgrounded and there were errors starting up how would we see those? I guess for MFEs this may not be an issue since the browser should show those too.. so yeah maybe nothing.. just thinking out loud. yeah if we dockerize then we can just docker-compose logs too.

binodpant (approved):
love this script!! will we need to setup aliases / paths to run these conveniently,  or will devstack make these scripts available in a different way?

enterprise/up.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+if [[ -z "${DEVSTACK_WORKSPACE}" ]]; then
+  echo "DEVSTACK_WORKSPACE environment variable undefined. Exiting..."
+  exit 1
+fi
+
+cd $DEVSTACK_WORKSPACE/devstack/ && make dev.up.lms+ecommerce
+cd ../enterprise-catalog/ && make dev.up
+cd ../license-manager/ && make dev.up
```
cmltaWt0, line 10: The repo is not cloned into the DEVSTACK_WORKSPACE by default.
I'm wondering do we want to add it to repos or this script is a helper/convenient way to work with manually cloned repos?

enterprise/up.sh
```diff
@@ -0,0 +1,29 @@
...
+  echo "INFO: Starting Enterprise Learner Portal"
+  cd ../frontend-app-learner-portal-enterprise/ && npm start &
+fi
+
+if [[ $(lsof -i tcp:1991 | grep LISTEN | awk '{print $2}') ]]
+then
+  echo "INFO: Enterprise Admin Portal already running"
+else
+  echo "INFO: Starting Enterprise Admin Portal"
+  cd ../frontend-app-admin-portal/ && npm start &
```
cmltaWt0, line 25: The repo is not cloned to workspace by default. Question is the same as above.

PR #691, opened 2021-03-01
fix(lms,studio): set DJANGO_SETTINGS_MODULE in order to fix pylint

kdmccormick:
## Description

Any local use of pylint for LMS or Studio was being met with:
```
E5110: Django was not configured. For more information runpylint --load-plugins=pylint_django --help-msg=django-not-configured (django-not-configured)
```
Newer versions of pylint-django require `DJANGO_SETTINGS_MODULE` to be specified in order to analyze foreign key relationships. So, this commit specifies the settings module in docker-compose.yml.

This issue didn't surface on most IDAs because `DJANGO_SETTINGS_MODULE` was being specified otherwise (in its Dockerfile, or in Ansible, or in docker-compose.yml). This didn't surface in edx-platform PR builds because paver sets `DJANGO_SETTINGS_MODULE` equal to `{lms|cms}.envs.test`.

## Testing

on master:
```
~/devstack> make lms-up-without-deps-shell
root@lms> pylint lms/djangoapps/courseware/utils.py
************* Module lms.djangoapps.courseware.utils
lms/djangoapps/courseware/utils.py:1:0: E5110: Django was not configured. For more information runpylint --load-plugins=pylint_django --help-msg=django-not-configured (django-not-configured)
root@lms>
```
on my branch:
```
~/devstack> make down lms-up-without-deps-shell
root@lms> pylint lms/djangoapps/courseware/utils.py # should succeed (no output)
root@lms>
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -310,6 +310,7 @@ services:
     stdin_open: true
     tty: true
     environment:
+      DJANGO_SETTINGS_MODULE: lms.envs.devstack
       BOK_CHOY_HOSTNAME: edx.devstack.lms
       BOK_CHOY_LMS_PORT: 18003
       BOK_CHOY_CMS_PORT: 18031
@@ -420,6 +421,7 @@ services:
     stdin_open: true
     tty: true
     environment:
+      DJANGO_SETTINGS_MODULE: cms.envs.devstack
       BOK_CHOY_HOSTNAME: edx.devstack.studio
       BOK_CHOY_LMS_PORT: 18103
       BOK_CHOY_CMS_PORT: 18131


stvstnfrd (approved):
I had been manually setting this; thanks for the commit!

kdmccormick:
gah, it breaks `paver update_assets` >.<

kdmccormick:
I ended up in a paver rabbit-hole with no obvious solution. I submitted a CR https://openedx.atlassian.net/browse/CR-3357 in hopes that BOM might know a better solution.

regisb:
Could we simply disable `django-not-configured` pylint warnings? I added this to my .pylintrc because I want to keep to possibility to run pylint on a codebase without defining valid settings (which is a perfectly valid feature).

kdmccormick:
@regisb Perhaps, but then pylint loses the ability to detect errors in [lazy relationships](https://docs.djangoproject.com/en/2.2/ref/models/fields/#foreignkey). 

Although, maybe that's totally fine. We could add it to pylintrc_tweaks. My opinion isn't very strong either way; I just want it to work the same in Devstack and edx-platform CI.

Fyi @jmbowman .

kdmccormick:
looking at https://openedx.atlassian.net/browse/BOM-2448 , it looks like the team is considering disabling this error in pylintrc_tweaks, like you suggest Régis.

awais786:
https://github.com/edx/edx-platform/pull/26938 @kdmccormick FYI

PR #698, opened 2021-03-12
Added Github Action for CI

Jawayria:
Added GitHub workflow for CI to replace Travis

JIRA: https://openedx.atlassian.net/browse/BOM-2455


.github/workflows/ci.yml
```diff
@@ -0,0 +1,64 @@
...
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: installations and version upgrades
+        run: |
+          docker version
+          sudo apt-get update
+          sudo apt install apt-transport-https ca-certificates curl software-properties-common
+          curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+          sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu bionic test"
```
mraarif: we should download `focal` now instead of bionic

provision-xqueue.sh
```diff
@@ -6,8 +6,8 @@ set -x
 docker-compose up -d xqueue
 
 # Update dependencies
-docker-compose exec xqueue bash -c 'source /edx/app/xqueue/xqueue_env && cd /edx/app/xqueue/xqueue && make requirements'
+docker-compose exec -T xqueue bash -c 'source /edx/app/xqueue/xqueue_env && cd /edx/app/xqueue/xqueue && make requirements'
```
mraarif, line 9: maybe add a little explanation why did we need to add this

Jawayria: Added -T to disable pseudo-tty allocation and resolve "The input device is not a tty" error.

.github/workflows/ci.yml
Jawayria: Changed os to ubuntu 16 instead of replacing bionic with focal as travis was using ubuntu 16 currently.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,64 @@
...
+jobs:
+
+  run_ci:
+    runs-on: ${{ matrix.os }}
+    env:
+      DEVSTACK_WORKSPACE: /tmp
+      SHALLOW_CLONE: 1
+    strategy:
+      matrix:
+        os: [ ubuntu-16.04 ]
```
mraarif: should use ubuntu 20 here instead

Jawayria: Previously I mentioned ubuntu 20 here but I was asked to change to ubuntu 16

mraarif: was it failing on ubuntu 20?

Jawayria: It failed on ubuntu 20 when I used focal instead of bionic

mzulqarnain1 (approved):

UsamaSadiq (approved):

PR #708, opened 2021-03-30
feat: Warn developers about accidentally using default-services commands

timmc-edx:
- Pause and warn when pull/provision/up/check are run without a service
  specified
- Introduce a `*.default` variation for when using the default set is
  intentional

We've seen that a lot of developers use `dev.pull` and `dev.up` and then
experience the resulting pain around bandwidth and memory. This is an
experiment in education -- rather than responding in chat when someone
asks where all the RAM has gone, can we guide people away from these
commands in a tighter loop?

ARCHBOM-1672

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -197,7 +197,10 @@ dev.clone.ssh: ## Clone service repos using SSH method to the parent directory.
 # Developer interface: Docker image management.
 ########################################################################################
 
-dev.pull: dev.pull.$(DEFAULT_SERVICES) ## Pull latest Docker images required by default services.
+dev.pull: ## Deprecated: Use dev.pull.default or a set of service names, e.g. dev.pull.lms+studio
+	@scripts/make_warn_generic.sh "$@"
+
+dev.pull.default: dev.pull.$(DEFAULT_SERVICES) ## Pull latest Docker images required by default services.
 
 dev.pull.%: ## Pull latest Docker images for services and their dependencies.
 	docker-compose pull --include-deps $$(echo $* | tr + " ")
@@ -212,7 +215,10 @@ dev.pull.without-deps.%: ## Pull latest Docker images for specific services.
 # Developer interface: Database management.
 ########################################################################################
 
-dev.provision: dev.check-memory ## Provision dev environment with default services, and then stop them.
+dev.provision: ## Deprecated: Use dev.provision.default or a set of service names, e.g. dev.provision.lms+studio
+	@scripts/make_warn_generic.sh "$@"
+
+dev.provision.default: dev.check-memory ## Provision dev environment with default services, and then stop them.
 	# We provision all default services as well as 'e2e' (end-to-end tests).
 	# e2e is not part of `DEFAULT_SERVICES` because it isn't a service;
 	# it's just a way to tell ./provision.sh that the fake data for end-to-end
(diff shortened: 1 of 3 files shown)


scripts/make_warn_generic.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+# Warn the developer that they've run a make command that uses a broad
+# service set and that often is not the best tool for the job.
+
+target="$1"
```
timmc-edx, line 8: I was tempted to use `$target` in the text block in order to customize it, but it would mess up the right margin. (This wouldn't be a problem if we were using the real cowsay command, but we can't rely on it being there.)

Alternatively, just the first line could be spoken by the cow, and the rest printed normally, along with the real make target name.

scripts/make_warn_generic.sh
```diff
@@ -0,0 +1,39 @@
...
+| commands like "make dev.pull.default".)                                 |
+\_________________________________________________________________________/
+       \
+        \   ^__^
+         \  (oo)\_______
+            (__)\       )\/\
+                ||----w |
+                ||     ||
+
+EOCOW
```
nasthagiri, line 21: [love + kudos] on adding "fun" stuff in a developer's day.

Makefile
```diff
@@ -197,7 +197,10 @@ dev.clone.ssh: ## Clone service repos using SSH method to the parent directory.
 # Developer interface: Docker image management.
 ########################################################################################
 
-dev.pull: dev.pull.$(DEFAULT_SERVICES) ## Pull latest Docker images required by default services.
+dev.pull: ## Deprecated: Use dev.pull.default or a set of service names, e.g. dev.pull.lms+studio
+	@scripts/make_warn_generic.sh "$@"
+
+dev.pull.default: dev.pull.$(DEFAULT_SERVICES) ## Pull latest Docker images required by default services.
```
nasthagiri, line 203: [appreciate] I like this approach of introducing a new command so folks can be explicit about wanting all the default services while deprecating the old commands.

[request] Given this awesome approach, can we also look to define a new pared-down definition of `DEFAULT_SERVICES`?  Right now, it's the following list:
```
credentials+discovery+ecommerce+edx_notes_api+forum+frontend-app-payment+frontend-app-publisher+frontend-app-learning+gradebook+lms+studio
```

What if we reduce it to the following list instead, during this redefinition of the commands?
```
discovery+frontend-app-publisher+frontend-app-learning+lms+studio
```

Essentially, we would be removing the following from the "core/default":
```
credentials+ecommerce+edx_notes_api+forum+frontend-app-payment+gradebook
```

As long as we make it easy for devs to create and use their own lists easily, starting with the fewest needed services to startup LMS would work.

We can also do a developer-poll to see what folks think.

timmc-edx: Good question. I don't have enough of a sense of how people use the defaults, or whether they indeed modify them as one part of the docs suggests doing.  But I _suspect_ that we'd be best served by just not relying on the defaults at all.

nasthagiri: Let's validate any assumptions we have with a quick poll to developers. You can do so in discuss.openedx.org and in devstack-questions. We can help you put together the poll, if needed.

I suggest doing that before changing the interface with this PR.

jinder1s: [clarification @nasthagiri] Do you mean hold off on the deprecation of dev.pull/dev.provision? Or do you mean only hold off on  changing/getting rid of DEFAULT_SERVICES?

nasthagiri: hmm... I'm proposing we assess the list of services that will be part of `dev.xxx.default`.  Notes and Forums don't seem necessary to be part of what every developer needs to run, by default.

jinder1s: k, then I will continue reviewing this PR and push for further discovery to happen external to this PR.

scripts/make_warn_generic.sh
jinder1s: I support only saying the first line with cow and rest printed normally. It would be less confusing if target name is said in the prompt rather than be implicit.

timmc-edx: Pushed the change to separate out the first line.

jinder1s (approved):

Makefile
timmc-edx: I think this is something we should groom and prioritize separately, but that we'll get more information as people encounter the change in this PR.

timmc-edx:
Reverted in #712 since it was breaking on some people's computers. Will try out a new branch on Mac as well (oops).

PR #709, opened 2021-03-31
Add a dev.up command for enterprise

raq929:
So people developing enterprise apps don't have to bring up unnecessary containers, and have what they need

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -317,6 +317,9 @@ dev.up.without-deps.shell.%: ## Bring up a service by itself + shell into it.
 	make dev.up.without-deps.$*
 	make dev.shell.$*
 
+dev.up.enterprise: ## Bring up services needed to run enterprise MFEs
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -317,6 +317,9 @@ dev.up.without-deps.shell.%: ## Bring up a service by itself + shell into it.
 	make dev.up.without-deps.$*
 	make dev.shell.$*
 
+dev.up.enterprise: ## Bring up services needed to run enterprise MFEs
+	make dev.up.lms+redis+discovery+ecommerce
```
adamstankiewicz, line 321: Is this any different than just running `make dev.up.ecommerce+redis`?  Starting ecommerce already spins up lms and discovery.

raq929: oh, I didn't know that! yeah, that sounds like a reasonable change.

PR #715, opened 2021-04-02
feat: Provisiong script to award a program certificate in devstack

Tj-Tracy:
This script, along with a few manual steps, will provision a program certificate in devstack. It will add the edX user to a verified seat in the demo course, verify them, issue them a certificate, and then notify credentials. Credentials then awards the user with a program certificate. 

Full explanation (WIP)


credentials/generate_program_certificate.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+echo 'Attempting to award a program certificate to the edX user:'
```
justinhynes: Was this supposed to have any information after the colon?

credentials/generate_program_certificate.sh
```diff
@@ -0,0 +1,28 @@
...
+docker-compose exec -T credentials bash -c 'source /edx/app/credentials/credentials_env && python /edx/app/credentials/credentials/manage.py copy_catalog'
+echo 'creating a program certificate configuration'
+docker-compose exec -T credentials bash -c 'source /edx/app/credentials/credentials_env && python /edx/app/credentials/credentials/manage.py create_program_certificate_configuration'
+
+echo 'Updating LMS...'
+echo 'creating a credentials API connection'
+docker-compose exec -T lms bash -c 'source /edx/app/edxapp/edxapp_env && python /edx/app/edxapp/edx-platform/manage.py lms create_credentials_api_configuration'
+echo 'changing edX user enrollment in demo course from audit to verified'
+docker-compose exec -T lms bash -c 'source /edx/app/edxapp/edxapp_env && python /edx/app/edxapp/edx-platform/manage.py lms change_enrollment -u edx -c course-v1:edX+DemoX+Demo_Course --from audit --to verified'
+echo 'manually verifying edX user'
```
justinhynes: SUPER tiny nit: maybe mention that this step is for id verification? 

"Verified" is an overloaded term and I think this will would differentiate it from the previous step (which is modifying the enrollment/seat type) if anything were to occur.

justinhynes (approved):
👍

credentials/generate_program_certificate.sh
Tj-Tracy: Nope! Just a typo.

credentials/generate_program_certificate.sh
Tj-Tracy: Changed it to "Manually ID verifying the edX user"

PR #716, opened 2021-04-05
fix: Update reset warning to be less alarming; exit 1 if declined

timmc-edx:
In PR #637 I made the reset command safer by refusing to discard any
uncommitted changes, but I never updated the warning message.

Also:

- Move warning and consent check closer to the code that the warning
  applies to so that it can be kept in sync more easily
- Exit with an error status if declined, to be consistent with other
  error exits.
- Clarify other log messages

Checklist:

- Tested on both Mac and Linux (if changed Makefile or shell scripts)
    - [x] Mac - `make dev.reset-repos` and check if `y`/`n`/[some garbage] response behaves as expected
    - [x] Linux
    - N/A

### DIFF
--- a/repo.sh
+++ b/repo.sh
@@ -168,19 +168,26 @@ clone_private ()
 
 reset ()
 {
+    read -p "This will switch to master and pull changes in your local git checkouts. Would you like to proceed? [y/n] " -r
+    if [[ ! $REPLY =~ ^[Yy]$ ]]; then
+        echo "Cancelling."
+        exit 1
+    fi
+
     for repo in ${repos[*]}
     do
         [[ $repo =~ $name_pattern ]]
         name="${BASH_REMATCH[1]}"
 
         if [ -d "$name" ]; then
+            # Try to switch branch and pull, but fail if there are uncommitted changes.
             (cd "$name"; git checkout -q master && git pull -q --ff-only) || {
                 echo >&2 "Failed to reset $name repo. Exiting."
                 echo >&2 "Please go to the repo and clean up any issues that are keeping 'git checkout master' and 'git pull' from working."
                 exit 1
             }
         else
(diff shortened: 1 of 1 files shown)


feanil (approved):

repo.sh
```diff
@@ -168,19 +168,26 @@ clone_private ()
...
+    read -p "This will switch to master and pull changes in your local git checkouts. Would you like to proceed? [y/n] " -r
+    if [[ ! $REPLY =~ ^[Yy]$ ]]; then
+        echo "Cancelling."
+        exit 1
+    fi
+
     for repo in ${repos[*]}
     do
         [[ $repo =~ $name_pattern ]]
         name="${BASH_REMATCH[1]}"
```
jinder1s, line 181: What do you think of adding this here: 
` printf "Starting to reset [%s] repo. \n" "$name"`

jinder1s:
I tested this in mac and it seemed to work fine. Feel free to merge with or without addressing my comment above.

repo.sh
jinder1s: I got the following warning while running this code: 
`Can't find Husky, skipping post-checkout hook
You can reinstall it using 'npm install husky --save-dev' or delete this hook`

This was not captured in any of the logging for this. I added the printf statement to see which repo this was occuring in.

jinder1s: Not sure where we want to be in terms to opting for minimum output

timmc-edx: Might be a bit noisy. What would be the benefit -- knowing which repo had issues?

timmc-edx: I see -- yeah, I think I'll skip it for now, since that particular warning doesn't seem to be actually interfering with the reset.

jinder1s: Yep, you would only gain better understanding of which repo had issues. Since we already catch the errors and report in which repo they are happening, I think it's not necessary.

PR #718, opened 2021-04-05
feat: Warn developers about accidentally using default-services commands

timmc-edx:
- Pause and warn when `dev.pull` or `dev.up` are run without a service specified
- Introduce a `*.large-and-slow` variation for when using the default set is intentional
- Improve documentation around when to use the large set or a specific service; update some doc locations to use one or the other more explicitly

We've seen that a lot of developers use `dev.pull` and `dev.up` and then experience the resulting pain around bandwidth and memory. This is an experiment in education -- rather than responding in chat when someone asks where all the RAM has gone, can we guide people away from these commands in a tighter loop?

(Implementation note: Keeping the main text out of the cowsay allows us to have variable length text (interpolate the make target) without messing up the speech bubble borders or having to install the cowsay package.)

The first commit is cherry-picked from PR #708 (commit 80ed748) which was reverted for a bugfix; if you previously reviewed that PR, see the second commit's message for a description of what has changed since then.

----

    - Already tested on: Linux
    - Testing instructions:
        - Run `make dev.pull.lms` and confirm it does not warn and does pull the right things
        - Run `make dev.pull` and `make pull` and confirm they give the warning (can cancel out of it)
        - Run `make dev.pull` and allow it to continue, and confirm that it does not give an error after running to completion
        - Run `make dev.pull.large-and-slow` and confirm that it does *not* warn
        - Run `make dev.reset` (or any other all/most-services workflow depending on up/pull) and confirm you are not warned
        - If you're feeling lucky, also test `make dev.up`
    - Will _inform_ the parties listed in "All Open edX developers" on How We Announce and ask for feedback
    - Will also ask for contributions to the "common combinations" list (MFEs? other services usually run together?)

### DIFF
--- a/scripts/make_warn_default_large.sh
+++ b/scripts/make_warn_default_large.sh
@@ -41,6 +41,6 @@ the command "make $target.large-and-slow".)
 
 EOF
 
-read -r -p $'(You can cancel the command now or press ENTER to continue.)\n'
+read -r -p $'(You can cancel the command now with Ctrl-C or press ENTER to continue.)\n'
 
 make --no-print-directory "$target.large-and-slow"
--- a/scripts/make_warn_default_large.sh
+++ b/scripts/make_warn_default_large.sh
@@ -1,9 +1,10 @@
 #!/bin/bash
-# Warn the developer that they've run a make command that uses a broad
-# service set and that often is not the best tool for the job.
+# Warn the developer that they've run a make command that uses a very
+# large set of services and that often is not the best tool for the
+# job.
 #
 # This script is used in the Makefile for commands that should be run
-# as `make $target.default` instead.
+# as `make $target.large-and-slow` instead if that's what's intended.
 
 target="$1"
 
@@ -25,7 +26,7 @@ The command "make $target" will operate on a large default set of
 services and their dependencies. This can make your task take longer
 than necessary.
 
-You may prefer to use something like "make $target.lms+studio" to
+You may prefer to use something like "make $target.lms" to
 target a smaller set of services.  Learn more about the commands you
 can run at:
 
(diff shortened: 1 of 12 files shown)


jinder1s:
"Run make dev.pull and make pull and confirm they give the warning (can cancel out of it)" 
Is the best way to cancel out by using `ctrl-c`??

jinder1s:
Okay, I followed the test instructions about and everything worked as intended.

timmc-edx:
Yes, I had left it unspecified because I wasn't sure what the key combo was on Mac -- if it's Ctrl-C on both Mac and Linux, specifying explicitly makes sense. Thanks!

scripts/make_warn_default_large.sh
```diff
@@ -0,0 +1,46 @@
...
+Without an explicit list of services, many devstack Make targets pull
+down Docker images you don't need or take up extra memory and CPU. You
+might even run into bugs in unrelated services.
+
+(If you *really* want the large default set of services, you can use
+the command "make $target.large-and-slow".)
+
+EOF
+
+read -r -p $'(You can cancel the command now or press ENTER to continue.)\n'
```
jinder1s: It might be a good idea to give instructions on how to cancel? Though, it can be argued most developers know how to use ctrl-c

jinder1s (approved):

jinder1s:
This might be a good candidate for ADR? You'd mostly be coping PR description to ADR.

Based on parking lot, this can go in so many ways. At a minimum, commit msg with a lot of the stuff from PR description.

timmc-edx:
Good point -- I'll drop in a quick ADR.

timmc-edx:
One thing is bothering me -- there's some chance this will mess up a non-interactive/non-terminal workflow due to the blocking `read` call. One option would be to skip the warn-and-block if someone sets an environment variable, perhaps `DEVSTACK_NO_WARN_LARGE_SET`. Then anyone who runs into this would have a quick workaround that does not rely on changing any checked-in files.

But it occurs to me that the script could also skip if it sees that an `options.local.mk` exists and contains `DEFAULT_SERVICES`, which is documented in the README as a way to alter that variable. If someone has that set, they probably know what they're doing and *aren't* using the default large set. One downside: Different people would see different behavior and might get confused by that.

timmc-edx:
...I think I've talked myself out of that. Anyone running such a job on their own computer can locally revert the change; anyone running such a job on a CI server or similar probably won't be able to change the environment variable any more easily than just using a different make target in their script.

timmc-edx:
Let me know if the ADR looks good.

jinder1s (approved):

jinder1s:
ADR looks good

PR #723, opened 2021-04-08
feat: Skip large-set warning if DEFAULT_SERVICES is overridden

timmc-edx:
Some developers make use of the option of overriding `DEFAULT_SERVICES` in
options.local.mk (a git-ignored overrides file) as mentioned in the docs.
These people presumably know what they're doing and have already selected
a smaller set of services, so there's no point in bothering them with a
warning.

----

    - Already tested on: Linux
    - Testing instructions: Run `make dev.pull` with and without an `options.local.mk` containing `DEFAULT_SERVICES = lms`
    - Not needed for this tweak

### DIFF
--- a/scripts/make_warn_default_large.sh
+++ b/scripts/make_warn_default_large.sh
@@ -8,7 +8,8 @@
 
 target="$1"
 
-cat <<"EOCOW" >&2
+show_warning_and_wait() {
+    cat <<"EOCOW" >&2
  _________________________________________________________________________
 /                                                                         \
 | Are you sure you want to run this command for *all* Open edX services?  |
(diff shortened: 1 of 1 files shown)


jinder1s:
I thought you decided against this in last PR?

robrap (approved):
Worked on my mac.
1. @jinder1s has a question.
2. At what point will we introduce tests?

timmc-edx:
I had decided against `DEVSTACK_NO_WARN_LARGE_SET` but didn't have as strong a feeling about options.local.mk: https://github.com/edx/devstack/pull/718#issuecomment-815120905

My only concern about the latter was that cifferent people would see different behavior and might get confused by that. But by printing the "skipping the warning" line, I think we can help avoid that situation.

Currently there are tests of make targets that explicitly list services, but having tests for other cases would be nice. I'm not actually sure how we'd go about shoehorning a test for this into the current CI setup, though.

jinder1s (approved):
That makes sense, Thanks!

PR #728, opened 2021-04-27
test: Test a few CLI interactions, starting with large-set warning

timmc-edx:
This adds basic CI testing for the large-set warning introduced in
https://github.com/edx/devstack/pull/718 (ARCHBOM-1672).

Tests can be run locally if `expect` is installed.

Also turn off fail-fast in Github Actions; it's not uncommon for one
test to be flaky, and this allows us to get results from the other tests
even if one fails early.

ref ARCHBOM-1752

----

    - Already tested on: Linux
    - Testing instructions: Install `expect`, run new test files as scripts (as in CI file)


feanil (approved):

timmc-edx:
Merging without testing on Mac, since this is aimed primarily at CI.

PR #732, opened 2021-05-03
feat: Adding performance data collection

jinder1s:
Collecting data on thes make targets: dev.pull and dev.up.%
The data collected:
- the time make target is called
- how long it took for target to finish
- the name of the target

Currently, the data is only collected, but not saved. For now, it is
just printed out to stderr. We plan to eventually send it to segment as
part of follow up work.

[ARCHBOM-1766](https://openedx.atlassian.net/browse/ARCHBOM-1766)

----

    - Already tested on: mac by jinder
    - Testing instructions: *[commands and expected behavior]*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -283,6 +283,7 @@ dev.up.large-and-slow: dev.up.$(DEFAULT_SERVICES) ## Bring up default services.
 
 dev.up.%:
 	@scripts/send-metrics.py "dev.up.$*"
+
 impl-dev.up.%: dev.check-memory ## Bring up services and their dependencies.
 	docker-compose up -d $$(echo $* | tr + " ")
 ifeq ($(ALWAYS_CACHE_PROGRAMS),true)
--- a/Makefile
+++ b/Makefile
@@ -58,8 +58,7 @@
         dev.sync.requirements dev.sync.up dev.up dev.up.attach dev.up.shell \
         dev.up.without-deps dev.up.without-deps.shell dev.up.with-programs \
         dev.up.with-watchers dev.validate docs e2e-tests e2e-tests.with-shell \
-		impl-dev.pull \
-        help requirements selfcheck upgrade upgrade \
+        help requirements impl-dev.pull selfcheck upgrade upgrade \
         validate-lms-volume vnc-passwords
 
 # Load up options (configurable through options.local.mk).
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,7 @@
         dev.sync.requirements dev.sync.up dev.up dev.up.attach dev.up.shell \
         dev.up.without-deps dev.up.without-deps.shell dev.up.with-programs \
         dev.up.with-watchers dev.validate docs e2e-tests e2e-tests.with-shell \
+		impl-dev.pull \
         help requirements selfcheck upgrade upgrade \
         validate-lms-volume vnc-passwords
 
@@ -198,7 +199,10 @@ dev.clone.ssh: ## Clone service repos using SSH method to the parent directory.
 ########################################################################################
 
 dev.pull:
-	@scripts/make_warn_default_large.sh "$@"
+	@scripts/send-metrics.py "$@"
+
+impl-dev.pull: ##
+	@scripts/make_warn_default_large.sh "dev.pull"
 
 dev.pull.large-and-slow: dev.pull.$(DEFAULT_SERVICES) ## Pull latest Docker images required by default services.
 	@echo # at least one statement so that dev.pull.% doesn't run too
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -278,7 +281,9 @@ dev.up:
 dev.up.large-and-slow: dev.up.$(DEFAULT_SERVICES) ## Bring up default services.
 	@echo # at least one statement so that dev.up.% doesn't run too
 
-dev.up.%: dev.check-memory ## Bring up services and their dependencies.
+dev.up.%:
+	@scripts/send-metrics.py "dev.up.$*"
```
timmc-edx, line 285: I'm not sure if it matters, but I think we want a blank line between these.

scripts/send-metrics.py
```diff
@@ -0,0 +1,43 @@
...
+    exit_code = completed_process.returncode
+    time_diff = t1.timestamp() - t0.timestamp()
+    output = {"target":make_target, "t0":t0, "time_diff":time_diff, "exit_code": exit_code}
+    print(f"send metrics info: {output}", file=sys.stderr)
+
+def run_target(make_target):
+    return subprocess.run(["make", f"impl-{make_target}"])
+
+def main(make_target):
+    # Collect data only if user has concented to data collection by creating a file named: ~/.config/devstack/metrics.json
```
timmc-edx:
```suggestion
    # Collect data only if user has consented to data collection by creating a file named: ~/.config/devstack/metrics.json
```
nit: typo

timmc-edx:
Tested on Linux with `make dev.pull` and `make dev.up.forum` with/without blank config file; saw metrics printout only when config file was in place.

timmc-edx (approved):

regisb (commented):
Please don't use this pattern for other make commands. This breaks compatibility with make options (`-j`, `-B`, `--directory`) and arguments (`KEY=VALUE`). The `send-metrics.py` script should not be responsible for running `make` commands. Instead, the start and end times of a command can be stored as environment variables, which are then transmitted (or not) by the `send-metrics.py` script.

timmc-edx:
Hmm, good point. Do you have examples of how this interferes with devstack-specific workflows, or is this just on general principle? (For example, I would be surprised to hear of people using `-B` with devstack; every target should be in the `.PHONY` list.)

We don't have a lot of options for how to wrap or instrument make targets, but there might be some way we can pass along arguments to the wrapped invocation.

timmc-edx:
(Note that we also don't plan on using this specific `X`/`impl-X` pattern for large numbers of targets -- this is for a prototype. However, the wrapped-invocation pattern might continue even if we use a different method of indirection.)

timmc-edx:
Good news! GNU Make (and probably other implementations) sets a `MAKEFLAGS` environment variable which the inner invocation is able to see, so almost all command line options will be inherited. (This excludes `--directory` and `--makefile`.) You can observe this with `make -s dev.up.lms`.

regisb:
I can certainly find some examples of commands that will fail with this specific implementation, and we should be able to find solutions that address these issues -- so that's not really the point. The thing is that the wrapping script will eventually become more and more complex, and less readable. More generally speaking, adding yet another level of indirection makes it difficult to understand and debug commands. Devstack is already quite complex and includes multiple levels of indirection, between make, bash, docker and docker-compose commands.

A development environment is a low-level piece of code: users should be able to copy-paste some commands and have clear expectations of the outcome. Wrapping bash commands in Python code adds yet another layer of complexity that developers need to investigate when they face an issue.

Instead of using a custom Python script, can I propose the following approach, which uses the more familiar `time` command?

Create a `make` file:

```
#! /usr/bin/sh
MAKE=$(pwd)/make /usr/bin/time --append --format="%E\t%C" --output=devstack-metrics.log make $@
curl -X POST -d @devstack-metrics.log http://devstackreporting.edx.org
rm -f devstack-metrics.log
```

Run all profile commands with `./make` instead of `make`. This is a lightweight approach that makes it very easy both to opt-in and -out of the profiling. Run `./make` instead of `make` to opt-in, and revert to `make` if profiling causes issues.

timmc-edx:
First, I should note that this is just a prototype -- we're not planning on instrumenting everything this way, or even more than a handful of targets. We considered a bunch of different options (I'll share a doc with you showing some of the brainstorming we did) and this actually wasn't our top choice, just a way to get some early metrics. Long-term we'd want to use something less obtrusive as far as the Makefile itself is concerned.

I think that switching to a different command line invocation is a non-starter, unfortunately -- we wouldn't get very little data if people had to remember to use a different (longer) command each time. And I'm no great fan of using make for these purposes in general, but it does confer some advantages (well, just one: tab-completion) such that moving away from it would be a hard sell.

PR #735, opened 2021-05-05
feat: Send devstack metrics to Segment (if locally enabled)

timmc-edx:
- Gating changes from "does config file exist" to "does config file contain a Segment write key"
- try-blocks wrapped around everything that isn't running the user's intended command
- Most of our attributes are in place; still need anonymous user ID and an attribute for marking events resulting from automated tests

ARCHBOM-1761

----

    - Already tested on: Linux
    - Testing instructions:
        - Make sure the config file `~/.config/devstack/metrics.json` is *missing*
        - Run `DEVSTACK_METRICS_TESTING=debug make dev.up.redis` and confirm that it works as it would normally
            - You should also notice an error message about config failure; this should not appear if that env var is not set
        - Set up the config file with `{"segment_write_key":"MUy..."}` (get write key from team)
        - Run `DEVSTACK_METRICS_TESTING=debug make dev.up.redis` and confirm some metrics are printed out this time
        - Confirm that an anonymous user ID has appeared in the config file
        - Run `make dev.up.something_unique` and confirm that you see the event in New Relic (`select * from DevstackMetrics since 2 days ago`)
    - None needed yet, still not enabled

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,8 @@ jobs:
     env:
       DEVSTACK_WORKSPACE: /tmp
       SHALLOW_CLONE: 1
+      # Don't report metrics as real usage
+      DEVSTACK_METRICS_TESTING: ci
     strategy:
       matrix:
         os: [ ubuntu-latest ]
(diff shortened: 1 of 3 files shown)


jinder1s (approved):

jinder1s:
Actually, hold off on merging. I failed to see my input in new relic dashboard.

timmc-edx:
I see it, but we should figure out why you're not. :thinking:

jinder1s:
Okay,  I see it now. You are now clear to merge

PR #736, opened 2021-05-06
test: Fix corner case in warn-default test around early exit of process

timmc-edx:
Previously this test would only fail if the spawned make command timed
out, but wouldn't fail if it exited without providing the expected text.
This covers that case by using the `eof` keyword.

----

    - Already tested on: Linux
    - Testing instructions:
        - Have `expect` installed
        - Run `./tests/warn_default.exp; echo $?` and confirm the last output is `0` (success)
        - Change the script's spawn line to `spawn ls`
        - Re-run and confirm the last output is `1` (error)

### DIFF
--- a/tests/warn_default.exp
+++ b/tests/warn_default.exp
@@ -8,12 +8,14 @@ spawn make dev.pull
 expect {
     "Are you sure you want to run this command" {}
     timeout { exit 1 }
+    eof { exit 1 }
 }
 send "\n"
 
 expect {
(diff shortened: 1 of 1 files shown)


robrap (approved):

timmc-edx:
Merging before all tests have finished (since it's the same for all, and some have passed) and without a Mac check (since this is a core Expect feature and only *really* needs to work on CI anyhow).

PR #739, opened 2021-05-07
 build: Make a test.in file and upgrade dependencies (foundation for tests)

timmc-edx:
This allows us to keep the tests in our usual language rather than adding a new one. (Thanks Jeremy for finding this library!)

I've also moved the CLI tests to earlier in the CI run so that we get faster feedback and don't have to wait for cloning first. Really, they should be their own workflow file, but that's work for another day.

There is also a foundational commit for adding a `test.in` and upgrading deps before adding anything new.

----

    - Already tested on: Linux
    - Testing instructions:
        - These are run by CI, so you should see if they pass that way. But if you'd like to test that they can fail properly, try changing the spawn command in each test to be `make help` or something else that won't behave as the test expects it to. You should get an EOF or timeout failure.
        - There are preconditions for running one of the tests, but it will tell you what they are. :-)


jinder1s (approved):
This looks good except for the ecommerce build failing. And I think that's just cause ecommerce is flaky.

tests/metrics.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    Test that dev.up.% is instrumented for metrics collection.
+    """
+
+    # Some setup and preconditions first...
+
+    assert os.environ.get('DEVSTACK_METRICS_TESTING'), \
+        "You need a DEVSTACK_METRICS_TESTING=debug if running this test " \
+        "locally since this environment variable both enables printing of " \
+        "metrics and also marks sent metric events as test data."
```
robrap, line 17: Is it important to mark data as test data if we are using a fake write key?

tests/metrics.py
```diff
@@ -0,0 +1,43 @@
...
+    # will collect metrics.
+    os.makedirs(config_dir, exist_ok=True)
+    with open(config_path, 'w') as f:
+        f.write(json.dumps({"segment_write_key": "fake"}))
+
+    # OK, the actual test:
+
+    try:
+        p = pexpect.spawn('make dev.up.redis', timeout=60)
+        p.expect(r'Send metrics info:')
```
robrap, line 35: This all looks good.  Did you test the test by ensuring this breaks if any of the expected behavior doesn't happen?

robrap (approved):
No blockers. Just questions. Thanks.

tests/metrics.py
timmc-edx: Belt and suspenders. The roadmap for the epic eventually has us including a real write key in devstack itself rather than reading it from the config file, at which point only the environment variable would provide this protection.

tests/metrics.py
timmc-edx: Yep, as described in the testing instructions in the PR checklist.

PR #743, opened 2021-05-12
feat!: Change metrics is_test default from 'false' to 'no'

timmc-edx:
This is to work around a bug in New Relic dashboards where an attempt to
filter on the string false turns into a filter on a boolean false. It might
also be clearer to read.

Also adds a more detailed test of metrics data, both expected values and
a total allowed set of keys. (Includes test of `is_test` attribute.)

----

    - Already tested on: Linux
    - Testing instructions: Confirmed that test failed with env var `DEVSTACK_METRICS_TESTING=no`

### DIFF
--- a/scripts/send-metrics.py
+++ b/scripts/send-metrics.py
@@ -91,7 +91,11 @@ def send_metrics_to_segment(event_properties, config):
 
     May throw.
     """
-    event_properties = dict(is_test=test_mode or 'no', **event_properties)
+    # Get a shallow copy of the input and annotate it as a non-test
+    # event unless overridden by environment variable.
+    event_properties = event_properties.copy()
+    event_properties['is_test'] = test_mode or 'no'
+
     event = {
         'event': 'devstack.command.run',
         'userId': config['anonymous_user_id'],
--- a/scripts/send-metrics.py
+++ b/scripts/send-metrics.py
@@ -91,7 +91,7 @@ def send_metrics_to_segment(event_properties, config):
 
     May throw.
     """
-    event_properties = dict(**event_properties)
+    event_properties = dict(is_test=test_mode or 'no', **event_properties)
     event = {
         'event': 'devstack.command.run',
         'userId': config['anonymous_user_id'],
@@ -100,7 +100,6 @@ def send_metrics_to_segment(event_properties, config):
     }
 
     if test_mode:
-        event_properties['is_test'] = test_mode
         print(f"Send metrics info: {json.dumps(event)}", file=sys.stderr)
 
     # https://segment.com/docs/connections/sources/catalog/libraries/server/http-api/
@@ -148,7 +147,6 @@ def run_wrapped(make_target, config):
             'start_time': start_time.isoformat(),
             'duration': time_diff_millis,
             'exit_status': exit_code,
-            'is_test': 'no',
         }
         send_metrics_to_segment(event_properties, config)
     except Exception as e:
--- a/scripts/send-metrics.py
+++ b/scripts/send-metrics.py
@@ -148,7 +148,7 @@ def run_wrapped(make_target, config):
             'start_time': start_time.isoformat(),
             'duration': time_diff_millis,
             'exit_status': exit_code,
-            'is_test': 'false',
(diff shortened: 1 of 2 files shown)


scripts/send-metrics.py
```diff
@@ -91,7 +91,7 @@ def send_metrics_to_segment(event_properties, config):
 
     May throw.
     """
-    event_properties = dict(**event_properties)
+    event_properties = dict(is_test=test_mode or 'no', **event_properties)
```
robrap: [non-blocking] 
1. The code is probably fine as-is, but I'm not sure whether people might set `test_mode` incorrectly. 
2. I find the `dict()` call takes my brain a little longer to process then just setting the key, unless I am missing something.

The following is one possible update, that could be on 2 lines or 1.
```suggestion
    is_test = 'no' if not test_mode or test_mode == 'false' else test_mode
    event_properties['is_test'] = is_test
```

tests/metrics.py
```diff
@@ -82,4 +82,25 @@ def test_metrics():
...
+        assert sorted(data['properties'].keys()) == [
+            'command', 'command_type', 'duration',
+            'exit_status', 'is_test', 'start_time'
+        ], "Unrecognized attribute -- confirm that this addition is authorized."
+
+        assert data['event'] == 'devstack.command.run'
+        assert data['properties']['command'] == 'dev.up.redis'
+        is_test = data['properties']['is_test']
+        assert type(is_test) == str
+        assert is_test != 'no'
```
robrap: [non-blocking] Might be simpler to simply check that this equals what you expect it to equal. I'm guessing you are trying to make this resilient to changing that value, but not sure that adds much, and it is a lot more to read than:
```
assert data['properties']['is_test'] == 'WHATEVER_THIS_IS'
```

robrap (approved):

scripts/send-metrics.py
timmc-edx: I'm not too worried about `test_mode` since the only time it should be set is in CI tests or when a dev is running the tests manually, and the tests specifically ask you to set `DEVSTACK_METRICS_TESTING=debug` if you haven't set it.

On the dict, I still haven't figured out a great way to express "add these two dictionaries together" in Python. I come from a functional programming background where I could normally just do something in the spirit of `event_properties += {"is_test": ...}`, but apparently there's no natural facility for that in Python. (What I've written is close, but will throw if the key already exists.)

The reason I haven't used `event_properties['is_test'] = ...` is that that would entail mutating the input, which *currently* would have no effect on the running of the program but is the kind of "spooky action at a distance" I try very hard to avoid if I can help it, because it makes reasoning about the code so much harder.

What would you think of this?

```python
event_properties = event_properties.copy()  # shallow copy of input
event_properties['is_test'] = test_mode or 'no'  # mark as non-test by default
```

tests/metrics.py
timmc-edx: I wrote it this way to allow values of both `'ci'` (the usual value) and `'debug'` (what a dev running manually will probably be using). I could make it more explicit and force the dev to use `debug`:

```python
assert data['properties']['is_test'] in ['ci', 'debug']
```

scripts/send-metrics.py
robrap: I now see why you had the original code. I like this latest suggestion, but am ok with whatever you like best.

tests/metrics.py
robrap: I like this last suggestion, but I'm ok with this or the original. Just expressing my opinion as a non-blocking comment.

scripts/send-metrics.py
timmc-edx: OK, I'll go with that, and try to make sure it conveys intent better than the original code did. :-)

tests/metrics.py
timmc-edx: I appreciate the nudge -- I wasn't super thrilled with the code as written, and I think this would be an improvement as well.

timmc-edx:
(Test failures relate to squashed migrations that are fixed on another branch, and are unrelated to the CLI tests or metrics.)

PR #747, opened 2021-05-13
feat: Add metrics for git commit timestamp and , whether on master

timmc-edx:
`git_commit_time`:

This discovers the timestamp of the currently checked out commit and
includes it in the metrics so that we know how old a version of devstack
the developer is using. This has use both in correlating behavior changes
to devstack version and in seeing whether people use old versions of
devstack.

`git_checked_out_master`:

This allows us to filter out events made on other branches, since we’re
not sure how much time people spend on branches. While it is derivable
from `git_commit_time` plus a copy of the repo’s public reflog, collecting
it this way is useful because New Relic can’t integrate that public data
readily.

----

    - Already tested on: Linux
    - Testing instructions: With collection enabled, run `DEVSTACK_METRICS_TESTING=debug make dev.up.redis` and confirm that output of git attrs looks correct.

### DIFF
--- a/scripts/send-metrics.py
+++ b/scripts/send-metrics.py
@@ -140,13 +140,14 @@ def read_git_state():
     # timestamp; the latter could be older when commits have been
     # rewritten, and the former is more likely to be of interest when
     # looking at repo checkout age.
-    p = subprocess.run(
+    process = subprocess.run(
         ['git', 'show', '--no-patch', '--pretty=format:%cI|%D'],
         capture_output=True, check=True, timeout=5
     )
-    timestamp, reflist = p.stdout.decode().split('|', 2)
-    # Returns true if master is currently checked out. Does not return
-    # true in the following similar situations:
+    timestamp, reflist = process.stdout.decode().split('|', 2)
+    # Returns true if master is currently checked out. Returns false
+    # otherwise, which includes the following situations that are similar
+    # to, but different from a master checkout:
     #
     # - Detached-head state which happens to be on same commit as master
     # - Another branch is checked out but points to the same commit as
--- a/scripts/send-metrics.py
+++ b/scripts/send-metrics.py
@@ -131,6 +131,35 @@ def send_metrics_to_segment(event_properties, config):
             traceback.print_exc()
 
 
+def read_git_state():
+    """
+    Return additional, git-related attributes to be merged into event
+    properties.
+    """
+    # %cI gets the committer timestamp, rather than the author
+    # timestamp; the latter could be older when commits have been
+    # rewritten, and the former is more likely to be of interest when
+    # looking at repo checkout age.
+    p = subprocess.run(
+        ['git', 'show', '--no-patch', '--pretty=format:%cI|%D'],
+        capture_output=True, check=True, timeout=5
+    )
+    timestamp, reflist = p.stdout.decode().split('|', 2)
+    # Returns true if master is currently checked out. Does not return
+    # true in the following similar situations:
+    #
+    # - Detached-head state which happens to be on same commit as master
+    # - Another branch is checked out but points to the same commit as
+    #   master
+    # - On commit which *used to* be the tip of master, but is no longer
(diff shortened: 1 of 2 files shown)


scripts/send-metrics.py
```diff
@@ -131,6 +131,35 @@ def send_metrics_to_segment(event_properties, config):
...
+    timestamp, reflist = p.stdout.decode().split('|', 2)
+    # Returns true if master is currently checked out. Does not return
+    # true in the following similar situations:
+    #
+    # - Detached-head state which happens to be on same commit as master
+    # - Another branch is checked out but points to the same commit as
+    #   master
+    # - On commit which *used to* be the tip of master, but is no longer
+    #   (and is just in master's history)
+    is_at_master = "HEAD -> master" in reflist.split(', ')
```
timmc-edx, line 157: Note for reviewer: Depending on what we want, we could instead check if any of the entries is in `["master", "HEAD -> master"]` or even look for `"origin/master"` (although that depends on the dev's remote names, which is more work). For example, if someone has done `git checkout master; git pull; git checkout -b my-branch`, then this list will contain `master, origin/master, HEAD -> my-branch`. Technically they'd be on another branch, but they'd be on the same _commit_ as master. We probably still want to exclude this from some analysis, though.

It's also possible someone could have a different local branch name for master, but I don't know if that's a thing people do with non-feature-branches.

scripts/send-metrics.py
```diff
@@ -131,6 +131,35 @@ def send_metrics_to_segment(event_properties, config):
...
+def read_git_state():
+    """
+    Return additional, git-related attributes to be merged into event
+    properties.
+    """
+    # %cI gets the committer timestamp, rather than the author
+    # timestamp; the latter could be older when commits have been
+    # rewritten, and the former is more likely to be of interest when
+    # looking at repo checkout age.
+    p = subprocess.run(
```
robrap: nit: I like longer names like `process`(?) instead of `p`.

scripts/send-metrics.py
```diff
@@ -131,6 +131,35 @@ def send_metrics_to_segment(event_properties, config):
...
+        ['git', 'show', '--no-patch', '--pretty=format:%cI|%D'],
+        capture_output=True, check=True, timeout=5
```
robrap, line 145: What happens, and what  do we want to happen if this times out for some odd reason?

scripts/send-metrics.py
```diff
@@ -131,6 +131,35 @@ def send_metrics_to_segment(event_properties, config):
...
+    )
+    timestamp, reflist = p.stdout.decode().split('|', 2)
+    # Returns true if master is currently checked out. Does not return
+    # true in the following similar situations:
```
robrap: How about something like the following?:
```suggestion
    # Returns true if master is currently checked out. Returns false
    # otherwise, which includes the following situations that are similar to,
    # but different from a master checkout:
```

scripts/send-metrics.py
robrap: I think it is fine. I wouldn't be overly concerned with this until we are getting back data and have real questions which either are or are not fully answered by this.

robrap (commented):
Nice progress.

scripts/send-metrics.py
timmc-edx: Sure, `process` works.

scripts/send-metrics.py
timmc-edx: Timeout exception, which would be caught, and we'd fail to collect metrics for that event. Think it's worth giving a partial event instead?

robrap: I was more concerned with the user seeing errors. I don't think this is likely, so may not be worth any effort. I didn't know if `p` (or `process`) would just be None.

robrap: In other words, don't sweat this.

robrap (approved):
Nice.

PR #748, opened 2021-05-14
feat: Implement consent check for metrics reporting

timmc-edx:
This adds a proper consent check. The whole thing is still gated on the feature flag for now, since we haven't yet user-tested the consent flow or gotten final approval. This does not yet advertise the new make targets.

Feature changes:

- Require new `"consent"` key in config to gate collecting and reporting
- Send event for opt-in (but not for opt-out, of course)
- New make targets `metrics-opt-in` and `metrics-opt-out`

Supporting changes:

- Lift config path to global variable (in script and tests)
- Extract config keys to constants
- Generalize to allow multiple Segment event types

ref ARCHBOM-1786

----

    - Already tested on: Linux
    - Testing instructions: With the feature flag in place and `DEVSTACK_METRICS_TESTING=debug` exported in your shell, try opting in and out several times in a row to see what happens. (e.g. in, in, out, out, in). Confirm that metrics are only sent on a positive opt-in or for instrumented commands (e.g. `make dev.up.redis`) while opted in.

### DIFF
--- a/scripts/send-metrics.py
+++ b/scripts/send-metrics.py
@@ -11,7 +11,7 @@ https://openedx.atlassian.net/wiki/spaces/AC/pages/2720432206/Devstack+Metrics
 
 If environment variable ``DEVSTACK_METRICS_TESTING`` is present and non-empty, then
 metrics will be annotated to indicate that the event occurred in the
-course of running tests so that it does not pollute our main data. (TODO)
+course of running tests so that it does not pollute our main data.
 (Extra debugging information will also be printed, including the sent
 event data.)
 """
@@ -36,6 +36,9 @@ test_mode = os.environ.get('DEVSTACK_METRICS_TESTING')
 # Provisioned as a separate source particular to devstack
 segment_write_key = "MUymmyrKDLk6JVwtkveX6OHVKMhpApou"
 
+config_dir = path.expanduser("~/.config/devstack")
+config_path = path.join(config_dir, "metrics.json")
+
 
 def base64str(s):
     """Encode a string in Base64, returning a string."""
@@ -50,14 +53,19 @@ def prep_for_send():
     and the returned config object will contain:
 
     - 'anonymous_user_id', a string
+    - 'consent', a dict containing:
+
+      - 'decision', a boolean indicating whether the user has consented
+        (always true when this function returns a config object)
+      - 'timestamp', the ISO-8601 date-time when the user consented (or
+        declined)
 
     Failure may take the form of an exception or returning None.
     """
-    config_path = path.expanduser("~/.config/devstack/metrics.json")
     with open(config_path, 'r') as f:
         config = json.loads(f.read())
 
-    # Currently serving in place of a consent check -- gate on
+    # Gate pretty much all functionality on the
     # presence of this manually configured setting so that people
     # developing this script can test it.
     #
@@ -77,16 +85,19 @@ def prep_for_send():
     if not config.get('collection_enabled'):
         return None
 
-    # Set user ID on first run
+    # Actual consent check.
+    if not config.get('consent', {}).get('decision'):
+        return None
+
+    # Opt-in process should have set an anonymous user ID, which is
+    # required for sending events.
     if 'anonymous_user_id' not in config:
-        config['anonymous_user_id'] = str(uuid.uuid4())
-        with open(config_path, 'w') as f:
-            f.write(json.dumps(config))
+        return None
 
     return config
 
 
-def send_metrics_to_segment(event_properties, config):
+def send_metrics_to_segment(event_type, event_properties, config):
     """
     Send collected metrics to Segment.
 
@@ -98,7 +109,7 @@ def send_metrics_to_segment(event_properties, config):
     event_properties['is_test'] = test_mode or 'no'
 
     event = {
-        'event': 'devstack.command.run',
+        'event': event_type,
         'userId': config['anonymous_user_id'],
         'properties': event_properties,
         'sentAt': datetime.now(timezone.utc).isoformat(),
@@ -211,7 +222,7 @@ def run_wrapped(make_target, config):
             'exit_status': exit_code,
             **read_git_state()
         }
-        send_metrics_to_segment(event_properties, config)
+        send_metrics_to_segment('devstack.command.run', event_properties, config)
     except:
         # We don't want to warn about transient Segment outages or
         # similar, but there might be a coding error in the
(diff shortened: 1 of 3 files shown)


scripts/send-metrics.py
```diff
@@ -35,6 +35,12 @@
 # Provisioned as a separate source particular to devstack
 segment_write_key = "MUymmyrKDLk6JVwtkveX6OHVKMhpApou"
```
robrap, line 37: Outside of this PR, but should this use all caps constant naming convention as well?

scripts/send-metrics.py
```diff
@@ -49,14 +55,14 @@ def prep_for_send():
     and the returned config object will contain:
 
     - 'anonymous_user_id', a string
+    - 'consent', a timestamp of when the user consented
```
robrap: I think 'consent_timestamp` would be a better name, if that is what this is. Or, can this also be False?

scripts/send-metrics.py
```diff
@@ -73,19 +79,23 @@ def prep_for_send():
     # .. toggle_creation_date: 2021-05-11
     # .. toggle_target_removal_date: 2021-07-01
     # .. toggle_tickets: https://openedx.atlassian.net/browse/ARCHBOM-1788 (edX internal)
-    if not config.get('collection_enabled'):
+    if not config.get(FEATURE_KEY):
```
robrap: Maybe `METRICS_FEATURE_KEY`?

scripts/send-metrics.py
```diff
@@ -73,19 +79,23 @@ def prep_for_send():
...
+        return None
+
+    # Actual consent check.
+    if not config.get(CONSENT_KEY):
```
robrap: Maybe `METRICS_CONSENT_KEY`?

scripts/send-metrics.py
```diff
@@ -95,10 +105,9 @@ def send_metrics_to_segment(event_properties, config):
     # event unless overridden by environment variable.
     event_properties = event_properties.copy()
     event_properties['is_test'] = test_mode or 'no'
-
     event = {
-        'event': 'devstack.command.run',
-        'userId': config['anonymous_user_id'],
+        'event': event_type,
```
robrap, line 112: Note: **if** we were to go with a default Segment->Snowflake integration down the line, it may create a table per event type. Is there a strong reason not to just reuse `devstack.command.run` for this?

scripts/send-metrics.py
```diff
@@ -189,5 +196,125 @@ def main(args):
...
+            "Thank you for contributing to devstack development! "
+            "You can opt out again at any time with `make metrics-opt-out` "
+            f"or by deleting the file at {config_path}."
+        )
+        # Send record of opt-in so we can tell whether people are
+        # opting in even if they're not running any of the
+        # instrumented commands.
+        send_metrics_to_segment('devstack.opt-in', {}, config)
+    else:
+        print("Not opted-in.")
```
robrap: Can we avoid the negative?
```suggestion
        print("Opted-out.")
```

scripts/send-metrics.py
```diff
@@ -189,5 +196,125 @@ def main(args):
...
+        "\n"
+        "This will report usage information to a team at edX (Arch-BOM) so that "
+        "devstack improvements can be planned and evaluated. The exact metrics "
+        "may change over time, but will include information such as the make "
+        "target, a timestamp, duration of command run, the version of devstack, "
+        "and an anonymous user ID. See "
+        "https://openedx.atlassian.net/wiki/spaces/AC/pages/2720432206/Devstack+Metrics "
+        "for more information. You can opt out again at any time.\n"
+        "\n"
+        "Type 'yes' or 'y' to opt in, or anything else to cancel."
```
robrap: [request] Let's add a link to this for Legal to review in our upcoming meeting.

scripts/send-metrics.py
```diff
@@ -189,5 +196,125 @@ def main(args):
...
+    except FileNotFoundError:
+        config = {}
+
+    config[CONSENT_KEY] = False
+    with open(config_path, 'w') as f:
+        f.write(json.dumps(config))
+
+    print(
+        "You have been opted out of reporting devstack command metrics to "
+        "the edX Arch-BOM team. This preference is stored in a config file "
```
robrap: Here and above, is it important to name Arch-BOM here? Just curious.

tests/metrics.py
```diff
@@ -45,13 +49,89 @@ def environment_as(config_data):
...
+    """
+    with open(config_path, 'r') as f:
+        config = json.loads(f.read())
+        if value == True:
+            # Should be a date at least
+            assert re.match(r'^[0-9]{4}-[0-9]{2}-[0-9]{2}', config.get('consent'))
+        elif value == False:
+            assert config.get('consent') == False
+        elif value == None:
+            assert 'consent'not in config
```
robrap:
```suggestion
            assert 'consent' not in config
```

robrap (commented):
Looking good. Let me know if you want me to do Mac testing now, or if I can wait until the dust settles on any changes.

scripts/send-metrics.py
timmc-edx: It can also be false, yeah.

scripts/send-metrics.py
timmc-edx: Feels different somehow. One's a variable, the other is a constant? Something like that.

robrap: I'm not clear why this isn't a constant.

scripts/send-metrics.py
timmc-edx: Done.

scripts/send-metrics.py
timmc-edx: They aren't actually opted out if they cancel out here -- it doesn't record anything.

scripts/send-metrics.py
timmc-edx: Mmm, not really. I want to have at least one reference to Arch-BOM so that people know who to talk to, but it could be in the script header or something. I'm removing these two and just leaving one in the feature-gating message ("Reach out to the Arch-BOM team at edX if you want to participate.")

scripts/send-metrics.py
robrap: I see.  Maybe `Cancelled opt-in`?  Or, leave what you have if you prefer.  Either way.

timmc-edx: I like that.

scripts/send-metrics.py
timmc-edx: Everything in here is about metrics. These are specific to the config file, though. `CONFIG_FEATURE_KEY`?

timmc-edx: (`config_key_feature_switch`?)

scripts/send-metrics.py
timmc-edx: Ah, interesting. Yeah, let's have this reuse `devstack.command.run` and just not include some of the keys (run time, exit status, etc.) Including the git commit ID could actually be pretty important, come to think of it.

timmc-edx: I've pushed up a change that implements this -- what do you think?

scripts/send-metrics.py
timmc-edx: Different kinds of constant, I suppose. I don't mean that this one changes over the course of the runtime, but that it's "data" as opposed to "fixed structural label". (To illuminate the difference: Inlining the former into the code would be reasonable, inlining the latter might be questionable!)

timmc-edx: But I'd also be fine just making all of these lowercase.

robrap: Nope. I'd prefer if this one is all-caps, but seeing as you disagree, I'd just leave things as-is.

scripts/send-metrics.py
robrap: Maybe:
```
a timestamp of when the user consented, or false if the user did not consent
```

scripts/send-metrics.py
robrap: The general term Nim is trying to use is toggle.  So, `FEATURE_TOGGLE_KEY` if you want to change the name at all.

scripts/send-metrics.py
timmc-edx: If consent is false, the config object isn't returned at all.

scripts/send-metrics.py
robrap: Wondering if we want to hard-code the event type with a comment explaining the implications of changing it. Or, you could leave this implementation, and just add a comment to the docstring, although it easily could be missed. Or, you could leave as-is. I'll leave this up to you.

robrap (approved):
Thanks @timmc-edx. I don't think there are any blockers left. Just some remaining minor proposals, and you can decide whether or not to adjust.

scripts/send-metrics.py
robrap: Ok. To my original comment, I think the name could be more specific. But, I won't let that be a blocker.

scripts/send-metrics.py
timmc-edx: I'll just make the comments in the tests (which verify the exact value here) a little stronger.

scripts/send-metrics.py
timmc-edx: Switched to lowercase with `config_key_` prefix.

scripts/send-metrics.py
timmc-edx: There's another way we could factor it out, with `"consent": {"decision": <bool>, "timestamp": <ts-str>}` which would allow us to do things in the future like "if they opted out, but it was a year ago, maybe ask again". What do you think?

robrap: In addition, I like it simply because it is more clear.

timmc-edx: Me too. :-) Union types make me uneasy.

scripts/send-metrics.py
timmc-edx: Wow, these get really verbose once I add a few more. I think I'll switch back to inlining.

scripts/send-metrics.py
```diff
@@ -77,7 +85,11 @@ def prep_for_send():
     if not config.get('collection_enabled'):
         return None
 
-    # Set user ID on first run
+    # Actual consent check.
+    if not config.get('consent', {}).get('decision'):
+        return None
+
+    # Set user ID if not present
```
timmc-edx: This is now also set in the opt-in code, which seems like the more appropriate place. I'm planning on removing it from here and just `return None` if it's not already set.

timmc-edx:
@robrap I've pushed up a few additional commits that I'll squash into the main commit. A couple are previously discussed, but two are new. Can you re-review those?

timmc-edx:
(Also a Mac test would be great.)

robrap (approved):

robrap:
@timmc-edx: exit_status is reported as null (missing?) instead of 0 for success of `dev.up.lms` for example. We could either update the dashboard to consider missing exit_status as success, or you could update the sent metrics for metrics-opt-in and metrics-opt-out. If you make the change, it could be on this or a follow-up PR.

robrap:
Note: I did a little of my testing without debug mode, but then I turned it on.

timmc-edx:
Hmm, that's troubling about the exit status! Will follow up in chat.

timmc-edx:
From discussion: metrics opt-in/out doesn't report an exit_status, which means they were showing up as failed commands due to the query condition `exit_status != 0`. The query has been changed to exclude anything with a missing exit status.

PR #749, opened 2021-05-17
feat: Allow metrics logging of make runs interrupted by Ctrl-C

timmc-edx:
This traps SIGINT in the metrics wrapper just before launching the
subprocess and untraps it right after (including if a signal is caught.)

This also moves the collection setup into a try/except block so that errors
during setup don't prevent the wrapped command from running.

----

    - Already tested on: Linux
    - Testing instructions: Try Ctrl-C during `make dev.pull`

### DIFF
--- a/scripts/send-metrics.py
+++ b/scripts/send-metrics.py
@@ -27,6 +27,7 @@ import uuid
 from datetime import datetime, timezone
 from http.client import RemoteDisconnected
 from os import path
+from signal import SIG_DFL, SIGINT, signal
 from urllib.error import URLError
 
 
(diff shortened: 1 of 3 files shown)


scripts/send-metrics.py
```diff
@@ -165,12 +166,33 @@ def run_wrapped(make_target, config):
...
+        start_time = datetime.now(timezone.utc)
+
+        # Catch SIGINT (but only once) so that we can report an event
+        # when the developer uses Ctrl-C to kill the make command
+        # (which would normally kill this process as well).  This
+        # handler just ignores the signal and then unregisters itself.
+        signal(SIGINT, lambda _signum, _frame: signal(SIGINT, SIG_DFL))
+    except:
+        do_collect = False
+        traceback.print_exc()
```
timmc-edx, line 181: **Author note:** Basically the same as the existing `except` block below. Manually tested by putting `1/0` at end of `try` block.

tests/metrics.py
```diff
@@ -3,6 +3,7 @@
 import pexpect
 import pytest
 from contextlib import contextmanager
+from pexpect import EOF
```
timmc-edx, line 6: **Author note:** Another PR does this as well and actually does this replacement of `pexpect.EOF` with `EOF`; this is just to align with that PR in anticipation of it merging.

scripts/send-metrics.py
```diff
@@ -165,12 +166,33 @@ def run_wrapped(make_target, config):
...
 
     completed_process = run_target(make_target)
 
     # Do as much as possible inside try blocks
     try:
+        # Skip metrics reporting if setup failed
+        if not do_collect:
+            return
+
+        signal(SIGINT, SIG_DFL)  # stop trapping SIGINT (if haven't already)
```
timmc-edx, line 195: **Author's note:** Manually tested by adding a "while True: pass" after the metrics-sending line to cause a hang, sending a C-c during `make dev.pull`, and sending a second C-c during the hang to confirm it exits rather than swallowing the signal. (It fails to send metrics in this case, which is fine.)

dianakhuang (approved):
New changes look good to me. Tested on Mac to verify that this works.

PR #750, opened 2021-05-17
fix: Move wildcard make targets to below things they could capture

timmc-edx:
On Mac, `dev.up.without-deps.lms` was invoking `dev.up.%` with the
non-existent "service" `without-deps.lms`, rather than the intended
`dev.up.without-deps.%` with the `lms` service. It worked on Linux,
so there must be some kind of implementation-dependent difference in how
wildcard targets are matched (first match vs. most specific).

Moving the wildcards to below anything that could be a competing match
should solve this.

Non-wildcard/non-wrapped versions of the same target are kept grouped
together.

----

    - Already tested on: Linux
    - Testing instructions: Run `make dev.up.without-deps.lms` and see if it doesn't error any more on Mac
    - [ ] Post-merge: Announce on #dev and request a few more Mac users to confirm this is working

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -198,6 +198,11 @@ dev.clone.ssh: ## Clone service repos using SSH method to the parent directory.
 # Developer interface: Docker image management.
 ########################################################################################
 
+dev.pull.without-deps: _expects-service-list.dev.pull.without-deps
+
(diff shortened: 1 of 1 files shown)


sarina (approved):
image

PR #754, opened 2021-05-24
fix: Update text of metrics opt-in.

dianakhuang:
We want to be more explicit about the data collected
for devstack metrics. Opt-in now lists all data collected.

https://openedx.atlassian.net/browse/ARCHBOM-1800

----

    - Already tested on: Mac
    - Testing instructions: run `make metrics-opt-in` and read through the updated opt-in text.

### DIFF
--- a/tests/metrics.py
+++ b/tests/metrics.py
@@ -84,7 +84,6 @@ def test_initial_opt_in_accept():
     with environment_as({'collection_enabled': True}):
         p = pexpect.spawn('make metrics-opt-in', timeout=10)
         p.expect_exact("Allow devstack to report anonymized usage metrics?")
-        p.expect("https://")  # gives a URL for more info
         p.expect("Type 'yes' or 'y'")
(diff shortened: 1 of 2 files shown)


timmc-edx (approved):

tests/metrics.py
```diff
@@ -100,7 +99,7 @@ def test_initial_opt_in_accept():
         # https://openedx.atlassian.net/wiki/spaces/AC/pages/2720432206/Devstack+Metrics
         #
         # Additional metrics require approval (as do changes to
-        # existing ones).
+        # existing ones). Update the list of metrics displayed during opt-in in send-metrics.py
```
timmc-edx: Maybe this phrasing instead, though?  I want to make sure it's clear that it's an *additional* requirement.
```suggestion
        # existing ones). Changes to metrics also require an update to the
        # list of metrics displayed during opt-in.
```

PR #758, opened 2021-05-27
feat: adding metrics collections to more targets

jinder1s:
particularly:
dev.clone...
dev.provision...
dev.pull...
- [ ] dev.up still needs to be done
This extends the work from https://github.com/edx/devstack/pull/732. 

[inform] We do not plan to continue using this method of metrics collect beyond these few essential make targets. Since the future of devstack is uncertain(i.e. potential replacement of devstack with tutor), it did not make sense to invest in finding a different method of metric collection.

----

    - Already tested on: jinder1s
    - Testing instructions: Run the modified targets and make sure they start correctly. We don't need to check if they end correctly cause making sure send_metrics scripts work correctly is done elsewhere.
- [ ] Made a plan to communicate any major developer interface changes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -292,8 +292,11 @@ dev.drop-db.%: ## Irreversably drop the contents of a MySQL database in each mys
 
 dev.up.attach: _expects-service.dev.up.attach
 
-dev.up.attach.%: ## Bring up a service and its dependencies + and attach to it.
+impl-dev.up.attach.%: ## Bring up a service and its dependencies + and attach to it.
 	docker-compose up $*
+
+dev.up.attach.%: ## Bring up a service and its dependencies + and attach to it.
+	@scripts/send-metrics.py wrap "dev.up.attach.$*"
 
 dev.up.shell: _expects-service.dev.up.shell
 
@@ -313,8 +316,11 @@ dev.up.with-watchers.%: ## Bring up services and their dependencies + asset watc
 
 dev.up.without-deps: _expects-service-list.dev.up.without-deps
 
-dev.up.without-deps.%: dev.check-memory ## Bring up services by themselves.
+impl-dev.up.without-deps.%: dev.check-memory ## Bring up services by themselves.
 	docker-compose up --d --no-deps $$(echo $* | tr + " ")
+
+dev.up.without-deps.%: ## Bring up services by themselves.
+	@scripts/send-metrics.py wrap "dev.up.without-deps.$*"
 
 dev.up.without-deps.shell: _expects-service.dev.up.without-deps.shell
 
--- a/Makefile
+++ b/Makefile
@@ -187,12 +187,17 @@ dev.checkout: ## Check out "openedx-release/$OPENEDX_RELEASE" in each repo if se
 
 dev.clone: dev.clone.ssh ## Clone service repos to the parent directory.
 
-dev.clone.https: ## Clone service repos using HTTPS method to the parent directory.
+impl-dev.clone.https: ## Clone service repos using HTTPS method to the parent directory.
 	./repo.sh clone
 
-dev.clone.ssh: ## Clone service repos using SSH method to the parent directory.
+dev.clone.https: ## Clone service repos using HTTPS method to the parent directory.
+	@scripts/send-metrics.py wrap "$@"
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -187,20 +187,28 @@ dev.checkout: ## Check out "openedx-release/$OPENEDX_RELEASE" in each repo if se
 
 dev.clone: dev.clone.ssh ## Clone service repos to the parent directory.
```
jinder1s, line 190: [reviewer question] I did not end up creating a metrics collecting wrapper for this because it itself is just a wrapper of dev.clone.ssh. Should I add a wrapper for this?

Makefile
```diff
@@ -220,18 +228,24 @@ dev.pull.%: ## Pull latest Docker images for services and their dependencies.
...
+impl-dev.provision: dev.check-memory ## Provision dev environment with default services, and then stop them.
 	# We provision all default services as well as 'e2e' (end-to-end tests).
 	# e2e is not part of `DEFAULT_SERVICES` because it isn't a service;
 	# it's just a way to tell ./provision.sh that the fake data for end-to-end
 	# tests should be prepared.
 	$(WINPTY) bash ./provision.sh $(DEFAULT_SERVICES)+e2e
 	make dev.stop
 
-dev.provision.%: ## Provision specified services.
+dev.provision:
```
jinder1s: this needs a comment

Makefile
```diff
@@ -187,20 +187,28 @@ dev.checkout: ## Check out "openedx-release/$OPENEDX_RELEASE" in each repo if se
...
+	@scripts/send-metrics.py wrap "$@"
 
 ########################################################################################
 # Developer interface: Docker image management.
 ########################################################################################
 
 dev.pull.without-deps: _expects-service-list.dev.pull.without-deps
 
-dev.pull.without-deps.%: ## Pull latest Docker images for specific services.
+dev.pull.without-deps.%:
```
jinder1s: this needs a comment

Makefile
```diff
@@ -187,20 +187,28 @@ dev.checkout: ## Check out "openedx-release/$OPENEDX_RELEASE" in each repo if se
...
 	./repo.sh clone
 
-dev.clone.ssh: ## Clone service repos using SSH method to the parent directory.
+dev.clone.https: ## Clone service repos using HTTPS method to the parent directory.
+	@scripts/send-metrics.py wrap "$@"
+
+impl-dev.clone.ssh: ## Clone service repos using SSH method to the parent directory.
 	./repo.sh clone_ssh
 
+dev.clone.ssh:
```
jinder1s: this needs a comment

Makefile
timmc-edx: `dev.pull.%` is mentioned in the ticket, although this would be interesting to capture as well.

Makefile
```diff
@@ -220,18 +228,24 @@ dev.pull.%: ## Pull latest Docker images for services and their dependencies.
 # Developer interface: Database management.
 ########################################################################################
 
-dev.provision: dev.check-memory ## Provision dev environment with default services, and then stop them.
+impl-dev.provision: dev.check-memory ## Provision dev environment with default services, and then stop them.
```
timmc-edx: We probably want to move the dependency down into the body as `make dev.check-memory` so that the execution time is included.

Makefile
```diff
@@ -187,20 +187,28 @@ dev.checkout: ## Check out "openedx-release/$OPENEDX_RELEASE" in each repo if se
 
 dev.clone: dev.clone.ssh ## Clone service repos to the parent directory.
 
-dev.clone.https: ## Clone service repos using HTTPS method to the parent directory.
+impl-dev.clone.https: ## Clone service repos using HTTPS method to the parent directory.
```
timmc-edx, line 192: All of the new targets that don't have wildcards should be added to the `.PHONY` block, alphabetically.

Makefile
timmc-edx: We could add one, if we were interested in whether people run the top-level one or the ssh/https ones directly, but I think there's no need -- we can add this later if there's some reason to measure that specifically.

timmc-edx (approved):

PR #759, opened 2021-05-27
feat: Print invitation to opt-in to metrics

timmc-edx:
Invitation is only printed for people who have the feature toggle enabled (for now) and who have neither consented nor declined.

Major supporting changes:

- Change `prep_for_send` into `check_for_consent`, since there is now an in-between state (feature toggled enabled, but decision not made) where we want the wrapper to behave differently. There are three possible return states now, factored across two booleans.

Minor changes:

- Move documentation of config file format to module docstring
- Tightens values for `collection_enabled` -- has to be True, not just truthy
- Treat falsey user ID (e.g. empty string) as missing
- Bugfix: Check for missing user ID in opt-in as well (to match the other check)

Also includes some preliminary improvements to tests as a foundational commit.

----

### DIFF
--- a/scripts/send-metrics.py
+++ b/scripts/send-metrics.py
@@ -14,6 +14,16 @@ metrics will be annotated to indicate that the event occurred in the
 course of running tests so that it does not pollute our main data.
 (Extra debugging information will also be printed, including the sent
 event data.)
+
+Config file schema:
+
+- 'collection_enabled' (boolean) feature toggle gating several aspects
+  of metrics collection (see later description)
+- 'anonymous_user_id' (string) high-entropy, non-identifying unique user ID
+- 'consent' (dict) present when user has either consented or declined
+  metrics collection
+    - 'decision' (boolean) indicates whether the user has consented or declined
+    - 'timestamp' (string) ISO-8601 date-time when the user consented or declined
 """
 
 import base64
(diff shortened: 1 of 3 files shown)


scripts/send-metrics.py
```diff
@@ -247,16 +271,23 @@ def do_wrap(make_target):
...
-        run_wrapped(make_target, consented_config)
+    if consent_state.get('ok_to_collect'):
+        run_wrapped(make_target, consent_state.get('config'))
     else:
         run_target(make_target)
+        if consent_state.get('print_invitation'):
+            print(
+                "Would you like to assist devstack development by sending "
+                "anonymous usage metrics to edX? Run `make metrics-opt-in` "
+                "to learn more!",
```
timmc-edx, line 287: Author note: I can't decide if this looks weird or not. If I hadn't implemented this target, I might think running this would opt me in right away, rather than giving me info and asking for consent.

...but I'm not really sure how to rephrase this and _also_ keep it short, so I'm just putting it up for review now.

scripts/send-metrics.py
```diff
@@ -14,6 +14,17 @@
...
 event data.)
+
+Config file schema:
+
+- 'collection_enabled' (boolean) feature toggle gating several aspects
+  of metrics collection (see later description)
+- 'anonymous_user_id' (string) high-entropy, non-identifying unique user ID
+- 'consent' (dict) present when user has either consented or declined
+  metrics collection
+
```
dianakhuang: Nit: this space is a little confusing to me.

scripts/send-metrics.py
```diff
@@ -45,22 +56,16 @@ def base64str(s):
...
-    - 'anonymous_user_id', a string
-    - 'consent', a dict containing:
-
-      - 'decision', a boolean indicating whether the user has consented
-        (always true when this function returns a config object)
-      - 'timestamp', the ISO-8601 date-time when the user consented (or
-        declined)
-
-    Failure may take the form of an exception or returning None.
+    May throw if config file cannot be read.
```
dianakhuang: What sort of exception does this throw?

scripts/send-metrics.py
dianakhuang: We discussed this in parking lot and decided that this is fine for now and we can do some user testing with it once we have other engineers on board with it.

dianakhuang (approved):
Some minor comments

scripts/send-metrics.py
timmc-edx: I think that's the rst creeping in. :-P Maybe 4 space indentation, no line break?

scripts/send-metrics.py
timmc-edx: I/O exception, JSON parsing, file not found, that sort of thing. Maybe also KeyError. Basically, "don't call this from the critical path without a try/except wrapper".

dianakhuang: Nit: I would phrase it as `May throw an exception if config file cannot be read.` so that it's clear.

timmc-edx: Thanks, I like that..

PR #761, opened 2021-05-28
test: Move CLI expect-tests to separate workflow; test Mac support there

timmc-edx:
A few test changes were required to account for slight differences in environments:
    
- Fix Ctrl-c test: Did not account for nondeterministic output ordering
- Fix "make interrupt" regex (not sure why this was working on Linux before!)

----

    - Will announce on edx-internal #dev and openedx #general -- not that there are any interface changes, but this Mac/Docker/GH Actions integration was non-trivial and it would be good to spread the solution around a bit.

### DIFF
--- a/.github/workflows/cli-tests.yml
+++ b/.github/workflows/cli-tests.yml
@@ -1,0 +1,85 @@
+name: CLI tests
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches:
+      - '**'
+
+jobs:
+
+  run_ci:
+    runs-on: ${{ matrix.os.image }}
+    env:
+      DEVSTACK_WORKSPACE: /tmp
+      SHALLOW_CLONE: 1
+      # Don't report metrics as real usage
+      DEVSTACK_METRICS_TESTING: ci
+    strategy:
+      matrix:
+        os:
+          - name: linux
+            image: ubuntu-20.04 # Focal Fossa
+          - name: mac
+            image: macos-10.15 # Catalina
+        python-version:
+          - '3.8'
+      fail-fast: false
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: setup python
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Docker installation - Linux
+        if: ${{ matrix.os.name == 'linux' }}
+        run: |
+          docker version
+          sudo apt-get update
+          sudo apt install apt-transport-https ca-certificates curl software-properties-common
+          curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+          sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu focal test"
+          sudo apt update
+          sudo apt install docker-ce
+          docker version
+          docker-compose --version
+
+      # Cache boot2docker for speedup and to avoid ratelimiting
+      - name: Docker cache - Mac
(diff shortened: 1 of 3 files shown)


tests/metrics.py
```diff
@@ -279,16 +279,18 @@ def test_handle_ctrl_c():
         # Make sure wrapped command has started before we interrupt,
         # otherwise signal handler won't even have been registered
         # yet.
-        p.expect(b'Are you sure you want to run this command')
+        p.expect(r'Are you sure you want to run this command')
```
timmc-edx, line 282: Fixing an unrelated typo here. (No functional change, and could be a regular string, but using "raw strings" is a good reminder that these are in fact regexes.)

.github/workflows/cli-tests.yml
```diff
@@ -0,0 +1,85 @@
...
+          # correct in the general case, but these Docker environment
+          # variables shouldn't contain newlines or escape sequences.
+          # They'll look like this:
+          #   DOCKER_TLS_VERIFY="1"
+          #   DOCKER_HOST="tcp://192.168.99.100:2376"
+          #   DOCKER_CERT_PATH="/Users/runner/.docker/machine/machines/default"
+          #   DOCKER_MACHINE_NAME="default"
+          #
+          # Docs on GITHUB_ENV:
+          # https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable
```
timmc-edx: I'm inclined to remove these lines from the comment.

.github/workflows/cli-tests.yml
```diff
@@ -0,0 +1,85 @@
+name: CLI tests
```
robrap, line 4: I feel like I'd like a file-level comment, like a module docstring, to state the purpose of the workflow. Something to add a bit more color to what "cli-tests" is vs "ci".

.github/workflows/cli-tests.yml
robrap: I think this comment is fine. It made it immediately clear to me that $GITHUB_ENV was a standard thing, and not something you are setting up as part of the workflows.

robrap (approved):
I think this is all good, but not certain and would trust better that it passes, and that it doesn't pass when broken (which I presume you have tested).

.github/workflows/cli-tests.yml
timmc-edx: Yeah, that sounds good.

.github/workflows/cli-tests.yml
timmc-edx: (I had highlighted lines 71 through 78 to make this comment, but it looks like the resulting lines-scope indicator is subtle. Oh well.)

Makes sense about GITHUB_ENV.

How about the Docker env vars? If I keep them I should either remove the quotes or add `export ` in front of each one in order to make it look like either the input or the output format...

robrap: I liked the large comment. Maybe include fewer variables (like 2) if you think it is getting too long, but I'd just make it a Before and After to make it clear on both sides.

timmc-edx:
Incidentally, this ends up using boot2docker, which the README specifically says not to use. :man_facepalming: It seems to work for some Docker-requiring actions But I'm not sure how to install Mac Docker for Desktop in a Github Action.

robrap:
Still approved as-is. But, does your comment https://github.com/edx/devstack/pull/761#issuecomment-853889581 warrant a comment in the code somewhere?

timmc-edx:
Yeah, I'll add a comment about that and we can explore other options later if we want.

PR #767, opened 2021-06-03
fix: Error out of provisioning when any part errors (fix silent CI fail)

timmc-edx:
CI tests were silently failing during provisioning since the provisioning scripts did not set the `-e` option in bash. This commit changes the scripts to include the recommended `set -eu -o pipefail` stanza as well as include a more conservative `-e` on any docker-exec bash calls.

In places where existing errors were uncovered, I suppressed errors just for the offending code with a `+e`/`-e` pair and left a comment starting with `# FIXME[bash-e]`. Owning teams can prioritize and fix these at their own pace.

This may cause some false positive errors (there are commands which use a non-zero exit code to signal things other than an error) but they should be individually reverted or fixed as they are discovered.

Also:

- Extract ANSI color code variables to a shared file and use them. Most of the provisioning scripts tried to use the color codes, but they weren't defined. I discovered this when I tried to add `set -eu` to the scripts.
- Remove the `set -e` from the CI Github Action, since it is almost certainly a no-op.
- Add `set -x` to all provisioning scripts for easier debugging (some already had it)
- Ensure provisioning scripts have bash shebang
- Tweak function in `programs/provision.sh` to accept variable list of arguments (and stop defaulting one of the args)

----

Communication plan:

After merge I will ping the owning team for each added FIXME to let them know about the issue and how to go about testing its removal. Where they are specific to one service, I'll let the team know they can fix it at their leisure; where they're in shared files (i.e. `provision-ida-user.sh`) I will indicate that this blocks other teams from seeing errors and ask them to make it higher priority to get their provisioning fully working.

----

    - Already tested on: Linux
    - Testing instructions: Re-provision devstack: `make dev.destroy && make dev.reset-repos && make dev.pull && make dev.provision && make dev.up` then wait about 5 minutes and run `make dev.check`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,9 +53,6 @@ jobs:
       - name: pull
         run:  make dev.pull.${{matrix.services}}
 
-      - name: set exit option
-        run:  set -e
-
       - name: provision
         run:  make dev.provision.${{matrix.services}}
 
--- a/check.sh
+++ b/check.sh
@@ -16,9 +16,7 @@
 # Note that passing in a non-existent service will not fail if there are
 # other successful checks.
 
-set -e
-set -o pipefail
-set -u
+set -eu -o pipefail
 
 # Grab all arguments into one string, replacing plus signs with spaces.
 # Pad on either side with spaces so that the regex in `should_check` works correctly.
--- a/destroy.sh
+++ b/destroy.sh
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
-
-set -e
+set -eu -o pipefail
 
 read -p "This will delete all data in your devstack. Would you like to proceed? [y/n] " -r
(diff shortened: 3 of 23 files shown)


timmc-edx:
The ecommerce CI build failed in the most recent run, but ecommerce devstack tests have been flaky for a long time. I was able to get provision/up/check to pass locally and in a previous build.

jinder1s:
Coding wise, this PR looks good, but I wasn't able to test this correctly on a mac due to what seemed like unrelated issues. My mysql and mysql5.7 continers keep exiting and I haven't been able to figure it out. This also happens in master branch.


And since i"m taking PTO tomorrow and friday, I won't be able to continue reviewing this.

timmc-edx:
Thanks to @edx-abolger for running the manual tests on MacOS!

dianakhuang (approved):

PR #768, opened 2021-06-03
fix: Convey wrapped process's exit code in metrics wrapper

timmc-edx:
----

    - Already tested on: Linux
    - Testing instructions: Run `make dev.pull.xxx; echo $?` with and without opt-in and confirm that `2` is printed.
- [x] Made a plan to communicate any major developer interface changes (N/A)

### DIFF
--- a/scripts/send-metrics.py
+++ b/scripts/send-metrics.py
@@ -199,6 +199,8 @@ def read_git_state():
 def run_wrapped(make_target, config):
     """
     Runs specified make shell target and collects performance data.
+
+    Return exit code of process (negative for signals).
     """
     # Do as much as possible inside try blocks
     do_collect = True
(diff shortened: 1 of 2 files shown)


scripts/send-metrics.py
```diff
@@ -435,7 +451,8 @@ def main(args):
         if len(action_args) != 1:
             print("send-metrics wrap takes one argument", file=sys.stderr)
             sys.exit(1)
-        do_wrap(action_args[0])
+        conveyed_exit_code = do_wrap(action_args[0])
```
robrap, line 454: Nit: not sure if I'd prefer simple `exit_code`, but not a big deal either way.

scripts/send-metrics.py
```diff
@@ -288,6 +298,12 @@ def do_wrap(make_target):
                 file=sys.stderr
             )
 
+    if subprocess_exit_code < 0:
+        # Translate to shell convention.
+        return 128 + -subprocess_exit_code
```
robrap, line 303: I'm not clear if this is unit tested below, or if it can be.

robrap (approved):

scripts/send-metrics.py
timmc-edx: I originally had it all named `exit_code` but it ended up being confusing, and I wanted to have the boundary between the two kinds of exit codes be clear. I think this makes sense for symmetry.

scripts/send-metrics.py
timmc-edx: I had tried earlier and run into an issue (`make` turns all errors into `2`) but I could try calling the script directly.

timmc-edx: The confusing thing is that the one signal it handles is SIGINT, which is value 2, but that's also the exit code that make uses. With SIGINT it will exit with code 130, but then pexpect translates that *back* into signalstatus = 2.

timmc-edx: Oh, no, I'm wrong! pexpect actually shows exit code 130. There's something I'm missing about how child processes work in Linux and how exit codes and signals interact (or don't). Anyway, test added.

PR #769, opened 2021-06-04
 test: Fix a flaky metrics test due to signal/exit race

timmc-edx:
Also add some comments based on what I learned while fixing this.

----

### DIFF
--- a/scripts/send-metrics.py
+++ b/scripts/send-metrics.py
@@ -211,6 +211,12 @@ def run_wrapped(make_target, config):
         # when the developer uses Ctrl-C to kill the make command
         # (which would normally kill this process as well).  This
         # handler just ignores the signal and then unregisters itself.
+        #
+        # There is no guarantee of whether this signal will be caught
+        # first or the child process will exit first. This could
+        # theoretically result in unregistering the signal handler
+        # before the signal has occurred, resulting in failure to
+        # report the event, but there's no way to prevent this.
         signal(SIGINT, lambda _signum, _frame: signal(SIGINT, SIG_DFL))
     except:
         do_collect = False
--- a/tests/README.rst
+++ b/tests/README.rst
@@ -6,3 +6,10 @@ Expect); if you're editing or creating tests it is highly recommended
 you read up on the gotchas in here:
 
 https://pexpect.readthedocs.io/en/stable/overview.html
+
+Debugging tips
+--------------
+
+If an expectation fails (or an unexpected timeout or EOF occurs) then pexpect will throw an exception which contains a printout of the internal state of the pexpect instance. This includes the ``buffer``, ``before``, and ``after`` buffers, which are essential for seeing what the child process's output was.
+
+However, pexpect truncates two of these to 100 characters by default. To see more, set the undocumented ``str_last_chars`` attribute on the pexpect object to something larger, or to 0 for the full output.
--- a/tests/metrics.py
+++ b/tests/metrics.py
@@ -323,11 +323,16 @@ def test_signal_conversion():
         p.sendintr()  # send Ctrl-C to process group
         # Confirm that the process is actually catching the signal, as
         # proven by it printing some things before ending.
-        p.expect(r'Send metrics info:.*"exit_status": ?-2')
+        p.expect(r'Send metrics info:')
         p.expect(EOF)
 
         # This time we're calling the script directly, so we see the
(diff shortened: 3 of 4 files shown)


scripts/send-metrics.py
```diff
@@ -211,6 +211,12 @@ def run_wrapped(make_target, config):
         # when the developer uses Ctrl-C to kill the make command
         # (which would normally kill this process as well).  This
         # handler just ignores the signal and then unregisters itself.
+        #
+        # There is no guarantee of wther this signal will be caught
```
robrap:
```suggestion
        # There is no guarantee of whether this signal will be caught
```

timmc-edx:
Notes on flakiness:

- This test had never failed locally (Ubuntu 20.04)
- It only ever failed on the ubuntu-20.04 Github Action, and every single time.
- It never failed on the mac GH Action runs
- The failures stopped on Github when I prevented the full-provisioning CI tests from running

scripts/send-metrics.py
```diff
@@ -211,6 +211,12 @@ def run_wrapped(make_target, config):
...
+        # first or the child process will exit first. This could
+        # theoretically result in unregistering the signal handler
+        # before the signal has occurred, resulting in failure to
+        # report the event, but there's no way to prevent this.
```
robrap, line 219: There is no way to prevent this, or it isn't worth implementing a better workaround? I'm literally asking, because I don't actually want to think very deeply about this. :)

tests/README.rst
```diff
@@ -6,3 +6,10 @@ Expect); if you're editing or creating tests it is highly recommended
 you read up on the gotchas in here:
 
 https://pexpect.readthedocs.io/en/stable/overview.html
+
+Debugging tips
+--------------
+
+If an expectation fails (or an unexpected timeout or EOF occurs) then pexpect will throw an exception which contains a printout of the internal state of the pexpect instance. This includes the ``buffer``, ``before``, and ``after`` buffers, which are essential for seeing what the child process's output was.
+
+However, these outputs are truncated to 100 characters. To see the more, set the undocumented ``str_last_chars`` attribute on the pexpect object to something larger, or to 0 for the full output.
```
robrap:
```suggestion
However, these outputs are truncated to 100 characters. To see more, set the undocumented ``str_last_chars`` attribute on the pexpect object to something larger, or to 0 for the full output.
```

robrap: Also, I'm curious why you truncate to begin with. Does it make it more readable? If so, maybe something like "...are truncated to 100 characters to make the output more readable.", or something that explains why all the complexity.

tests/metrics.py
```diff
@@ -320,14 +320,19 @@ def test_signal_conversion():
...
-        p.expect(r'Send metrics info:.*"exit_status": ?-2')
+        p.expect(r'Send metrics info:')
         p.expect(EOF)
 
         # This time we're calling the script directly, so we see the
         # script exiting with code 130 (128 + SIGINT).
+        #
+        # ...or, depending on timing and/or operating system, the make
+        # process dies *first* with its generic exit code 2 and the
+        # wrapper exits before it even receives the signal. So, we
```
robrap:
```suggestion
        # process may die *first* with its generic exit code 2 and the
        # wrapper would exit before it even receives the signal. So, we
```

robrap (commented):
Minor comments and suggestions.

scripts/send-metrics.py
timmc-edx: No way to prevent it, as far as I can tell. I think the SIGINT can be delivered in any order to the wrapper, its child process (Make), and whatever process Make kicks off (e.g. docker-compose). If docker exits first, then Make can exit normally with exit code 2, and that's what the wrapper sees. Or maybe the wrapper gets the signal first, but by the time it handles it, Make has exited. Or Make is killed by the signal, so the wrapper sees that...

(It's possible there's some ordering guarantee I'm not aware of, but I also don't want to do that research for this. :-P)

tests/README.rst
timmc-edx: That truncation is [part of the pexpect library](https://github.com/pexpect/pexpect/blob/5eed1a31a2853a09b7367c59fbd1a4a0b53341df/pexpect/pty_spawn.py#L194). I'll clarify that.

robrap (approved):
Great. Thanks!

PR #774, opened 2021-06-10
feat: Allow metrics opt in without checking feature flag

timmc-edx:
This changes the effect of the feature flag to only gate the invitation.

There is a prefatory commit to add some tests which then change on the feature commit.

----

### DIFF
--- a/tests/metrics.py
+++ b/tests/metrics.py
@@ -13,8 +13,9 @@ from pexpect import EOF
 
 #### Utilities
 
-# A substring to identify whether an invitation has been made
+## Substrings to use as probes:
 invitation = 'Would you like to assist devstack development'
+consent_question = "Allow devstack to report anonymized usage metrics?"
 
 config_dir = os.path.expanduser('~/.config/devstack')
 config_path = os.path.join(config_dir, 'metrics.json')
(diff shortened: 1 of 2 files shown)


tests/metrics.py
```diff
@@ -13,8 +13,9 @@
 
 #### Utilities
 
-# A substring to identify whether an invitation has been made
+## Substrings to use as probes:
 invitation = 'Would you like to assist devstack development'
```
robrap, line 17: Should this also have a `?`?

tests/metrics.py
```diff
@@ -198,35 +206,52 @@ def test_feature_flag_missing():
...
         p = pexpect.spawn('make dev.up.redis', timeout=60)
         p.expect(EOF)
         assert 'Send metrics info:' not in p.before.decode()
-        assert invitation in p.before.decode()
+        assert invitation not in p.before.decode()
+
+
+def test_feature_flag_does_not_gate_collection():
+    """
+    Test that feature flag is required for collection even with consent.
```
robrap: Function name and docstring seem to disagree, right?

tests/metrics.py
timmc-edx: It doesn't need to be a full statement -- just a probe to unambiguously match part of the invitation.

tests/metrics.py
timmc-edx: Thanks! Fixed now. That test is also kind of redundant with a later one, but that just means it can be removed entirely when the feature toggle is removed.

robrap (approved):

PR #781, opened 2021-06-15
feat: print out untracked files

jinder1s:
Print out untracked files when running "./repo.sh reset". Reset only
modifies repositories in a way that allows changes to be recovered. As
such, reset does to automatically delete untracked files.
Without this new echo, it was easy to not know that one of the
repositories is in a modified state.

----

### DIFF
--- a/repo.sh
+++ b/repo.sh
@@ -183,7 +183,7 @@ reset ()
             # Try to switch branch and pull, but fail if there are uncommitted changes.
             if (cd "$name"; git checkout -q master && git pull -q --ff-only);
             then
-                # Echo untracked files to simplify debugging and make it easier to see tha reseting does not remove everything
+                # Echo untracked files to simplify debugging and make it easier to see that resetting does not remove everything
                 untracked_files="$(cd ${name} && git ls-files --others --exclude-standard)"
                 if [[ $untracked_files ]];
                 then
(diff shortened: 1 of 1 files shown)


repo.sh
```diff
@@ -181,11 +181,20 @@ reset ()
 
         if [ -d "$name" ]; then
             # Try to switch branch and pull, but fail if there are uncommitted changes.
-            (cd "$name"; git checkout -q master && git pull -q --ff-only) || {
+            if (cd "$name"; git checkout -q master && git pull -q --ff-only);
+            then
+                # Echo untracked files to simplify debugging and make it easier to see tha reseting does not remove everything
```
timmc-edx:
```suggestion
                # Echo untracked files to simplify debugging and make it easier to see that resetting does not remove everything
```
nit: typos

repo.sh
```diff
@@ -181,11 +181,20 @@ reset ()
...
+                untracked_files="$(cd ${name} && git ls-files --others --exclude-standard)"
+                if [[ $untracked_files ]];
```
timmc-edx, line 188: Oh huh, I didn't know you could do this! I thought you had to use -n or -z.

timmc-edx (approved):

PR #782, opened 2021-06-16
Add elasticsearch 7.10 container.

adzuci:
This is the first step towards testing devstack services on Elastic Search 7.10 which is the latest version that AWS supports.

----

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -79,6 +79,25 @@ services:
       - bootstrap.memory_lock=true
       - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
 
+  # This is meant to be used to test ES upgrades.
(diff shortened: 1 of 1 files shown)


dianakhuang (approved):
I would say get rid of the original `elasticsearch` container, but I think technically the analytics containers still use it.

adzuci:
> I would say get rid of the original `elasticsearch` container, but I think technically the analytics containers still use it.

I can follow up with @schenedx to see if he knows whether or not the 1.5 containers are still needed, but I think that for now we are using all three.

schenedx:
@adzuci We have a ticket in our current sprint to upgrade the ES version from 1.9 to the new version. See https://openedx.atlassian.net/browse/MST-769. CC @Zacharis278

PR #783, opened 2021-06-17
invert dependencies so that backends now depend on frontends (plus some cleanup)

kdmccormick:
## Description

The intent of this PR is to resolve an immediate developer experience issue: _since the Learning MFE is now enabled by default, the command `make dev.up.lms` on its own is no longer sufficient to run the default Open edX courseware experience_.

We hope that the fix for this issue will help us resolve a deeper, upcoming developer experience issue: _soon [^1], several micro-frontends will need to be started alongside backend services in order to provide a working Open edX experience out-of-the-box_.

Changes are best reviewed by-commit.

1. The first commit is some doc fixes related to devserver restarting.
2. The next commit is a superficial-yet-breaking interface change in order to make frontend app names consistent in Devstack. We are doing this clean-up opportunistically, since we'll need to announce workflow-breaking changes from this PR anyway.
3. The next commit is a non-user-facing style cleanup (some YAML alphabetization).
4. The next commit is a non-user-facing refactor (pulling XQueue into the main docker-compose.yml file).
5. The penultimate commit is a major interface change, along with an ADR. In short, backend services will be defined as depending on frontend services, such that `make dev.up.<backend-service>` automatically starts all required frontends service.
6. The final commit makes it so local NPM packages can be used with devstack microfrontends. There is an included ADR, and I am [making frontend-build doc updates here](https://github.com/edx/frontend-build/pull/193).

I will communicate all interface changes via email and Slack.

[^1]: When? Not sure; it's up to individual squads that own MFEs. ~We do know that Engage squad has work in-flight to make the Course Home features run in the Learning MFE by default.~ (update: they just did!) Also, several MFEs run as the default in Lilac, so it would make sense if those were made to run in Devstack by default as well.

## Supporting information

[TNL-8407](https://openedx.atlassian.net/browse/TNL-8407)

Further details are in the included ADRs (0004 and 0005).

Related PR to update frontend-build docs: https://github.com/edx/frontend-build/pull/193

## Testing instructions

Suggested smoke test for service dependencies:

* Ensure that `make dev.up.lms` brings up at least LMS, frontend-app-gradebook and frontend-app-learning. Ensure that those services are usable.
* Ensure that `make dev.attach.frontend-app-gradebook` attaches to the Gradebook MFE container.
* Ensure that `make dev.up.xqueue` brings up XQueue and MySQL 5.7.
* Ensure that `make dev.down` brings down all services.

Suggested smoke test for local package mounting:

* Start LMS and fa-learning with `make lms-up`.
* Load the demo course, confirming that it loads in the Learning MFE using the default (bootstrap blue) Open edX theme.
* Clone https://github.com/edx/brand-edx.org into the `src/` directory within your devstack workspace.
* Run `npm install && npm build` in that repo.
* Follow [these update docs](https://github.com/edx/frontend-build/blob/kdmccormick/local-package-docs/README.rst#local-module-configuration-for-webpack), copying `module.config.js` into frontend-app-learning.
* In `module.config.js`, set `@edx/brand` to the directory `../src/brand-edx.org`. Comment out all other local modules.
* Shell into frontend-app-learning (`make frontend-app-learning-shell`) and run `npm install`. Exit shell.
* Restart frontend-app-learning.
* Load the demo course again, confirming that it shows the edX.org elm-green theme.


docker-compose-host-nfs.yml
```diff
@@ -45,22 +45,22 @@ services:
   registrar-worker:
     volumes:
       - ${DEVSTACK_WORKSPACE}/registrar:/edx/app/registrar/registrar
-  gradebook:
+  frontend-app-gradebook:
```
kdmccormick, line 52: These changes will require a simple Devstack restart (`make dev.down`, `make dev.up...`). I will note this in my email.

docker-compose-host.yml
```diff
@@ -50,16 +50,14 @@ services:
   registrar-worker:
     volumes:
       - ${DEVSTACK_WORKSPACE}/registrar:/edx/app/registrar/registrar
-  gradebook:
+  frontend-app-gradebook:
     volumes:
-      - ${DEVSTACK_WORKSPACE}/frontend-app-gradebook:/edx/app/gradebook:cached
-      - gradebook_node_modules:/edx/app/gradebook/node_modules
-      - gradebook_tox:/edx/app/gradebook/.tox
```
kdmccormick, line 57: Frontends don't use tox; the `gradebook_tox` line must have been copypasta. Removing.

docker-compose-sync.yml
```diff
@@ -59,11 +59,11 @@ volumes:
     external: true
   registrar-sync:
     external: true
-  gradebook-sync:
```
kdmccormick, line 62: Yes, several frontends are missing from this file. I'm not going tackle that in this PR. I also doubt that anyone uses the Docker Sync stuff.

docs/decisions/0003-backends-depend-on-frontends.rst
```diff
@@ -0,0 +1,60 @@
...
+
+
+Docker-compose service dependencies
+***********************************
+
+Devstack uses docker-compose to orchestrate containers by defining services in ``docker-compose.yml``. Note that "services" here encompasses backends, frontends, and generic resources like MySQL.
+
+Each service definition may indicate a list of depentent services using the ``depends_on`` key. Dependencies are transitive, and may not be cyclical. When a developer runs ``make dev.up.<service>``, docker-compose is invoked in order to start both the service as well as its dependencies. For example, LMS depends on Mongo and Discovery, among other services. So, running ``make dev.up.lms`` will start not just LMS, but also Mongo, Discovery, all of Discovery's dependencies, and so on.
+
+Currently, micro-frontend services (those prefixed with ``frontend-app-``) are defined to depend on backends, but not vice versa. So, starting frontend-app-learning will automatically start LMS, but start LMS will not automatically start frontend-app-learning. This makes sense under logic that "frontends depend on the APIs of backends in order to function".
```
kdmccormick: s/but start LMS/but starting LMS

docs/decisions/0003-backends-depend-on-frontends.rst
```diff
@@ -0,0 +1,60 @@
...
+
+However, it can be argued that the opposite dependency relationship also makes sense. That is, one may assert that backends should depend on frontends in order to expose their APIs in a usable way. One could further assert that frontends shouldn't have hard dependencies on backend APIs, and should instead gracefully degrade when some or all of its APIs are unavailble.
+
+
+Decision
+========
+
+Whichever dependency direction (frontends depend on backends, or vice versa) is more logically sound, we conclude that, for the puposes of Devstack, *asserting that backends depend on frontends is more useful to developers*. Specifically, it is beneficial to current and future developer workflows if ``make dev.up.lms`` automatically starts and learning-related frontends, ``make dev.up.studio`` automatically starts all authoring-related frontends, ``make dev.up.ecommerce`` starts all purchasing-related frontends, and so on.
+
+A necessary corollary to this decision is that *all micro-frontends should be available in Devstack*, which is not currently the case. Core frontends such as frontend-app-account and frontend-app-authn missing as of this ADR.
```
kdmccormick: s/as of this/as of the writing of this

docs/decisions/0003-backends-depend-on-frontends.rst
```diff
@@ -0,0 +1,60 @@
...
+
+Consequences
+============
+
+* ``docker-compose.yml`` will be updated to reflect that backends services depend on frontend-app services, not the other way around.
+* An email and Slack message will be sent out to explain this change and how we anticipate that it will impact developer workflows.
+* A Community Engineering ticket will be sumitted to address an issue ``webpack.config.js`` handling in Devstack, which is currently causing developers to run MFEs outside of Devstack.
```
kdmccormick: I don't really understand the `webpack.config.js` issue (is that even the right file name?), but I remember it being mentioned as a problem. Feel free to recommend rewording.

@davidjoy is this something CE could ticket as part of frontend ownership? I could link the ticket here.

docs/decisions/0003-backends-depend-on-frontends.rst
```diff
@@ -0,0 +1,60 @@
...
+* Arch-BOM will circulate a recommendation that squads add new and existing MFEs to Devstack, following the pattern established by the MFEs already in Devstack.
```
kdmccormick: @inventhouse and others have raised that there's currently unclear/inconsistent guidance on when and whether to add MFEs to devstack. I think that this PR could be a good pivot point to circulate some new guidance.

@robrap Is this something Arch-BOM would be able to do?

kdmccormick (commented):
Notes/questions for reviewers + some typos I will fix.

docker-compose-host-nfs.yml
doctoryes: After this PR merges, if a dev has the pre-changes devstack started, then pulls the latest devstack changes, then runs `make dev.down`, will all containers be properly stopped?

docs/decisions/0003-backends-depend-on-frontends.rst
```diff
@@ -0,0 +1,60 @@
...
+
+
+Rejected Alternatives
+=====================
+
+* Keep the old dependency relationships, but add convenience targets (such as ``dev.up.domain.learning``) to start groups of related micro-frontends. We determine that this would increase the already-large congnitive overhead of the Devstack interface.
+* Invert dependency relationships as described in this ADR, and also add targets such as ``make dev.up.lms-backend`` in order to start LMS without associated frontends. We determine that this would create a cascade of new inconsistencies in the Devstack interface: since only one of ``lms`` or ``lms-backend`` could exist as a docker-compose service, rules for the other would have to be hard-coded into the Makefile as special cases.
```
iloveagent57: How about another alternative: keep the old dependency relationships, but add a convenience target like `dev.up.frontends` that brings up _all_ the frontend containers.  So if you want the full LMS learning experience, you just need `make dev.up.lms+frontends`.  This has a benefit for developers of frontends who don't use containers to do local MFE development, and instead run node/npm directly on their workstation.

docs/decisions/0003-backends-depend-on-frontends.rst
```diff
@@ -0,0 +1,60 @@
...
+
+Currently, micro-frontend services (those prefixed with ``frontend-app-``) are defined to depend on backends, but not vice versa. So, starting frontend-app-learning will automatically start LMS, but start LMS will not automatically start frontend-app-learning. This makes sense under logic that "frontends depend on the APIs of backends in order to function".
+
+However, it can be argued that the opposite dependency relationship also makes sense. That is, one may assert that backends should depend on frontends in order to expose their APIs in a usable way. One could further assert that frontends shouldn't have hard dependencies on backend APIs, and should instead gracefully degrade when some or all of its APIs are unavailble.
```
doctoryes: `unavailable`

docs/decisions/0003-backends-depend-on-frontends.rst
```diff
@@ -0,0 +1,60 @@
...
+
+
+Decision
+========
+
+Whichever dependency direction (frontends depend on backends, or vice versa) is more logically sound, we conclude that, for the puposes of Devstack, *asserting that backends depend on frontends is more useful to developers*. Specifically, it is beneficial to current and future developer workflows if ``make dev.up.lms`` automatically starts and learning-related frontends, ``make dev.up.studio`` automatically starts all authoring-related frontends, ``make dev.up.ecommerce`` starts all purchasing-related frontends, and so on.
```
doctoryes: It's actually a core assumption that if you start the LMS or Studio in devstack, then the UI/UX of the app will be fully functional. This decision maintains that core assumption - and seems correct to me.

docs/decisions/0003-backends-depend-on-frontends.rst
doctoryes: `submitted`

docs/decisions/0003-backends-depend-on-frontends.rst
```diff
@@ -0,0 +1,60 @@
...
+
+A necessary corollary to this decision is that *all micro-frontends should be available in Devstack*, which is not currently the case. Core frontends such as frontend-app-account and frontend-app-authn missing as of this ADR.
+
+Consequences
+============
```
doctoryes: Perhaps also mention how to use devstack post-merge of these changes while manually running MFEs from a local repo clone via `npm install && npm start`? In order to demonstrate that it's still possible?

doctoryes (commented):
This change makes sense to me - just had a question or two.

docs/decisions/0003-backends-depend-on-frontends.rst
kdmccormick: I spun all up all devstack frontend containers, out of curiosity. This is what I see right after staring them:
```
NAME                                          CPU %     MEM USAGE / LIMIT
edx.devstack.frontend-app-course-authoring    62.50%    125.7MiB / 15.39GiB
edx.devstack.frontend-app-library-authoring   57.16%    114.6MiB / 15.39GiB
edx.devstack.frontend-app-payment             51.38%    116.8MiB / 15.39GiB
edx.devstack.frontend-app-learning            52.26%    129.2MiB / 15.39GiB
edx.devstack.frontend-app-program-console     44.14%    116.5MiB / 15.39GiB
edx.devstack.frontend-app-publisher           60.24%    123.7MiB / 15.39GiB
edx.devstack.frontend-app-gradebook           48.94%    132.9MiB / 15.39GiB

```
and then a couple minutes later after they've all finished running `npm install && npm start`:
```
edx.devstack.frontend-app-course-authoring    0.00%     741.7MiB / 15.39GiB
edx.devstack.frontend-app-payment             0.00%     618.4MiB / 15.39GiB
edx.devstack.frontend-app-publisher           0.00%     568.6MiB / 15.39GiB
edx.devstack.frontend-app-program-console     0.00%     554.2MiB / 15.39GiB
edx.devstack.frontend-app-gradebook           0.00%     422.4MiB / 15.39GiB
edx.devstack.frontend-app-library-authoring   0.00%     422.7MiB / 15.39GiB
edx.devstack.frontend-app-learning            0.00%     485.9MiB / 15.39GiB
```
That's well over 3.5G, with half of edX's frontends not even represented. Oof.

@iloveagent57 I'm not sure whether that supports my proposal or your proposal! In either case, we'll be sometimes be starting frontends that someone doesn't want, in a manner (Docker) that is quite overkill...

docs/decisions/0003-backends-depend-on-frontends.rst
robrap: puposes => purposes

docs/decisions/0003-backends-depend-on-frontends.rst
robrap: As far as this ADR goes, I think we should just make it a requirement that an MFE must be included in devstack following the pattern outlined when it becomes the default functionality for devstack.

You could leave the question of when MFEs must be added to devstack when they are not yet the default devstack behavior as out-of-scope for this particular ADR. Although, devstack is listed in the [MFE Definition of Done](https://openedx.atlassian.net/wiki/spaces/FEDX/pages/1917944547/MFE+Definition+of+Done).

docs/decisions/0003-backends-depend-on-frontends.rst
robrap: It seems that [MFE Definition of Done](https://openedx.atlassian.net/wiki/spaces/FEDX/pages/1917944547/MFE+Definition+of+Done) already includes devstack. It would help to get details about where the guidance exists that is unclear/inconsistent, or whether there are specific MFEs that seem like they'd warrant more nuanced guidance.

docs/decisions/0003-backends-depend-on-frontends.rst
robrap: In all cases, we should add docs (where?) about the workflow for developing in an MFE. Is it simply to stop the devstack MFE container before starting it locally? Do some people develop inside the container?

docker-compose-host-nfs.yml
kdmccormick: Good point; they will not be stopped until Docker itself or the OS is restarted.

I will include running `make dev.down` before git-pulling in the email I send out.

docs/decisions/0003-backends-depend-on-frontends.rst
kdmccormick: @robrap That doc was created in Oct 2020. Most of our current MFEs had been already built and deployed.

Adding MFEs to devstack has always been an ad-hoc sort of thing. I think the old Master's team did it first with frontend-app-gradebook, and the other six devstack MFEs just followed its example. I don't remember any sort of explicit recommendation being sent out that new and existing MFEs should be in devstack, so it's not surprising that only 7 of the ~14 MFEs we have are in devstack.

I recall there being a certain amount of push to *not* add MFEs to devstack as well, since doing so adds to a fully-running devstack's memory footprint, and isn't a clear value-add for teams that are happy with just running `npm install && npm start` locally. I also recall hearing that certain frontend dev workflows *don't* work in devstack, although I don't remember which ones.. perhaps those to do with changing Webpack config or locally developing on frontend packages? cc'ing @davidjoy in case he can add color here.

docs/decisions/0003-backends-depend-on-frontends.rst
kdmccormick: @robrap The goal is that one should be able to develop MFEs as they run in devstack, just like any other micro-service. I do so.

There seems to certain dev use cases, though, that don't work in this flow. I'm not familiar with them. I'm asking David Joy if he can elaborate in this comment thread: https://github.com/edx/devstack/pull/783#discussion_r658256278

docs/decisions/0003-backends-depend-on-frontends.rst
kdmccormick: Agreed, and incorporated.

docs/decisions/0003-backends-depend-on-frontends.rst
kdmccormick: On the basis of memory consumption, I continue to learn towards my proposal (only start MFEs needed by the backend).

Since the LMS will eventually need ~6 frontends, and nearly every service needs LMS for auth, every service may end up spinning up 6+ frontends, which ain't great. However, this could end up being a positive forcing function for extracting auth from LMS, and then decoupling other services from LMS.

(If it turns out that certain dev flows un-fixably *do* require working outside of Docker, though, then I would lean towards @iloveagent57 's solution, since it's a way better story for locally running MFEs)

docs/decisions/0003-backends-depend-on-frontends.rst
davidjoy: Yes... so the issue here is related to module.config.js.  (https://github.com/edx/frontend-build#local-module-configuration-for-webpack)

The mechanism relies on webpack being able to reach up and out of the working directory and into peer directories to find local libraries.  (i.e., an MFE build reaching out of the MFE directory and into your local copy of paragon to grab its assets, rather than the version of paragon downloaded into node_modules)

In a Docker container, none of those peer directories are mounted in the container, meaning that if you have a module.config.js with local modules defined, it explodes and the build fails because it can't resolve the directories.

If it were possible to just mount common libraries if they exist, that'd be an easy solution (i.e., the brand, paragon, frontend-platform, headers, footers, etc.)

So long as people are technically able to run an MFE outside docker by shutting down the docker container and starting the MFE up outside of devstack, that can also be a solution (the one that we use for frontend-app-learning today).  I don't think we should hold up doing the "right" thing w/r/t devstack for this, so long as its still technically possible.

kdmccormick: @davidjoy
> If it were possible to just mount common libraries if they exist, that'd be an easy solution (i.e., the brand, paragon, frontend-platform, headers, footers, etc.)

I think this is doable! We do a similar thing with Python packages in micro-services using the `${DEVSTACK_WORKSPACE}/src` directory, which we mount wholesale into every micro-service. I made [this PR](https://github.com/edx/devstack/pull/790) to do the same thing for frontends. The idea is that devs would put frontend packages into `src/`, which would show up in the Docker container as `/edx/src/<package>`.

I tried this just now, and I'm seeing `Module not found: Can't resolve '@edx/frontend-platform' in '/edx/app/frontend-app-learning/node_modules/@edx/frontend-lib-special-exams/dist/data'`. Perhaps I'm doing the `module.config.json` thing wrong..

If it turns out that `src/` -> `/edx/src` isn't the right mount pattern, we could do something else. Docker-Compose is pretty flexible here; it won't complain if a mount directory doesn't exist. For example, we could also individually mount paragon, frontend-build, frontend-platform, etc. into any directory of the base microfrontend container; perhaps right into `/edx/app`, so that the libraries would be sibling directories with the frontend app code. Happy to experiment here and find something that works.

> So long as people are technically able to run an MFE outside docker by shutting down the docker container and starting the MFE up outside of devstack, that can also be a solution (the one that we use for frontend-app-learning today). I don't think we should hold up doing the "right" thing w/r/t devstack for this, so long as its still technically possible.

The "right" devstack thing is the one that makes development the least painful. Merging this PR with that the caveat that folks will have to work around it a bunch in order to do frontend package development doesn't feel right to me :)

kdmccormick: @davidjoy and I talked, and decided that it would be best to mount an explicit list of common frontend packages, from siblings of the host `devstack/` directory to `/edx/app` within the frontend containers. For example, if devstack lived in `~/devstack`, then `~/paragon` would be mounted to `/edx/app/paragon` of each frontend container (if it exists), allowing the frontend app at `/edx/app/frontend-app-blah` to install paragon using the module.config.js strategy.

This PR attempts that, but I'm getting an error when trying to run frontend-app-learning with local packages: https://github.com/edx/devstack/pull/795

robrap: @kdmccormick: I personally like that `/edx/src` contains all the github repos that I know are used from inside containers, and would prefer that unless there is a strong reason why it needs to be something different (e.g. `/edx/app`). If there is a strong reason, can you document that in the ADR eventually? Thank you!

davidjoy: My recommendation was based on how I expect folks to have their workspace laid out.  Personally I have a `src` directory with all my repos in it.  It feels rather weird that I'd have a `src` directory in my `src` directory where some of my repos need to be if I happen to want to use them locally in an MFE.  

I think the difference is perhaps that needing to use module.config.js is a temporary thing that you do sometimes, but otherwise the projects are independently develop-able and don't exist only in relation to a particular MFE.  Changing the workspace organization to accommodate it feels backward to me.

robrap: I’m still not clear if there is a difference here from Python and its use of the existing src directory that is a sister directory to the main repos in devstack. Maybe a quick look at this next week together?

kdmccormick: @robrap @davidjoy It seems like this is a debate of simplicity-of-implementation (follow what devstack does now) vs. simplicity-of-interface (don't require packages to be stuffed into ~/src) and I can't think of a solution that doesn't sacrifice a bit of one or the other.

Here is my new proposal, which leans towards simplicity-of-implementation: https://github.com/edx/devstack/pull/803. Please let me know what you think, keeping in mind that I'm approaching the maximum amount of time I can spend on this devstack change.

robrap: I think the ADR did a great job clarifying the decision. Thank you!

docker-compose-sync.yml
kdmccormick: Update: sync and NFS are slated for removal soon, so this is even further moot.

docs/decisions/0003-backends-depend-on-frontends.rst
kdmccormick: Good idea. I just added a **Running micro-frontends outside of devstack** section to workflow.rst.

docs/decisions/0005-frontend-package-mounts.rst
```diff
@@ -0,0 +1,111 @@
...
+    frontend-app-XX:
+      volumes:
+        - ${DEVSTACK_WORKSPACE}/frontend-app-XX:/edx/app/frontend-app-XX:cached
+        - frontend_app_XX_node_modules:/edx/app/frontend-app-XX/node_modules
+        - ${DEVSTACK_WORKSPACE}/src:/edx/app/src:cached  # <--- This line is new!
+
+This will cause the ``${DEVSTACK_WORKSPACE}/src`` folder to mounted at ``/edx/app/src`` of each frontend service, similar to how that folder is mounted at ``/edx/src`` of each backend service. Via ``module.config.js``, frontend developers will then be able to specify ``../src/PACKAGE`` as the path of any local frontend package. This scheme has the benefit of:
+
+* working within a frontend Devstack container, since ``../src/PACKAGE`` resolves to ``/edx/app/src/PACKAGE``, and
+* working oustide of Devstack, since ``../src/PACKAGE`` points to ``PACKAGE`` when ``src`` is a sibiling of the frontend application repository.
```
davidjoy:
```suggestion
* working oustide of Devstack, since ``../src/PACKAGE`` points to ``PACKAGE`` when ``src`` is a sibling of the frontend application repository.
```

davidjoy (approved):
This is an approval in spirit more than in detail, as I'm hardly a docker expert.

kdmccormick:
Partially reverting this in #815

PR #784, opened 2021-06-21
fix!: Fix duration metric and rename

timmc-edx:
- `duration` changes name to `duration_ms` (more explicit)
- duration is now correct, as opposed to being milliseconds modulo seconds

The old usage metrics calculation for command duration used the
microseconds property of a timedelta, which only gets the microseconds
*component* of the difference -- not the difference *in* microseconds.
(It was then divided by 1000 to get milliseconds.)

The new calculation takes the difference and then divides it by
milliseconds. (This continues to compute the floor, using integer
division.)

Since the previous data was unusable, the name change doesn't interfere
with any analysis we were doing.

----

- [x] Update metrics page on Confluence
- [x] Update dashboard in New Relic

### DIFF
--- a/scripts/send_metrics.py
+++ b/scripts/send_metrics.py
@@ -34,7 +34,7 @@ import sys
 import traceback
 import urllib.request as req
 import uuid
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from http.client import RemoteDisconnected
 from os import path
 from signal import SIG_DFL, SIGINT, signal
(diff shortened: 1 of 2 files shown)


jinder1s (approved):

jinder1s:
I also modified here: https://openedx.atlassian.net/wiki/spaces/AC/pages/2720432206/Devstack+Metrics

timmc-edx:
OK, I'll update in New Relic, then.

PR #790, opened 2021-06-25
feat: mount src/ into /edx/src of frontend containers

kdmccormick:
Should allow installation of local versions of NPM
packages via module.config.js, allowing frontend devs
to test out frontend library changes within devstack.

The mounts follow the same pattern that micro-services
do, which allows devs to develop local versions of
Python packages alongside devstack micro-services.

(this is a PR into [another PR](https://github.com/edx/devstack/pull/783))


davidjoy (changes requested):
Hm, not sure I understand the syntax there.  The first volume on each has the form like:

`${DEVSTACK_WORKSPACE}/frontend-app-payment`

Whereas the new line is:

`${DEVSTACK_WORKSPACE}/src`

Doesn't that imply that `frontend-app-payment` is a peer of `src`, rather than inside `src`?  

When using module.config.js, the module is a peer of the MFE, i.e.:

`frontend-app-payment`
`paragon`

It seems like this code is implying it would be:

`frontend-app-payment`
`src/paragon`

Which seems wrong.  Am I misunderstanding?

kdmccormick:
@davidjoy You're understanding is right. See my comment here: https://github.com/edx/devstack/pull/783#discussion_r658843094

In short -- I made this PR to match what backend services do right now. If it's not the right approach for frontends, we can adjust this PR so, say, we are mounting sibling packages from the host into sibling packages in the container (so, `paragon` would be mounted at `/edx/app/paragon`, just like `frontend-app-blah` is mounted at `/edx/app/frontend-app-blah`).

I have docker-compose chops but not very many frontend dev chops, so if you have a recommendation on what the best setup is, I can try my hand at making it happen :)

davidjoy:
Heyo - yeah, I think the right thing to do would be to mount the libraries as siblings of the frontend-app, which I think mirrors how everyone has their dev setups organized (I'd guess).  I wish there were a way we could avoid explicitly mounting libraries that we expect people may want to use with module.config.js, as I expect the list will keep expanding over time and folks will get confused when new ones don't work.  frontend-lib-special-exams for example.

Can we mount with a wildcard?  Such as all directories that match `frontend-*`?  If so, then we could explicitly add paragon and then wildcard in everything else that's compatible with module.config.js in a single line, avoiding everything else someone might have in their working directories.  What do you think?

kdmccormick:
Closing in favor of https://github.com/edx/devstack/pull/795

kdmccormick:
Reopening this here: https://github.com/edx/devstack/pull/803

PR #793, opened 2021-07-02
feat!: Don't bring up chrome and firefox containers by default.

feanil:
These containers are only used for e2e, and a11y tests.  Most users
don't need to run these so bringing these containers up as a part of
the default set seems wasteful.

BREAKING CHANGE: Previously these two containers came up by default as
dependencies of the `lms` and `studio` containers.  Now they would have
to be explicitly added as a part of the up command.

eg. `make dev.up.lms+chrome+firefox`

----

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -638,10 +638,10 @@ metrics-opt-out:
 docs: ## generate Sphinx HTML documentation, including API docs
 	tox -e docs
 
-e2e-tests: dev.up.lms+studio ## Run the end-to-end tests against the service containers.
+e2e-tests: dev.up.lms+studio+firefox+chrome ## Run the end-to-end tests against the service containers.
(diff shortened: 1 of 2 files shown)


timmc-edx (approved):
This looks good, but the "default set" in the commit message is a bit ambiguous; this PR doesn't actually change what happens if you run `make dev.up` (which runs the "large and slow" default set). I'm not sure it should, for that matter. :-) Maybe refer instead to not bringing up chrome and firefox for LMS and Studio? Up to you, though.

robrap (commented):
Thanks for this.
1. You should update https://github.com/edx/devstack/blob/1e29afba5572d6db2a18a581e4f0602a074e5fe8/docs/testing_and_debugging.rst#running-end-to-end-tests
2. Don't forget the interface change communication task.

timmc-edx:
@feanil Do you plan to update the e2e testing docs separately?

feanil:
@timmc-edx the e2e docs tell you to use targets that I have updated in this PR so there should not need a doc update.

timmc-edx:
I was thinking of the earlier part of the doc that talks about how to connect to the browser instances, run paver a11y/bokchoy tests, etc. -- those will no longer work unmodified. But I guess we've also been moving away from these browser-based tests, so maybe that's all going away soon?

PR #465, opened 2019-12-22
Using NFS mounted volumes for MacOSX

devalih:
To improve the performance issues found with both dev.sync.up and dev.up.

NFS is the best method we've found so far for the heavy Open edX devstack on MacOSX.

To use it, follow new README instructions on MacOS

PR #692, opened 2021-03-01
Remove unsupported claims about NFS superiority on mac.

iloveagent57:
Metrics or it didn't happen.

PR #797, opened 2021-07-14
Caution MacOS users against blind NFS use

connorhaugh:
Added a note to README.rst which describes potential issues with using nfs.

The current docs lead MacOS users to choose NFS.  Per [Previous Discussion ](https://github.com/edx/devstack/pull/465) The original osxfs Docker back end was replaced by gRPC-FUSE in Docker Desktop Community 2.4.0.0 (released on 2021-09-30), which largely eliminates the need for NFS mounts or other previous file synchronization hacks. People have issues with devstack from time-to-time that seem to go away when they stop using NFS.

A note in the README would not remove functionality, but inform new users of Devstack NFS might not always be the best option.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -163,6 +163,8 @@ The default devstack services can be run by following the steps below.
 
 4. Optional: You have an option to use NFS on MacOS which may improve the performance significantly. To set it up ONLY ON MAC, do:
 
+    Note -
+    Using NFS leads to increased complexity and might cause errors. Improvements to Docker's default FS have made performance improvements negligiable.
     .. code:: sh
 
         make dev.nfs.setup
--- a/README.rst
+++ b/README.rst
@@ -163,8 +163,11 @@ The default devstack services can be run by following the steps below.
 
 4. Optional: You have an option to use NFS on MacOS which may improve the performance significantly. To set it up ONLY ON MAC, do:
 
-    Note -
-    Using NFS leads to increased complexity and might cause errors. Improvements to Docker's default FS have made performance improvements negligible.
+    **Note**
+    Using NFS leads to increased complexity and might cause errors. Improvements to Docker's default FS have made performance improvements negligible.  `Deprecation`_ of NFS is forthcoming. 
+
+    .. _Deprecation: https://openedx.atlassian.net/browse/DEPR-161
+
     .. code:: sh
 
         make dev.nfs.setup
--- a/README.rst
+++ b/README.rst
@@ -164,7 +164,7 @@ The default devstack services can be run by following the steps below.
 4. Optional: You have an option to use NFS on MacOS which may improve the performance significantly. To set it up ONLY ON MAC, do:
 
     **Note**
-    Using NFS leads to increased complexity and might cause errors. Improvements to Docker's default FS have made performance improvements negligible.  `Deprecation`_ of NFS is forthcoming. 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -163,6 +163,8 @@ The default devstack services can be run by following the steps below.
 
 4. Optional: You have an option to use NFS on MacOS which may improve the performance significantly. To set it up ONLY ON MAC, do:
 
+    Note -
+    Using NFS leads to increased complexity and might cause errors. Improvements to Docker's default FS have made performance improvements negligiable.
```
doctoryes: `negligible`

README.rst
```diff
@@ -163,6 +163,8 @@ The default devstack services can be run by following the steps below.
 
 4. Optional: You have an option to use NFS on MacOS which may improve the performance significantly. To set it up ONLY ON MAC, do:
```
doctoryes, line 164: Depending on the current usage of NFS on MacOS *and* whether it still provides a performance boost, a better option here would be to at least deprecate NFS usage - if not remove it altogether. @edx/arch-bom You've been gathering data on Devstack usage - perhaps you'd know how widely used NFS is for devstacks?

dianakhuang: We have a DEPR for remove NFS support going forward, but it hasn't been accepted yet. We're hoping to get rid of it in the next few months: https://openedx.atlassian.net/browse/DEPR-161

README.rst
connorhaugh: Good catch.

README.rst
connorhaugh: It was my understanding from the previous discussion that deprecation was coming. This PR simply aims warn new MACOS users until deprecation.

ormsbee: Maybe make note of the DEPR, and let this merge for now? NFS seems to trip up a lot of folks setting up their devstack for the first time.

ormsbee (approved):

ormsbee:
@connorhaugh: When squashing your commits, please note our [commit conventions](https://open-edx-proposals.readthedocs.io/en/latest/oep-0051-bp-conventional-commits.html).

README.rst
doctoryes: I agree that making note of the DEPR ticket in the instructions will discourage NFS usage even more - and would be a Good Thing™️ .

connorhaugh: Will add a reference to the DEPR ticket in the note.

README.rst
```diff
@@ -163,6 +163,11 @@ The default devstack services can be run by following the steps below.
...
```
robrap, line 164: I highly recommend we remove the following line altogether: 
```
Optional: You have an option to use NFS on MacOS which may improve the performance significantly. To set it up ONLY ON MAC, do:
```
It is not correct and highly misleading.

I think the newly proposed note is reasonable, but the audience should be those that have already set up NFS, assuming we don't get (and don't want) new users of the optional NFS option.

ashultz0:
looks like when this is merged https://github.com/edx/devstack/pull/692 should be rejected and closed as superfluous

robrap:
@ashultz0: In general that seems right, but https://github.com/edx/devstack/pull/692 contains some other formatting fixes as well that could still land.

README.rst
robrap: @connorhaugh: It looks like you already have an approval. Feel free to either make a new edit with my recommendation, or just merge when ready and I can follow-up with another iteration.

README.rst
robrap: Here is my attempt at taking a much stronger stance: https://github.com/edx/devstack/pull/800

PR #799, opened 2021-07-16
Copy NFS deprecation note from main readme to troubleshooting

ashultz0:
Also modifies the sync section below to no longer say "maybe you can
use NFS instead".

Original note from https://github.com/edx/devstack/pull/797

----

- [ N/A ] Made a plan to communicate any major developer interface changes (or N/A)

### DIFF
--- a/docs/troubleshoot_general_tips.rst
+++ b/docs/troubleshoot_general_tips.rst
@@ -189,43 +189,11 @@ through `Understanding Git Conceptually`_. It explains core Git principles in wa
 that makes it easier to use the simpler ``git`` commands more effectively
 and easier to use the more complicated ``git`` commands when you have to.
 
-
-Troubleshooting: Performance
-----------------------------
-
-Improve Mac OSX Performance using nfs
--------------------------------------
-
-.. note::
-    Using NFS leads to increased complexity and might cause errors. Improvements to Docker's default FS have made performance improvements negligible. `Deprecation`_ of NFS is forthcoming.
-
-    .. _Deprecation: https://openedx.atlassian.net/browse/DEPR-161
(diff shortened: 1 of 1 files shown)


timmc-edx (approved):

docs/troubleshoot_general_tips.rst
```diff
@@ -196,15 +196,17 @@ Troubleshooting: Performance
...
 Improve Mac OSX Performance with docker-sync
 --------------------------------------------
 
 .. note::
-    docker-sync is no longer actively supported. See section for nfs above for
-    possible alternative.
+    docker-sync is no longer actively supported
 
 Docker for Mac has known filesystem issues that significantly decrease
 performance for certain use cases, for example running tests in edx-platform. To
```
robrap: @ashultz0 Thanks for finding this. What is the benefit of keeping this at all? Why wouldn't we just delete all of this?

ashultz0: @robrap when I asked Connor about it it seemed like deprecation was a work in progress which is why the main readme got a note rather than just ripping it out. I'd be happy to just delete the nfs section here and maybe the sync section too if you and @timmc-edx think it's time.

robrap: I know that deprecation and removal is a process, but I personally think it is confusing to leave tips that we hope people will not follow. I'd rather just remove altogether. See my related PR as well: https://github.com/edx/devstack/pull/800

ashultz0:
@robrap @timmc-edx went further and removed the sections on improving performance using things we don't want you to use

robrap (approved):

PR #815, opened 2021-08-09
revert: feat!: invert dependencies so that backends now depend on frontends

kdmccormick:
## Description

```
The original commit made it so frontend services were started
as dependencies whenever LMS (or other services) were started.
The increased number of Docker containers is causing resource
depletion for at least one developer.
As a short-term fix, we will revert this commit.

This reverts most of commit 54fb57ffd94d1d723de0920b49922e0100abdae7,
although it leaves in place the changes to docs/workflow.rst, as they
are still relevant.
It also keeps ADR #4, but amends it.
```

## Other information

This reverts part of (but not all of) https://github.com/edx/devstack/pull/783. Specifically, it puts the old dependency relationships back in place, but keeps the changes to frontend app names and frontend package mounting.

I will follow up this merge with a new email to the org.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -13,22 +13,27 @@ The Devstack runs as multiple containers with `Docker Compose`_ at its core.
 
 A Devstack installation includes the following Open edX components by default:
 
-* The Learning Management System (LMS).
-* LMS micro-frontends, including Gradebook and Learning (a.k.a. the "new courseware experience").
+* The Learning Management System (LMS)
+* The Learning micro-frontend (A.K.A the new Courseware experience)
 * Open Response Assessments (ORA2), among other LMS plug-ins.
-* Discussion Forums.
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -13,22 +13,27 @@ The Devstack runs as multiple containers with `Docker Compose`_ at its core.
 
 A Devstack installation includes the following Open edX components by default:
 
-* The Learning Management System (LMS).
-* LMS micro-frontends, including Gradebook and Learning (a.k.a. the "new courseware experience").
+* The Learning Management System (LMS)
```
robrap, line 16: If you want some of these updates, do you plan to use a separate PR?

robrap (approved):

README.rst
kdmccormick: @robrap The README formatting? I don't plan on reintroducing those changes.

PR #816, opened 2021-08-10
feat: Added dev.prune to clean up dangling images, also prune on dev.reset

inventhouse:
I kept looking this command up and it was frequently an issue when I ran `dev.reset`, so I added these in my local workarea; I should've shared this long ago.

----

Not really major, but I'll announce on Slack anyway

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -204,7 +204,7 @@ dev.clone.ssh: ## Clone service repos using SSH method to the parent directory.
 # Developer interface: Docker image management.
 ########################################################################################
 
-dev.prune: ## Prune dangling docker images, useful when you get the 'no space left on device' error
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -204,6 +204,10 @@ dev.clone.ssh: ## Clone service repos using SSH method to the parent directory.
 # Developer interface: Docker image management.
 ########################################################################################
 
+dev.prune: ## Prune dangling docker images, useful when you get the 'no space left on device' error
```
kdmccormick:
```suggestion
dev.prune: ## Prune dangling docker images, containers, and networks. Useful when you get the 'no space left on device' error
```

kdmccormick (approved):
One suggestion, looks good otherwise 👍🏻

PR #818, opened 2021-08-11
feat: add insights service

zacharis278:
Adds the insights service to devstack, this will be followed up by also adding the analytics-data-api.

This build for this image isn't using the same ansible plays as some of the new IDA's so there's a few env differences to get this working with the existing image. Ideally, I'd like to just get this working sooner and revisit changes to the docker image at a later point.

depends on: https://github.com/edx/edx-analytics-dashboard/pull/1166

----


zacharis278:
self reminder: add insights to the table in readme

schenedx (approved):
👍

timmc-edx (approved):
Only scanned through quickly, but looks fairly parallel to how other services are set up, so I'll call that good. :-P

PR #820, opened 2021-08-17
feat: add optional coursegraph service

kdmccormick:
## Description

```
Add a "coursegraph" service to devstack using
the official Neo4j Docker image.

The service is "extra"; that is, it will not start, be
pulled, or be provisioned by default. Instead,
provisioning can be done on-demand,
which will dump data from LMS to Coursegraph:

  make dev.provision.coursegraph

Commands like `dev.up.<service>`, `dev.attach.<service>`, etc
are available for coursegraph, as they are for any other
Devstack service. Further documentation will live in the
edx-platform repo, in:

  ./openedx/core/djangoapps/coursegraph/README.rst

Why add Coursegraph? It is a tool that is relied upon
by many edX developers and support specialists,
yet very few of us know how to operate it or fix it
when it goes down. By making it part of devstack,
it will be easier to debug issues and test out Neo4j
upgrades. Furthermore, this enables developers
to trial complex Cypher queries on custom data
instead of simply trusting what they see returned
by production Coursegraph.

TNL-8386
```

## Supporting information

I needed to run Coursegraph locally in order to debug the issues that edX has been having with its Coursegraph instance ([TNL-8386](https://openedx.atlassian.net/browse/TNL-8386)). The easiest way to do so was via the official Neo4j Docker image. At that point, I'd nearly added to devstack, so I figured I would make this PR to bring it all the way.

Supporting documentation is being added here: https://github.com/edx/edx-platform/pull/28489

## Testing instructions

Follow the testing instructions of https://github.com/edx/edx-platform/pull/28489

## Deadline

None

## Other information

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -255,21 +255,23 @@ impl-dev.provision.%: dev.check-memory ## Provision specified services.
 dev.provision.%: ## Provision specified services.
 	@scripts/send_metrics.py wrap "dev.provision.$*"
 
-dev.backup: dev.up.mysql+mysql57+mongo+elasticsearch+elasticsearch7+elasticsearch710 ## Write all data volumes to the host.
+dev.backup: dev.up.mysql+mysql57+mongo+elasticsearch+elasticsearch7+elasticsearch710+coursegraph ## Write all data volumes to the host.
 	docker run --rm --volumes-from $$(make --silent --no-print-directory dev.print-container.mysql) -v $$(pwd)/.dev/backups:/backup debian:jessie tar zcvf /backup/mysql.tar.gz /var/lib/mysql
 	docker run --rm --volumes-from $$(make --silent --no-print-directory dev.print-container.mysql57) -v $$(pwd)/.dev/backups:/backup debian:jessie tar zcvf /backup/mysql57.tar.gz /var/lib/mysql
 	docker run --rm --volumes-from $$(make --silent --no-print-directory dev.print-container.mongo) -v $$(pwd)/.dev/backups:/backup debian:jessie tar zcvf /backup/mongo.tar.gz /data/db
 	docker run --rm --volumes-from $$(make --silent --no-print-directory dev.print-container.elasticsearch) -v $$(pwd)/.dev/backups:/backup debian:jessie tar zcvf /backup/elasticsearch.tar.gz /usr/share/elasticsearch/data
 	docker run --rm --volumes-from $$(make --silent --no-print-directory dev.print-container.elasticsearch7) -v $$(pwd)/.dev/backups:/backup debian:jessie tar zcvf /backup/elasticsearch7.tar.gz /usr/share/elasticsearch/data
 	docker run --rm --volumes-from $$(make --silent --no-print-directory dev.print-container.elasticsearch710) -v $$(pwd)/.dev/backups:/backup debian:jessie tar zcvf /backup/elasticsearch710.tar.gz /usr/share/elasticsearch/data
+	docker run --rm --volumes-from $$(make --silent --no-print-directory dev.print-container.coursegraph) -v $$(pwd)/.dev/backups:/backup debian:jessie tar zcvf /backup/coursegraph.tar.gz /data
 
-dev.restore: dev.up.mysql+mysql57+mongo+elasticsearch+elasticsearch7+elasticsearch710 ## Restore all data volumes from the host. WILL OVERWRITE ALL EXISTING DATA!
+dev.restore: dev.up.mysql+mysql57+mongo+elasticsearch+elasticsearch7+elasticsearch710+coursegraph ## Restore all data volumes from the host. WILL OVERWRITE ALL EXISTING DATA!
 	docker run --rm --volumes-from $$(make --silent --no-print-directory dev.print-container.mysql) -v $$(pwd)/.dev/backups:/backup debian:jessie tar zxvf /backup/mysql.tar.gz
 	docker run --rm --volumes-from $$(make --silent --no-print-directory dev.print-container.mysql57) -v $$(pwd)/.dev/backups:/backup debian:jessie tar zxvf /backup/mysql57.tar.gz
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -515,6 +515,9 @@ dev.static.%: ## Rebuild static assets for the specified service's container.
 
 dev.reset: dev.down dev.reset-repos dev.prune dev.pull.large-and-slow dev.up.large-and-slow dev.static dev.migrate ## Attempt to reset the local devstack to the master working state without destroying data.
 
+dev.destroy.coursegraph: dev.down.coursegraph ## Remove all coursegraph data.
```
adzuci, line 520: Do we have any other detroy commands for stateful parts of devstack like mongo or mysql?  On one hand I feel like we should have fewer make targets, but on the other if they are useful it could be helpful to have a semi-common syntax for destroying stateful things.

README.rst
```diff
@@ -312,6 +312,8 @@ Instead of a service name or list, you can also run commands like ``make dev.pro
 +------------------------------------+-------------------------------------+----------------+--------------+
 | `xqueue`_                          | http://localhost:18040/api/v1/      | Python/Django  | Extra        |
 +------------------------------------+-------------------------------------+----------------+--------------+
+| `coursegraph`                      | http://localhost:7474/browser       | Java (devtool) | Extra        |
```
adzuci: Is devtool the right wording given that operators and admins who are not devs may use coursegraph?

docker-compose.yml
```diff
@@ -31,6 +31,26 @@ services:
       - ../edx-e2e-tests/upload_files:/edx/app/e2e/edx-e2e-tests/upload_files
       - ../edx-platform/common/test/data:/edx/app/edxapp/edx-platform/common/test/data
 
+  coursegraph:
+    container_name: "edx.${COMPOSE_PROJECT_NAME:-devstack}.coursegraph"
+    hostname: coursegraph.devstack.edx
+    # Try to keep this in sync with the NEO4J_VERSION declared within
```
adzuci, line 37: Thank you for adding this comment!  I think it could be worth adding a reciprical comment to https://github.com/edx/configuration/blob/master/playbooks/roles/neo4j/defaults/main.yml#L25 as well.

docker-compose.yml
```diff
@@ -31,6 +31,26 @@ services:
...
+          - edx.devstack.coursegraph
+    ports:
+      - "7474:7474"  # Expose Web interface at http://localhost:7474.
+      - "7687:7687"  # Expose Bolt interface at bolt://user:password@localhost:7687.
+    volumes:
+      - coursegraph_data:/data
+    stdin_open: true
+    tty: true
+    environment:
+      NEO4J_AUTH: "neo4j/edx"
```
adzuci: I don't understand how `NEO4J_AUTH` is used when spinning up coursegraph in devstack, but I trust that you do and feel comfortable reviewing this PR.

adzuci (approved):
I'm approving this PR with the caveat that I haven't actually tested launching or destroying coursegraph on my local devstack.  If you feel that further testing is needed it may make sense to ask someone to walk through the steps to spin up and down coursgraph locally.

docker-compose.yml
kdmccormick: It's the initial username/password. I'll add a comment clarifying that.

docker-compose.yml
kdmccormick: Will do 👍🏻

README.rst
kdmccormick: Good point -- I'll just call it a "tool". Just want to make it clear that it's not a production service.

Makefile
kdmccormick: No, but we could, and I think they're helpful. I think the more often we can prevent people from running the full `make dev.destroy`, the better.

Makefile
```diff
@@ -515,6 +515,9 @@ dev.static.%: ## Rebuild static assets for the specified service's container.
```
doctoryes, line 517: Perhaps Neo4J's volume should also be included in the `dev.backup` and `dev.restore` commands above in the Makefile?

provision-coursegraph.sh
```diff
@@ -0,0 +1,23 @@
...
+# a weird state.
+docker-compose rm --force --stop coursegraph
+docker-compose pull coursegraph
+docker-compose up -d coursegraph
+
+sleep 10  # Give Neo4j some time to boot up.
+
+echo -e "${GREEN}   Updating LMS courses in Coursegraph...${NC}"
+
+docker-compose exec lms bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform/ && ./manage.py lms dump_to_neo4j --host coursegraph.devstack.edx --user neo4j --password edx'
```
doctoryes, line 21: Should the Makefile support a re-syncing command that's outside of the more extreme action of re-provisioning? Or is re-provisioning the prescribed way of re-syncing the LMS course data?

doctoryes (commented):
This addition is long overdue and a huge step towards us being able to maintain coursegraph. 👍🏼

Makefile
kdmccormick: Good idea, I'll add that.

provision-coursegraph.sh
kdmccormick: @doctoryes yeah, I went back and forth between having a less-extreme command like `make dev.sync-coursegraph`, versus just stuffing it all into provisioning.

The advantage of the current setup is that it'll be trivial to push out updates to the coursegraph image, since folks will need to pull down the new image in order to sync data. The disadvantage is that it adds ~30 seconds to the syncing process. 

What do you think?

Makefile
kdmccormick: Done, and tested.

provision-coursegraph.sh
doctoryes: I guess it depends on how frequently/quickly developers will want to re-sync the data. Going with this solution at first seems fine. If someone needs a tighter iteration loop, then it's no longer YAGNI and they can add it?

doctoryes (approved):
👍🏼

provision-coursegraph.sh
kdmccormick: I like that take 👍🏻

PR #823, opened 2021-08-18
fix: in coursegraph provision, run lms command instead of exec'ing 

kdmccormick:
```
This ensures that the provisioning script
works whether or not the lms was already started.
If an lms container doesn't exist, then `docker-compose run`
ensures that one is created to run the management
command.

Using `docker-compose exec`, the script would fail with
"no container lms_1" if lms wasn't already running.
```

quick fix to https://github.com/edx/devstack/pull/820


doctoryes (approved):
👍🏼

PR #826, opened 2021-08-24
docs: remove docker-sync docs

robrap:
Docker sync has been deprecated in DEPR-162, and these instructions
were misleading for Mac users. The instructions were updated to simply
state that users should not be using docker-sync.

----


kenclary (approved):
thanks for fixing this :)

PR #843, opened 2021-10-13
feat: allow manual triggering of provisioning tests and alert OpsGenie on failure

rgraber:
Allow the provisioning-tests workflow to be triggered manually for eventual use by other repositories (so they can trigger devstack tests on updates). Also set up OpsGenie alerts for test failures on master that autoclose if the build subsequently succeeds.

- [N/A ] Made a plan to communicate any major developer interface changes (or N/A)

### DIFF
--- a/.github/workflows/provisioning-tests.yml
+++ b/.github/workflows/provisioning-tests.yml
@@ -9,8 +9,8 @@ on:
     branches:
       - '**'
   schedule:
-    # run at 9:30 am M-F
-    - cron: '30 13 * * 1-5'
+    # run at 7:30 am M-F
+    - cron: '30 11 * * 1-5'
 
 jobs:
 
--- a/.github/workflows/provisioning-tests.yml
+++ b/.github/workflows/provisioning-tests.yml
@@ -8,6 +8,12 @@ on:
   pull_request:
     branches:
       - '**'
+  workflow_dispatch:
+      inputs:
+          calling_repo:
+              description: 'Name of the repository triggering this action'
+              required: false
+              default: '@edx/devstack'
 
 jobs:
 
@@ -32,6 +38,15 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
+
+      - name: set calling repository variable
+        run: |
+          if [ -z ${{ github.events.inputs.calling_repo }} ]
+          then
+            echo "CALLING_REPOSITORY=@edx/devstack" >> $GITHUB_ENV
+          else
+            echo "CALLING_REPOSITORY=${{ github.events.inputs.calling_repo }}" >> $GITHUB_ENV
+          fi
 
       - name: installations and version upgrades
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/provisioning-tests.yml
```diff
@@ -66,5 +81,31 @@ jobs:
...
+        uses: dawidd6/action-send-mail@v3
+        with:
+           server_address: email-smtp.us-east-1.amazonaws.com
+           server_port: 465
+           username: ${{secrets.EDX_SMTP_USERNAME}}
+           password: ${{secrets.EDX_SMTP_PASSWORD}}
+           subject: 'Failure: Devstack provisioning tests for ${{matrix.services}} #${{github.run_id}}'
+           to: devstack-provisioning-tests@edx.opsgenie.net
+           from: github-actions <github-actions@edx.org>
+           body: Devstack provisioning tests in ${{github.repository}} for ${{matrix.services}} failed! For details see "github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}". Run triggered by ${{ env.CALLING_REPOSITORY }}.
```
timmc-edx: Nit: Adding protocol to URL.
```suggestion
           body: Devstack provisioning tests in ${{github.repository}} for ${{matrix.services}} failed! For details see "https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}". Run triggered by ${{ env.CALLING_REPOSITORY }}.
```

We'll also likely want to link to a runbook or otherwise indicate what the receiver of the alert should do.

timmc-edx:
The email-sending part looks great. I'm less confident in the remote-call thing, mostly because some of our IDAs have a lot of merges (edx-platform, I'm looking at you) and each remote trigger would, as currently configured, run provisioning tests for a larger number of IDAs. I wonder if we should just run this on a daily schedule instead.

rgraber:
> The email-sending part looks great. I'm less confident in the remote-call thing, mostly because some of our IDAs have a lot of merges (edx-platform, I'm looking at you) and each remote trigger would, as currently configured, run provisioning tests for a larger number of IDAs. I wonder if we should just run this on a daily schedule instead.

I'm going to chat with SRE to see what this might look like and how we might mitigate a flood of requests. For now, I think having the manual trigger option is useful anyway

rgraber:
> > The email-sending part looks great. I'm less confident in the remote-call thing, mostly because some of our IDAs have a lot of merges (edx-platform, I'm looking at you) and each remote trigger would, as currently configured, run provisioning tests for a larger number of IDAs. I wonder if we should just run this on a daily schedule instead.
> 
> I'm going to chat with SRE to see what this might look like and how we might mitigate a flood of requests. For now, I think having the manual trigger option is useful anyway

After talking to SRE, going with the scheduling option

.github/workflows/provisioning-tests.yml
```diff
@@ -8,6 +8,9 @@ on:
   pull_request:
     branches:
       - '**'
+  schedule:
+    # run at 9:30 am M-F
+    - cron: '30 13 * * 1-5'
```
timmc-edx: [suggestion] I wonder if we want it a little earlier so that the alert/email is ready by the time we come in in the morning.

rgraber: oh good call. I was thinking about lining it up with the 9:30 alerts but didn't remember to align the ending rather than the start

timmc-edx (approved):
Looks great. I think this will be a big improvement just by itself, and maybe we can do the more advanced trigger-on-IDA-merge in a future PR.

PR #847, opened 2021-10-21
feat: add analytics data api

zacharis278:
@edx/masters-devs-cosmonauts 

One notable change here is I have updated the mysql users for analytics to be more inline with the other devstack services.

Depends on: https://github.com/edx/edx-analytics-data-api/pull/495


docker-compose.yml
```diff
@@ -173,8 +173,8 @@ services:
       default:
         aliases:
           - edx.devstack.mysql57
-    # ports:
-    #   - "3506:3306"
+    ports:
+      - "3506:3306"
```
ashultz0: I'm not docker syntax knowledgable but this change sticks out as unexpected, adding the analytics API means we have to change the mysql everyone uses?

zacharis278: definitely unintended. Not the first time I've almost merged this local change 🙄

zacharis278: I'm going to advocate for removing this comment altogether but I'll save that for it's own PR.

check.sh
```diff
@@ -115,6 +115,12 @@ if should_check insights; then
         "curl --fail -L http://localhost:18110/health/"
 fi
 
+if should_check analyticsapi; then
+    echo "Running Analytics Data API Devstack tests: "
+    run_check analyticsapi_heartbeat analyticsapi \
+        "curl --fail -L http://localhost:18100/health/"
```
kdmccormick:
```suggestion
        "curl --fail -L http://localhost:19001/health/"
```

zacharis278: thanks for catching that!

schenedx (approved):
👍

PR #854, opened 2021-11-08
feat: add kafka containers to devstack

rgraber:
Add kafka services to devstack. These include:
- zookeeper: keep track of Kafka nodes, topics, and messages. See https://dattell.com/data-architecture-blog/what-is-zookeeper-how-does-it-support-kafka/ for a simple but thorough explanation of what it does
- kafka: the actual message broker; specifically, the storage layer
- schema-registry: storage layer for data schemas
- kafka-rest-proxy: RESTful interface to the Kafka cluster, for easier local debugging and development
----

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -133,6 +133,60 @@ services:
       - ../edx-e2e-tests/upload_files:/edx/app/e2e/edx-e2e-tests/upload_files
       - ../edx-platform/common/test/data:/edx/app/edxapp/edx-platform/common/test/data
 
+  # Events broker
+  kafka:
+    image: confluentinc/cp-server:6.2.1
+    hostname: kafka.devstack.edx
+    container_name: "edx.${COMPOSE_PROJECT_NAME:-devstack}.kafka"
+    depends_on:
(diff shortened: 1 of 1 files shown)


regisb:
Why is Kafka necessary in the devstack?

rgraber:
> Why is Kafka necessary in the devstack?

We are in the process of developing an event bus for the platform, and for now kafka is our broker of choice. Adding it to devstack will be a big aid for local development.

docker-compose.yml
```diff
@@ -188,6 +188,89 @@ services:
         aliases:
           - edx.devstack.redis
 
+  zookeeper:
```
dianakhuang: I think we should add some comments about what these containers are here, because without context, `zookeeper` and `kafka` are pretty confusing.

dianakhuang (approved):
Looks good on my machine, but I would recommend making sure a Linux dev also gives it a shot.

rgraber:
@timmc-edx confirmed it on linux so I'm going to go ahead and merge

PR #855, opened 2021-11-09
fix: Update insights port for provisioning.

dianakhuang:
The port for insights on devstack was updated, but
the provisioning script still pointed to the old port.

This was a bug pointed out on the open edX Slack:
https://openedx.slack.com/archives/CDAG4KN2C/p1636442134001000

### DIFF
--- a/provision-insights.sh
+++ b/provision-insights.sh
@@ -5,7 +5,7 @@ set -eu -o pipefail
 set -x
 
 name=insights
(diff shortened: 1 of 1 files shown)


dianakhuang:
fyi @Zacharis278 . Seemed like a small enough fix, but I want to make sure this actually fixes the issue.

PR #867, opened 2021-12-07
feat: replace rest proxy with kafka control center

rgraber:
The rest proxy was not as useful for monitoring/debugging kafka work as previously hoped. The control center is much easier to use. It is expensive on memory and CPU, but given that it is not actually required for anything except local debugging specific to kafka work it shouldn't be an undue burden.

result of `docker stats edx.devstack.kafka-control-center`  \:
```CONTAINER ID   NAME                                CPU %     MEM USAGE / LIMIT     MEM %     NET I/O           BLOCK I/O         PIDS
ee153703ae73   edx.devstack.kafka-control-center   11.73%    677.1MiB / 9.731GiB   6.80%     21.2MB / 14.6MB   91.7MB / 9.25MB   146
```
----

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -167,7 +167,7 @@ services:
         aliases:
           - edx.devstack.kafka
 
-  # browser app for monitoring local Kafka cluster
+  # browser app for monitoring local Kafka cluster. This is quite memory- and CPU-intensive, so it should only be used for local Kafka debugging
   kafka-control-center:
     image: confluentinc/cp-enterprise-control-center:6.2.1
     hostname: kafka-control-center.devstack.edx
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -167,25 +167,28 @@ services:
         aliases:
           - edx.devstack.kafka
 
-  # RESTful interface to the Kafka cluster
-  kafka-rest-proxy:
-    image: confluentinc/cp-kafka-rest:6.2.1
+  # browser app for monitoring local Kafka cluster
```
dianakhuang: Given the memory/CPU constraints, I would also leave a comment about when you want to bring up this container (i.e. only when doing Kafka debugging.)

dianakhuang (approved):

PR #891, opened 2022-01-21
build: use reusable pr-automerge-open-release workflow

nedbat:
Oops, one last PR to get the description correct at the top.


feanil (approved):
Seems like a good update but I also wonder if it makes more sense for that to live in the upstream workflow itself and then just link to that from here?

nedbat:
Hmm, good point.  I'll update throughout.

PR #892, opened 2022-01-24
fix: create reports_v1 db as well for analytics-data-api service

schenedx:
When provisioning the analytics-data-api service, the database `reports` was created, but the `reports_v1` db was not.
This is not a true reflection of the analytics-data-api's needs on multiple databases.

Make sure we also create the `reports_v1` db on devstack to ensure `make dev.provision.analyticsapi` command is going to be successful.

@openedx/masters-devs-cosmonauts Please review


alangsto (approved):
👍

michaelroytman (approved):

zacharis278 (approved):

PR #894, opened 2022-01-27
feat: add frontend app authn

zainab-amir:
Ticket: https://openedx.atlassian.net/browse/VAN-315

----

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -308,6 +308,8 @@ Instead of a service name or list, you can also run commands like ``make dev.pro
 +------------------------------------+-------------------------------------+----------------+--------------+
 | `frontend-app-gradebook`_          | http://localhost:1994/              | MFE (React.js) | Default      |
 +------------------------------------+-------------------------------------+----------------+--------------+
+| `frontend-app-authn`_              | http://localhost:1999/              | MFE (React.js) | Default      |
++------------------------------------+-------------------------------------+----------------+--------------+
 | `registrar`_                       | http://localhost:18734/api-docs/    | Python/Django  | Extra        |
 +------------------------------------+-------------------------------------+----------------+--------------+
 | `frontend-app-program-console`_    | http://localhost:1976/              | MFE (React.js) | Extra        |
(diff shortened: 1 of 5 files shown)


waheedahmed (approved):

kdmccormick:
@zainab-amir you asked in Slack:
> Does this change need to be reverted?
> Asking because @robrap  wanted me to confirm this from you.
> ADR: https://github.com/openedx/devstack/blob/master/docs/decisions/0004-backends-depend-on-frontends.rst

First of all: No, I do not think the entire change needs to be reverted. It's good that the Authn MFE is available in devstack!

The concern we found with the ADR you linked (described under "status: reverted") is that it would be unsustainable if starting the LMS also started dozens of MFEs automatically because of the high resource usage of each MFE's container. So, for the majority of MFEs, we decided that the LMS should **not** start them by default; instead, it's up to devstack users to start all the MFEs that they are going to use. By that logic, I would recommend that you keep the majority of this PR, but revert [the part that makes LMS require Authn MFE](https://github.com/openedx/devstack/pull/894/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R421-R432), instead leaving it the MFE as an optional component.

On the other hand, the Authn MFE is special because the login+registration flow is a gateway to almost every other feature of the platform... so, if the LMS **were** going to require a single MFE, it would be the Authn MFE. By that logic, it could be fine to leave this PR in place entirely.

I do not have a strong opinion either way. If I had to give a personal recommendation, I would say just leave it as-is and reevaluate if it ever becomes a problem. Our current plan is to replace Devstack with Tutor, so hopefully you won't have to worry about this for long.

robrap:
I like @kdmccormick’s recommendation of making this an exception unless and until it becomes a problem. Two possible follow-ups:
1. Note in the ADR, or add a new ADR, that although the original ADR remains rejected in general, an exception is being made for authn.
2. Send a communication about all of this so others do not think it is ok to add all ne MFEs by default. Noting that it is still ok to add the MFEs, if that was confusing.

waheedahmed:
Thank you for the thoughts @kdmccormick and @robrap!

@zainab-amir I guess let's remove the Authn MFE from LMS dependency for now since it's not quite yet ready for the community until we complete the optional fields work. Once that is done, maybe we can then add it back and announce it community-wide.

PR #902, opened 2022-02-16
fix: don't enable Authn MFE by default

zainab-amir:
Removing it as Authn MFE has not been officially released for openedX. 
----


waheedahmed (approved):
I guess, let's remove the container from the LMS dependency as well.

waheedahmed (approved):

PR #911, opened 2022-03-11
feat: try hitting mysql twice to account for restart

rgraber:
----

### DIFF
--- a/provision.sh
+++ b/provision.sh
@@ -138,7 +138,15 @@ done
 
 # In the event of a fresh MySQL container, wait a few seconds for the server to restart
 # See https://github.com/docker-library/mysql/issues/245 for why this is necessary.
-sleep 10
+sleep 2
+
+echo "${GREEN}Waiting for MySQL 5.7.${NC}"
+until docker-compose exec -T mysql57 bash -e -c "mysql -uroot -se \"SELECT EXISTS(SELECT 1 FROM mysql.user WHERE user = 'root')\"" &> /dev/null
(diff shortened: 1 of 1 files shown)


provision.sh
```diff
@@ -139,6 +139,14 @@ done
 # In the event of a fresh MySQL container, wait a few seconds for the server to restart
 # See https://github.com/docker-library/mysql/issues/245 for why this is necessary.
 sleep 10
+
+echo "${GREEN}Waiting for MySQL 5.7.${NC} to restart"
```
robrap: I’m not sure what ${NC} is, but the format of this message is a bit different from the following message.

provision.sh
```diff
@@ -139,6 +139,14 @@ done
...
+until docker-compose exec -T mysql57 bash -e -c "mysql -uroot -se \"SELECT EXISTS(SELECT 1 FROM mysql.user WHERE user = 'root')\"" &> /dev/null
+do
+  printf "."
+  sleep 1
```
robrap, line 147: Do we need a max tries that will also break the loop? I see we don’t have that for mongo. Is that because the overall script has a timeout?

provision.sh
rgraber: Good catch. $NC turns the color off.

provision.sh
rgraber: I'm actually not sure. Looking around now.

robrap (approved):

PR #920, opened 2022-04-13
feat: apple-silicon support

johnnagro:
## Description

Putting together a branch that supports Apple M1 macs. So far it is backwards compatible with Intel macs but reserving the right to make breaking changes.

## Unresolved Issues

- some node/npm `pngquant` version incompatibilities 

## Notes

- M1 platform is `arm64v8`, although we haven't had to specify that anywhere (yet)
- MySQL volumes are renamed to preserve old data and avoid data incompatibilities for Intel users who are upgrading

## Changes

- [platform supported mysql image](https://hub.docker.com/r/arm64v8/mysql/)
- [mysql volumes and images are renamed](https://github.com/openedx/devstack/pull/920/commits/b46e7c935e9c86aacab523fa746ce27a8f481570)
- [updated user/grant syntax for mysql8](https://stackoverflow.com/a/68205477)
- [redis server workaround](https://github.com/redis/redis/issues/8062#issuecomment-749311875)

## Testing

- so far mostly local testing on volunteer laptops
- we have a request into AWS for access to M1 hardware (in preview) for a CI setup

## Related Work

- https://github.com/openedx/devstack/pull/852
- https://github.com/openedx/license-manager/pull/416
- https://github.com/edx/enterprise-access/pull/55
- https://github.com/openedx/enterprise-catalog/pull/438
- https://github.com/openedx/license-manager/pull/417
- https://github.com/openedx/license-manager/pull/421
- https://github.com/openedx/enterprise-catalog/pull/447
- https://github.com/edx/enterprise-access/pull/59

----


jmbowman:
There's a previous effort towards this in https://github.com/openedx/devstack/pull/852 that was abandoned after hitting a couple of blockers, it may be useful for reference.

docker-compose.yml
```diff
@@ -258,7 +258,7 @@ services:
     container_name: "edx.${COMPOSE_PROJECT_NAME:-devstack}.redis"
     hostname: redis.devstack.edx
     image: redis:2.8
-    command: redis-server --requirepass password
+    command: env -i /usr/local/bin/redis-server --requirepass password
```
johnnagro: redis 5 shouldn't require this hack if we chose to upgrade

docker-compose.yml
```diff
@@ -237,28 +237,28 @@ services:
     volumes:
       - mongo_data:/data/db
 
-  mysql57:
+  mysql80:
```
johnnagro, line 240: The dev-env working group decided the most reasonable path forward here was to rename MySQL-related docker assets. this avoids data incompatibility and allows existing volumes (with data) to be preserved in the event we want to access that preexisting data.

docker-compose.yml
```diff
@@ -237,28 +237,29 @@ services:
...
     environment:
       MYSQL_ROOT_PASSWORD: ""
       MYSQL_ALLOW_EMPTY_PASSWORD: "yes"
-    image: mysql:5.7
+    image: mysql:8.0.28-oracle
     networks:
       default:
         aliases:
+          - edx.devstack.mysql80
           - edx.devstack.mysql57
```
johnnagro, line 252: keeping the old `edx.devstack.mysql57` alias provides backwards comparability for projects with local config that point to the old hostname

docker-compose.yml
```diff
@@ -237,28 +237,29 @@ services:
...
     command: mysqld --character-set-server=utf8 --collation-server=utf8_general_ci
-    container_name: "edx.${COMPOSE_PROJECT_NAME:-devstack}.mysql57"
-    hostname: mysql57.devstack.edx
+    container_name: "edx.${COMPOSE_PROJECT_NAME:-devstack}.mysql80"
+    hostname: mysql80.devstack.edx
     environment:
       MYSQL_ROOT_PASSWORD: ""
       MYSQL_ALLOW_EMPTY_PASSWORD: "yes"
-    image: mysql:5.7
+    image: mysql:8.0.28-oracle
```
rgraber, line 247: curiosity - why the oracle version?

docker-compose.yml
rgraber: Should we put an expiration date on this?

rgraber (commented):
Some questions, but it looks good to me. Definitely would advise waiting on review from Jeremy or someone with more authority

rgraber:
Just double checking before I approve: developers who are using devstack right now will not lose any data when they update their devstack from this PR, right? Will they need to do anything else (make dev.pull, I imagine?)? @johnnagro

johnnagro:
The PR #968 seems to work and would negate the need for this `apple-silicon` branch. It also provides a possible solution for other incompatible images (use buildx ourselves to create compatible versions of public Dockerfiles)

johnnagro:
We've been able to build/use a cross-platform-compatible Mysql5.7 image via https://github.com/openedx/devstack/pull/968 and no longer need this `apple-silicon` branch / Mysql8.0 attempt to get around the M1 incompatibility. I can work with individual users if they want to move data back into mainline devstack - just reach out to me.

PR #925, opened 2022-04-26
feat: Added Opensearch 1.2.0

syedimranhassan:
----
https://openedx.atlassian.net/browse/ISRE-1306
Adding OpenSearch for LMS, studio, cs_comments, and blockstore.
Our plan is to replace ES with OS for these services.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -118,6 +118,25 @@ services:
       - bootstrap.memory_lock=true
       - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
 
+  # This is meant to be used to test OS upgrades.
+  opensearch12:
+    container_name: "edx.${COMPOSE_PROJECT_NAME:-devstack}.opensearch12"
+    hostname: opensearch12.devstack.edx
+    image: opensearchproject/opensearch:1.2.0
+    networks:
+      default:
+        aliases:
+          - edx.devstack.opensearch12
+    ports:
+      - "9202:9200"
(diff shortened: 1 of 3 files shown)


NIXKnight (approved):
LGTM

docker-compose.yml
```diff
@@ -418,6 +438,7 @@ services:
       - memcached
       - mongo
       - mysql57
+      - opensearch12
```
robrap, line 441: 1. Can you update the PR description with a little more background about how and when each of the services will rely on this?
2. Will this just be needed for a rare use case? I’m not clear on what “test OS upgrades” entails, but it seems like a more rare activity and that maybe this isn’t a required dependency for most uses of LMS or other services? This is a question. If not, maybe there should simply be doc changes for when and how to bring this container up?

dianakhuang: I think the best reference for this is the Elasticsearch -> OpenSearch ticket we have on the DEPR board. We are migrating all our services off Elasticsearch and for those that can't use MySQL full text search, we're going to use OpenSearch. https://github.com/openedx/public-engineering/issues/16

syedimranhassan: Yeah, we are migrating LMS, studio, cs_comments(forum), and blockstore from Elasticsearch -> OpenSearch.

robrap: Ok. Seems fine if you think we need both dependencies now.

robrap (approved):

PR #976, opened 2022-10-07
Insights-configuration for docker

aht007:
----

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -451,7 +451,7 @@ dev.shell.analyticsapi:
 	docker exec -it edx.devstack.analyticsapi env TERM=$(TERM) /edx/app/analytics_api/devstack.sh open
 
 dev.shell.insights:
-	docker-compose exec insights env TERM=$(TERM) bash -c 'eval $$(source /edx/app/insights/insights_env; echo PATH="$$PATH";) && /edx/app/insights/devstack.sh open'
+	docker-compose exec insights env TERM=$(TERM) bash -c 'eval $$(source /edx/app/insights/insights_env; echo PATH="$$PATH";) && /bin/bash'
 
 dev.shell.%: ## Run a shell on the specified service's container.
 	docker-compose exec $* /bin/bash
--- a/configuration_files/insights.yml
+++ b/configuration_files/insights.yml
@@ -1,0 +1,58 @@
+---
+
+APPLICATION_NAME: Insights
+BACKEND_SERVICE_EDX_OAUTH2_PROVIDER_URL: http://edx.devstack.lms:18000/oauth2
+CACHES:
+    default:
+        BACKEND: django.core.cache.backends.memcached.MemcachedCache
+        KEY_PREFIX: default_env-default_deployment-insights
+        LOCATION:
+        - edx.devstack.memcached:11211
+CDN_DOMAIN: null
+CMS_COURSE_SHORTCUT_BASE_URL: http://edx.devstack.lms:18000/course
+COURSE_API_URL: http://edx.devstack.lms:18000/api/courses/v1/
+CSRF_COOKIE_NAME: insights_csrftoken
+CSRF_COOKIE_SECURE: false
+DATABASES:
+    default:
+        ENGINE: django.db.backends.mysql
+        HOST: edx.devstack.mysql
+        NAME: dashboard
+        PASSWORD: secret
+        PORT: '3306'
+        USER: rosencrantz
+DATA_API_AUTH_TOKEN: edx
+DATA_API_URL: http://edx.devstack.analyticsapi:18100/api/v0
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -451,7 +451,7 @@ dev.shell.analyticsapi:
 	docker exec -it edx.devstack.analyticsapi env TERM=$(TERM) /edx/app/analytics_api/devstack.sh open
 
 dev.shell.insights:
-	docker-compose exec insights env TERM=$(TERM) bash -c 'eval $$(source /edx/app/insights/insights_env; echo PATH="$$PATH";) && /edx/app/insights/devstack.sh open'
+	docker-compose exec insights env TERM=$(TERM) bash -c 'eval $$(source /edx/app/insights/insights_env; echo PATH="$$PATH";) && /bin/bash'
```
aht007, line 454: We have removed the `devstack.sh` file usage in the newer image.
Essentially the file was running these commands in this step:

>         . /edx/app/insights/nodeenvs/insights/bin/activate
>         . /edx/app/insights/venvs/insights/bin/activate
>         cd /edx/app/insights/insights
> 
>         /bin/bash

We have already done all these above steps by setting the WORKDIR in docker image and also putting the nodeenv and python virtual env in path.

ashultz0:
things are going wrong as I test in ways that seem beyond my lack of docker expertise - short version is /edx/app/insights/insights/ seems to now be at /edx/app/insights/edx_analytics_dashboard/ and a few things need to be changed accordingly. That move seems to be part of https://github.com/openedx/edx-analytics-dashboard/pull/1333

what follows is what I did thrashing around with my limited docker experience in devstack:

changed my docker-compose to `image: openedx/insights-dev:${OPENEDX_RELEASE:-latest}` per slack and ran successfully with my local checkout after resolving a bunch of needed migrations and provisioning

then I removed the local checkout and insights would not come up:
`python: can't open file '/edx/app/insights/insights/manage.py': [Errno 2] No such file or directory`

when I went to the insights shell I found that there was no `edx/app/insights/insights/` instead code was at `edx/app/insights/edx_analytics_dashboard/`

Searching for insights/insights I changed the call in `docker-compose.yml` to point at `edx/app/insights/edx_analytics_dashboard/manage.py` instead and insights can now start but cannot call 

so I go to provision again and find that the provision script also points at insights/insights and has to be changed to point at insights/edx_analytics_dashboard which goes fine until it has to talk to the lms to provision. The provision script recreates the lms container and seems to break it:
`python: can't open file '/edx/app/edxapp/edx-platform/manage.py': [Errno 2] No such file or directory`

but dev.up.lms works fine by itself, it's only the provision script that brings up a mangled container. I went through the rest of provisioning manually and that seems to work, and then reprovisioned the analytics API.

Also had to make static directly in the container after trying to just use the command in the weirdly broken provisioning flow, that does not seem to have actually webpacked right. I'm not sure what's going on there.

So eventually I got it running but I'm pretty sure that these two repos have to agree on insights/insights vs insights/edx_analytics_dashboard/ and that ends up in a weird number of places. It's unfortunate insights was ever named differently than the repo name but that's long in the past.

aht007:
@ashultz0 `edx/app/insights/insights` is still at its place as well as `edx/app/insights/edx-analytics-dashboard`. The difference between both of them is that `edx-analytics-dashboard` directory is the code directory which was also previously present in Ansible based Images and it hosts the code that is copied while building the image. On the other hand, `edx/app/insights/insights` is the directory that is mounted from docker-compose and this behavior was same previously as well. It is still mounted and it is still present in the new Docker based image.
References for above mentioned files and Ansible configurations:

- [Insights_code_dir](https://github.com/openedx/configuration/blob/master/playbooks/roles/insights/defaults/main.yml#L206)
- [edx/app/insights/insights is mounted](https://github.com/openedx/devstack/blob/master/docker-compose-host.yml#L54)

Can you please explain a bit more about the steps that you performed while testing this and what checkout are you talking about?

ashultz0:
@aht007 the checkout is that because I've done a bunch of work on insights in the past year I had an existing insights checkout, so my initial run was just on that and then I moved it out of the way to try to use the provided dashboard. Maybe the only way to test this would be on a completely clean devstack but I was hoping to spend less than a day on the test.

aht007:
@ashultz0 Maybe your checkout somehow messed up with the mounted directory at `/edx/app/insights/insights`, nevertheless it should still be present no matter what. Would it be possible for you to test the changes out again?

ashultz0:
I'm trying to make a separate clean devstack locally to test, we'll see if that works or if they just get tangled together at the docker layer

ashultz0 (approved):
it got a bit tangled but I think I understand how it got tangled trying to be clean last time as well as having some more ammunition for why we should fix devstack... and in the end insights is up and running ok.

PR #980, opened 2022-10-19
chore: Updating Python Requirements

UsamaSadiq:
Scheduled upgrade requirements workflow [build](https://github.com/openedx/devstack/actions/runs/3278457653/jobs/5396863583) failed to create pull request so triggered the PR manually.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -16,7 +16,7 @@ cffi==1.15.1
     #   pynacl
 charset-normalizer==2.1.1
     # via requests
-cryptography==38.0.2
+cryptography==38.0.1
     # via paramiko
 distro==1.8.0
     # via docker-compose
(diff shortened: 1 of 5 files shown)


requirements/base.txt
```diff
@@ -16,7 +16,7 @@ cffi==1.15.1
     #   pynacl
 charset-normalizer==2.1.1
     # via requests
-cryptography==38.0.2
+cryptography==38.0.1
```
aht007, line 19: Why are we having a downgrade here 🙄

UsamaSadiq: `cryptography==38.0.2` has been marked as `yanked` on PyPI so the version is automatically being downgraded. For details, see https://pypi.org/project/cryptography/#history

aht007 (approved):

PR #997, opened 2022-12-16
fix: recreate base sql init files to deal with migration squash

johnnagro:
- looks like a squash put the migration records in a funky state
- needed a new init
- in the process of generating one, found a syntax issue with a user init script improperly escaping a shell command


johnnagro:
note: without using the init db sql it passes
Screen Shot 2022-12-16 at 3 55 27 PM

PR #1021, opened 2023-02-15
feat: Rename dev.down to dev.remove-containers.

dianakhuang:
This also adds some help text and updates documentation.

----


dianakhuang:
Note that I did not update `workflow.rst`, because there's more cleanup that has to be done there, and I think the big changes should be done as part of the next ticket: https://github.com/openedx/devstack/issues/1014

rgraber (approved):
LGTM

PR #1022, opened 2023-02-17
fix: remove escapes from demo_hash_password

rgraber:
Leaving the escapes in results in `CommandError: The password hash provided for user honor is invalid.` when trying to provision for the first time. See https://github.com/openedx/devstack/pull/997.

----

### DIFF
--- a/provision-lms.sh
+++ b/provision-lms.sh
@@ -57,9 +57,9 @@ for user in honor audit verified staff ; do
   email="$user@example.com"
   # Set staff flag for staff user
   if [[ $user == "staff" ]] ; then
-    docker-compose exec -T lms bash -e -c "source /edx/app/edxapp/edxapp_env && python /edx/app/edxapp/edx-platform/manage.py lms --settings=devstack_docker --service-variant lms manage_user $user $email --initial-password-hash \'$demo_hashed_password\' --staff"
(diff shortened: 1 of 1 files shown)


dianakhuang (approved):

timmc-edx (approved):

provision-lms.sh
```diff
@@ -57,9 +57,9 @@ for user in honor audit verified staff ; do
   email="$user@example.com"
   # Set staff flag for staff user
   if [[ $user == "staff" ]] ; then
-    docker-compose exec -T lms bash -e -c "source /edx/app/edxapp/edxapp_env && python /edx/app/edxapp/edx-platform/manage.py lms --settings=devstack_docker --service-variant lms manage_user $user $email --initial-password-hash \'$demo_hashed_password\' --staff"
+    docker-compose exec -T lms bash -e -c "source /edx/app/edxapp/edxapp_env && python /edx/app/edxapp/edx-platform/manage.py lms --settings=devstack_docker --service-variant lms manage_user $user $email --initial-password-hash '$demo_hashed_password' --staff"
```
timmc-edx, line 60: There are a bunch of script injection possibilities here (before and after, and unrelated). If we wanted to be safe, we could use `printf` with `%q` -- but it's probably not worth fixing in this context.

PR #1025, opened 2023-02-22
docs: Clean up documentation.

dianakhuang:
- Removed out of date known issues.
- Cleaned up workflow commands.
- Removed references to out of date Pycharm instructions.

----

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -365,15 +365,6 @@ Some common service combinations include:
 .. _analyticsapi: https://github.com/openedx/edx-analytics-data-api
 
 
-Known Issues
-------------
-
-Currently, some containers rely on Elasticsearch 7 and some rely on Elasticsearch 1.5. This is
-because services are in the process of being upgraded to Elasticsearch 7, but not all of them
-support Elasticsearch 7 yet. As we complete these migrations, we will update the dependencies
-of these containers.
-
-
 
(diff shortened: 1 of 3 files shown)


rgraber (commented):
Why are we replacing the `lms-*` commands with `dev.*.lms` ? I usually use `lms-*`

dianakhuang:
Mostly this is the better-supported convention? But I don't have strong feelings about it and can go either way.

rgraber:
> Mostly this is the better-supported convention? But I don't have strong feelings about it and can go either way.

Actually this also brings up a question for me about why we have both and if that's more confusing than helpful. But I guess it makes sense to use dev.action.service since that's how it's described in the makefile.

dianakhuang:
I'm not sure why we introduced a new one, but I feel like the `lms-` commands were kept around for backwards compatibility.

rgraber (approved):
I know we're looking for quick wins, so approving, but in the future it might be useful to have examples of what kinds of changes require a full restart

rgraber:
> I know we're looking for quick wins, so approving, but in the future it might be useful to have examples of what kinds of changes require a full restart

Somehow I completely skipped over the examples that were already there. Made a request to add a line about how runserver occasionally just stops picking up changes because it doesn't feel like it.

PR #1038, opened 2023-03-21
fix: rename leftover `dev.down` references to `dev.remove-containers`

justinhynes:
Recently, as part of PR #1021, we had renamed the `dev.down` command to `dev.remove-containers` in order to try and make it a bit clearer what the command is doing behind the scenes. We still have a few commands that are trying to use the `dev.down` variant of the command. This PR updates a few remaining uses of `dev.down` to use the new syntax.

----

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -495,9 +495,9 @@ dev.static.%: ## Rebuild static assets for the specified service's container.
 ########################################################################################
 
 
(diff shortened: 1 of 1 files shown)


rgraber (approved):
LGTM. Nice catch.

PR #1040, opened 2023-03-22
fix: add `redis` as a service dependency for the course-discovery service

justinhynes:
The Discovery service uses redis but it is not currently listed as a service dependency in our `docker-compose.yml` file. This results in errors when adding/updating/removing data in Devstack's version of Discovery (e.g. adding a new organization).

Without this change, the Redis service will not start alongside the Discovery service, and this can result in errors like the following happening when using Discovery:
image

----

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -336,6 +336,7 @@ services:
       - memcached
       - mysql57
       - opensearch12
+      - redis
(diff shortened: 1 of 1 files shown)


johnnagro (approved):
⛵

PR #1047, opened 2023-04-06
build: Add checks for Makefile and docs

timmc-edx:
Also remove duplicate target from `.PHONY` list.

----


rgraber (approved):
LGTM

PR #1048, opened 2023-04-06
fix: Run CMS migrations on CMS container

timmc-edx:
I think this was just a copy-paste error. It worked OK most of the time, but it failed for me recently. I had only run `make requirements` in `make studio-shell` and then tried to `make studio-migrate`. It failed because the management command ran on lms and the migration depended on newer requirements.

----


rgraber (approved):
LGTM

PR #1056, opened 2023-04-21
fix: provisioning tests

robrap:
The provisioning tests were failing with the following error:
> The following packages have unmet dependencies:
>  docker-ce : Depends: containerd.io (>= 1.6.4)

 This change attempts to fix the error by installing
 containerd.io.

----


johnnagro (approved):
⛵

PR #1080, opened 2023-06-01
docs: No longer recommend that devs use devstack in a virtualenv

timmc-edx:
The only thing that most developers use from the devstack requirements file
is docker-compose, and that's already provided by a Docker installation on
some operating systems. So just change the Docker installation instructions
to mention docker-compose, and then drop virtualenvs from the getting
started and main workflow docs.

This commit does not change `docs/developing_on_named_release_branches.rst`
as that file is for an uncommon use-case. Removing virtualenv instructions
from that file would require getting multi-devstack set up in the first
place and then removing all the virtualenv referenves from the scripts that
are laid out in the doc. If it's worth doing that, it's best done by
someone who already uses those instructions.

Issue: https://github.com/openedx/devstack/issues/1076

----

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -52,8 +52,6 @@ You should run all ``make`` commands described below on your local machinge, *no
 from within a Virtual Machine, as these commands are meant to stand up a VM-like environment using
 Docker containers.
 
-For basic development purposes, you will not need a virtualenv. (You will only need to set up a virtualenv and run ``make requirements`` if you are working on devstack itself.)
-
 Directions to setup devstack
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -8,10 +8,12 @@ You will need to have the following installed:
 
 - make
 - Python 3.8
-- Docker
+- Docker, including ``docker-compose``
 
 This project requires **Docker 17.06+ CE**.  We recommend Docker Stable, but
-Docker Edge should work as well.
+Docker Edge should work as well. Ensure that your Docker installation includes
+``docker-compose``; on some operating systems (e.g. Ubuntu Linux) this may require
+a separate package.
 
 **NOTE:** Switching between Docker Stable and Docker Edge will remove all images and
 settings.  Don't forget to restore your memory setting and be prepared to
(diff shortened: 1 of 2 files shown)


docs/developing_on_named_release_branches.rst
```diff
@@ -23,7 +23,6 @@ How do I run multiple named Open edX releases on same machine?
 You can have multiple isolated Devstacks provisioned on a single computer now. Follow these directions **after installing at least two devstacks** to switch between them.
 
 #. If you haven't done so, follow the steps in :doc:`getting_started`, to install the master devstack or any other named release. We recommend that you have at least one devstack on the master branch.
-#. Change directory to your devstack and activate the virtual env.
```
timmc-edx: This was the only thing I really felt comfortable changing in this file. A large chunk of the file is dedicated to virtualenv management and it will take some time to figure out whether it should all be deleted, or whether it needs updating, or what.

timmc-edx:
I tested this by installing the `docker-compose` package on Ubuntu LTS, which currently has the same version as the one in the compiled requirements files. Then I deactivated and deleted my virtualenv and used dev.destroy and did a full reprovision, which worked fine. (Probably about an hour for pull + provision, in total.)

docs/getting_started.rst
```diff
@@ -50,9 +52,7 @@ You should run all ``make`` commands described below on your local machinge, *no
 from within a Virtual Machine, as these commands are meant to stand up a VM-like environment using
 Docker containers.
 
-However, you may want to run the ``make`` commands from within a Python 3 virtual
-environment. This will keep the Python packages required for Devstack separate from
-the ones installed globally on your system.
+For basic development purposes, you will not need a virtualenv. (You will only need to set up a virtualenv and run ``make requirements`` if you are working on devstack itself.)
```
robrap: I wonder if we should clarify that you don't need a virtualenv for the devstack directory and Makefile specifically. You may still use them for repos that are cloned as part of devstack. I'm not sure the best way to word what I am trying to say. :)

robrap: Also, "if you are working on devstack itself" is not clear. It sounds too similar to what I might say when simply using it. Maybe there is another way of saying this, or providing an example of something you might be improving?

timmc-edx: Maybe we want to just drop this paragraph entirely.

robrap: Sounds good. Hopefully this will already be obvious, or we could have a separate doc on this if we knew how to title it correctly. ;)

timmc-edx: Paragraph removed. Anything else to address?

robrap (approved):

PR #1087, opened 2023-06-07
chore: Add GitHub issue template for reporting devstack bugs.

dianakhuang:

----

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug-Report.yml
+++ b/.github/ISSUE_TEMPLATE/Bug-Report.yml
@@ -1,0 +1,42 @@
+name: Bug Report
+description: File a bug report
+title: "[Bug]: "
+labels: ["bug"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Please check the [devstack troubleshooting guide](https://edx.readthedocs.io/projects/open-edx-devstack/en/latest/troubleshoot_general_tips.html) and the [existing list of blocking bugs](https://github.com/openedx/devstack/labels/blocker) before filing a new issue.
+  - type: textarea
+    id: bug-report
+    attributes:
+      label: Describe the bug that you are seeing.
+    validations:
+      required: true
+  - type: input
+    id: container
+    attributes:
+      label: Did this happen on the host (your machine or the remote instance) or in the container?
+      description: e.g. Did this happen outside of running `make dev.shell.<service>` or inside running `make dev.shell.<service>`?
+    validations:
+      required: true
+  - type: textarea
+    id: reproduction-steps
+    attributes:
+      label: Steps to reproduce.
+      description: Do you have a way to replicate what you're seeing?
+    validations:
+      required: false
+  - type: dropdown
+    id: mac-type
+    attributes:
+      label: What system was this issue seen on?
+      description: What type of OS/hardware was devstack running on when you observed it?
+      options:
+        - Apple Silicon
+        - Apple Intel
+        - Hosted Devstack
+        - Linux
+        - Other
+    validations:
+      required: true


.github/ISSUE_TEMPLATE/Bug-Report.yml
```diff
@@ -0,0 +1,46 @@
...
+  - type: textarea
+    id: bug-report
+    attributes:
+      label: Describe the bug that you are seeing.
+    validations:
+      required: true
+  - type: input
+    id: container
+    attributes:
+      label: Did this happen on your machine or in the container?
```
robrap: What do you think of the following?
```suggestion
      label: Did this happen on the host (your machine or the remote instance), or in the container?
```

.github/ISSUE_TEMPLATE/Bug-Report.yml
```diff
@@ -0,0 +1,46 @@
...
+      description: e.g. Did this happen outside of running `make dev.shell.<service>` or inside running `make dev.shell.<service>`?
+    validations:
+      required: true
+  - type: textarea
+    id: reproduction-steps
+    attributes:
+      label: Steps to reproduce.
+      description: How could someone replicate what you're seeing?
+    validations:
+      required: true
```
robrap: Often we don't know. Either this could be optional, or we should have "I don't know." as an acceptable answer in the description?

.github/ISSUE_TEMPLATE/Bug-Report.yml
```diff
@@ -0,0 +1,46 @@
...
+  - type: dropdown
+    id: mac-type
+    attributes:
+      label: Apple Silicon Mac?
+      description: Are you running into this issue on an Apple Silicon Mac?
```
robrap: 1. Can we ask questions that work whether someone is self-reporting an issue, or if we (or someone) is adding an issue after the fact that more than one person may have seen? For example:
```suggestion
      description: Was this issue seen on an Apple Silicon Mac?
```
2. Should we **also** ask whether the issue was seen on an Apple Intel, or any non-Apple Silicon machine? Note that we may end up adding this data later, or we may have multiple cases by the time the ticket is created.

.github/ISSUE_TEMPLATE/Bug-Report.yml
```diff
@@ -0,0 +1,46 @@
...
+      options:
+        - Yes
+        - No
+        - Unknown
+    validations:
+      required: true
+  - type: dropdown
+    id: local-or-hosted
+    attributes:
+      label: Are you seeing this on hosted devstack or local devstack?
```
robrap: If we use the "Was this issue seen on..." language and this gets split into two questions, it would enable us to know whether or not we've ruled out hosted vs local. Otherwise, the issue might be filed under "hosted", but also be possible under "local".

.github/ISSUE_TEMPLATE/Bug-Report.yml
dianakhuang: Yeah, I think we can get away with a more nuanced answer with dropdowns. I'll think about how to structure this some more.

.github/ISSUE_TEMPLATE/Bug-Report.yml
```diff
@@ -0,0 +1,41 @@
...
+    id: reproduction-steps
+    attributes:
+      label: Steps to reproduce.
+      description: Do you have a way to replicate what you're seeing?
+    validations:
+      required: false
+  - type: dropdown
+    id: mac-type
+    attributes:
+      label: What system was this issue seen on?
```
robrap, line 33: Can we dump fixed text on the final PR that may not even be displayed while filling out the initial template. I'd like to see some form of the following (from the original ticket):

```
Follow-up checklist:
- [ ] Is the issue flaky or consistent?
- [ ] Does it affect multiple people or multiple types of systems?
- [ ] Update the devstack troubleshooting documentation page if necessary
  - [ ] Do we need a new troubleshooting section?
  - [ ] Did a troubleshooting section already exist, but it wasn't easy to find given the symptoms?
  - [ ] If a recurring issue, should we ticket an automated resolution in place of the doc?
```

This is also to help with the fact that seeing an issue on Apple Silicon (or any system) isn't helpful if we don't want to collect separate tickets for each instance, but we do want to collect details about other systems that this may be seen on.

robrap (commented):
Looks good. Hoping we can get something like my request in, but happy to iterate if you want that on a separate PR. Thoughts?

.github/ISSUE_TEMPLATE/Bug-Report.yml
dianakhuang: I was thinking about how to handle the follow-up checklist. It doesn't seem like there's a way to insert it into the text without asking the reporter about it, so I was thinking about looking into a GitHub action to add a comment with the follow-up checklist on bug reports.

robrap: Ok. Can you comment on the ticket about this idea? Also, I adjusted the checklist above and questions, so maybe we could copy it back into that ticket?

robrap (approved):

PR #1089, opened 2023-06-08
chore: Add comment to bug reports.

dianakhuang:
This workflow adds a follow-up checklist for arch-bom developers to use when triaging bugs.

----


dianakhuang:
Example comment: https://github.com/dianakhuang/devstack/issues/2#issuecomment-1583290082

robrap (approved):
That's great. I don't really know anything about github workflows, so if you want a real review on that front, you'd need to ask anyone else. :)  We can easily adjust questions and format of the comment in the future if we wish, so best just to get this going. Thank you!

PR #1092, opened 2023-06-12
docs: Add link to new bug reporting template.

dianakhuang:

----

### DIFF
--- a/docs/troubleshoot_general_tips.rst
+++ b/docs/troubleshoot_general_tips.rst
@@ -1,7 +1,10 @@
 Troubleshooting: Common Issues
-==============================
+##############################
 
 .. contents:: Table of Contents
+
+Known Issues
+============
 
 File ownership change
 ---------------------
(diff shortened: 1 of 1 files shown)


docs/troubleshoot_general_tips.rst
```diff
@@ -238,3 +238,11 @@ to shell into the LMS (``make lms-shell`` in devstack) and run ``npm ci``, follo
 See `the github issue`_ to follow the work being done on the resolution.
 
 .. _the github issue: https://github.com/openedx/devstack/issues/1072
+
+
+Your issue isn't on this list?
```
robrap: Problem: I don't think this stands out enough in the TOC.

Solutions:
Option 1: Maybe move this first so it stands out a little bit more, or have a note at the top that points to this section? Not sure if this actually solves it.
Option 2: Adjust headings in the TOC and have an "Issues" heading (or whatever you call it), with all the other items under it, and have this as a second top-level item that stands out. 
Option 3: Other?

docs/troubleshoot_general_tips.rst
```diff
@@ -238,3 +238,11 @@ to shell into the LMS (``make lms-shell`` in devstack) and run ``npm ci``, follo
...
+---------------------------------
+
+Please file `a bug report`_ with any information that could help us debug it.
```
robrap: Nit: Maybe add a note that the bug report template will also help someone identify if a bug report already exists that can be commented on, unless you want to pull that in directly, or hope they just see that in the template.

robrap (commented):
Thanks.

docs/troubleshoot_general_tips.rst
dianakhuang: I think it should be at the bottom because that's where I'd see it if I was scrolling through the list looking for my particular problem. I could try option 2, which does make sense to me.

robrap (approved):
Looks great. Thanks for doing that.

PR #1094, opened 2023-06-12
feat: Updated devstack mysql 5.7 to 8

farhanumar:
Updated mysql 5.7 to mysql 8 for new dev stack setup

----

### DIFF
--- a/provision-lms.sh
+++ b/provision-lms.sh
@@ -29,7 +29,7 @@ done
 # docker-compose exec -T  lms bash -e -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && paver update_db --settings devstack_docker'
 
 echo -e "Change MySQL Host to mysql80 for ${name}"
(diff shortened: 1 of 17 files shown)


provision-lms.sh
```diff
@@ -28,6 +28,9 @@ done
 # Run edxapp migrations first since they are needed for the service users and OAuth clients
 # docker-compose exec -T  lms bash -e -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && paver update_db --settings devstack_docker'
 
+echo -e "Change MySQL Host to mysql80"
+docker-compose exec -T lms  bash -e -c 'find /edx/ -type f -exec sed -i 's/mysql57/mysql80/g' {} +'
```
matthugs, line 32: this quoting scheme looks wrong. Do we want single-quotes for the sed command?

matthugs: also more generally why are we attempting to find-and-replace here at provision-time rather than in source code?

farhanumar: @matthugs This was just to verify that provisioning test work. This PR will not merge. That was work around to verify that images need updates for mysql hosts for the provisioning test to pass. The last commit in this PR passed the tests.

PR #1098, opened 2023-06-15
chore: credentials pointing to mysql8 container

farhanumar:
----
Pointing of credentials to mysql8 container.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -309,13 +309,13 @@ services:
     depends_on:
       - lms
       - memcached
-      - mysql57
+      - mysql80
     # Allows attachment to the credentials service using 'docker attach <containerID>'.
     stdin_open: true
     tty: true
     environment:
       CACHE_LOCATION: edx.devstack.memcached:11211
-      DB_HOST: edx.devstack.mysql57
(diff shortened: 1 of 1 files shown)


adzuci:
@farhanumar I see many tests are failing for this PR, could you sync with Nadeem to see if we may be able to resolve some of the failures?

I think it may also be worth working with @justinhynes to ask what data devstack users have in the mysql5.7 containers and create a make command to copy that data into mysql 8 for when this PR is eventually moved.  I think it may also be worth not merging this until we can do all services at once, adding a change log entry and possibly making some check to detect when devstack is launched for the first time on mysql8 so that we can make the migration smoother for devs.

farhanumar:
@adzuci I had investigated these failures and fixed them here https://github.com/openedx/devstack/pull/1094, mysql57 has been hard-coded in some of the container images.  I left it because this was something teams should fix it while they will be testing apps with mysql 8. The fix is easy see matt comment in https://github.com/openedx/devstack/pull/1094, mysql57 , its a string replacement in running container.

adzuci:
That makes sense to me, could you update the PR description to include a note about that and describe the next steps for moving this work forward?

PR #1101, opened 2023-06-15
feat: updated the node version for MFEs

abdullahwaheed:
----


rgraber (approved):
No problem with the code, but I would call this a feat: and not a refactor:

PR #1103, opened 2023-06-15
feat: Remove e2e-tests (including the whitelabel tests that used it)

timmc-edx:
The e2e tests repo is deprecated and is no longer part of Open edX.

Backported from commit 429ff222cc5ea7cf1d0c2555fe8fdcdc795a59ed.

----


timmc-edx:
These failures seem like they might be related to getting the wrong version of DB snapshots or images used in the CI checks. Not looking into it since Olive is now out of support.

PR #1116, opened 2023-07-06
fix: use oracle mysql80 image in devstack

rgraber:
Oracle's mysql80 image has arm64 builds. Tested on both Apple Silicon (M2) and Intel. Successfully brought up container on both.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -259,7 +259,7 @@ services:
     environment:
       MYSQL_ROOT_PASSWORD: ""
       MYSQL_ALLOW_EMPTY_PASSWORD: "yes"
-    image: mysql:8.0-oracle
+    image: mysql:8.0.33-oracle
     networks:
       default:
         aliases:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -253,6 +253,7 @@ services:
       - mysql57_data:/var/lib/mysql
 
   mysql80:
+    platform: linux/amd64
     command: mysqld --character-set-server=utf8 --collation-server=utf8_general_ci
     container_name: "edx.${COMPOSE_PROJECT_NAME:-devstack}.mysql80"
     hostname: mysql80.devstack.edx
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -253,6 +253,7 @@ services:
       - mysql57_data:/var/lib/mysql
 
   mysql80:
+    platform: linux/amd64
```
timmc-edx: From what I'm seeing [in the docker image's repo](https://github.com/docker-library/mysql/issues/778), there may be a performance hit to this (due to the emulation), and even some stability concerns with qemu. So it would make it work at all on Apple Silicon, but slower for other Mac users. Tough tradeoff. But there's also a series of tags using mysql package built by Oracle. Can you try the `8.0.33-oracle` image instead, leaving platform unspecified?

timmc-edx: (You should be able to test the other image by removing the container, removing the `platform` line, changing the image line, and then re-upping -- it should keep your data, no need to re-provision.)

rgraber: That worked on both Silicon and Intel, thanks!

docker-compose.yml
```diff
@@ -259,7 +259,7 @@ services:
     environment:
       MYSQL_ROOT_PASSWORD: ""
       MYSQL_ALLOW_EMPTY_PASSWORD: "yes"
-    image: mysql:8.0.26
+    image: mysql:8.0.33-oracle
```
timmc-edx, line 264:
```suggestion
    # Oracle-packaged version includes an `linux/arm64/v8` version, needed for
    # machines with Apple Silicon CPUs (Mac M1, M2)
    image: mysql:8.0.33-oracle
```

timmc-edx (approved):
Would recommend a comment of some sort (left a suggestion) but otherwise good to go.

docker-compose.yml
```diff
@@ -259,7 +259,9 @@ services:
     environment:
       MYSQL_ROOT_PASSWORD: ""
       MYSQL_ALLOW_EMPTY_PASSWORD: "yes"
-    image: mysql:8.0.26
+    # Oracle-packaged version includes an `linux/arm64/v8` version, needed for
```
timmc-edx: I just noticed that I wrote `an` instead of `a`. :weary:

rgraber: Fixed : P

PR #1118, opened 2023-07-11
chore: Remove reference to paver update_db.

dianakhuang:
We are planning on deprecating this method,
and this piece of code has been commented out
for a while anyway.

### DIFF
--- a/provision-lms.sh
+++ b/provision-lms.sh
@@ -26,8 +26,6 @@ for app in "${apps[@]}"; do
 done
 
 # Run edxapp migrations first since they are needed for the service users and OAuth clients
-# docker-compose exec -T  lms bash -e -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && paver update_db --settings devstack_docker'
-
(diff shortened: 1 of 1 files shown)


davidjoy (approved):

provision-lms.sh
```diff
@@ -26,7 +26,6 @@ for app in "${apps[@]}"; do
 done
 
 # Run edxapp migrations first since they are needed for the service users and OAuth clients
```
robrap, line 28: @dianakhuang: Why did you leave in the comment? Also, did you do any research about why this was here in the first place and why it was commented out?

robrap (changes requested):

provision-lms.sh
dianakhuang: I left in the comment because it's still running migrations. We're just not using `paver update_db` to run them.

robrap: Oops. I didn't realize those were migration commands.
1. Nit: Do you mind removing the newline between? That's just personal preference, so ok if you leave it.
2. Is this another place we could switch to the edx-platform Makefile migrate calls, and again there is the `showmigrations` question from https://github.com/openedx/edx-platform/pull/32705#issuecomment-1632843276.

robrap (approved):

provision-lms.sh
dianakhuang: Yeah, I have the new migrate commands run `showmigrations` as well, so I think replacing these with `make migrate` will be effectively the same.

PR #1140, opened 2023-07-31
fix: Use correct name for --detach param

timmc-edx:
`-d` is valid and so is `--detach`. But the `--d` that was used here (with two hyphens) only worked on some machines. I believe that was autocorrected to `--detach` on those machines because it was a unique prefix, as I was also able to use `--det`.

This might be a matter of versions, or of `docker-compose` vs. `docker compose` shims, or just difference by OS. But in any case we should use a documented version of the flag.

----


johnnagro (approved):

timmc-edx:
Switched to `-d` for consistency. Cancelling provisioning tests because they're expensive and don't exercise this target.

johnnagro (approved):

PR #1141, opened 2023-07-31
docs: More troubleshooting tips (starting from scratch; broken outline)

timmc-edx:
----

### DIFF
--- a/docs/troubleshoot_general_tips.rst
+++ b/docs/troubleshoot_general_tips.rst
@@ -250,7 +250,8 @@ Starting From Scratch
 
 If you think your devstack is broken beyond repair, you can start from scratch using ``make dev.destroy``, followed by the :doc:`getting_started` instructions.
 
-If you want to make absolutely sure that there are no lingering data volumes after the ``dev.destroy`` step, run ``docker volume ls --quiet | grep devstack`` -- if you see surviving devstack volumes that are currently mentioned in docker-compose.yml, there may be a bug.
+If you want to make absolutely sure that there are no lingering data volumes after the ``dev.destroy`` step, run ``docker volume ls --quiet | grep devstack`` -- if you see surviving devstack volumes that are currently mentioned in docker-compose.yml, there may be a bug. If you can reproduce the issue reliably, consider `reporting an issue <Reporting New Issues_>`_.
+
 
 Reporting New Issues
 ====================
--- a/docs/troubleshoot_general_tips.rst
+++ b/docs/troubleshoot_general_tips.rst
@@ -218,7 +218,7 @@ Another error you may get if the code and the image are out of sync is sql or Dj
 
 Demo course is empty in studio
 ------------------------------
-After provisioning and opening Studio, you may see an empty outline for the demo course. This usually means there is a disconnect between the block ids in mySQL and the corresponding data in Mongo.
+After provisioning and opening Studio, you may see an empty outline for the demo course. This usually means there is a disconnect between the block ids in mySQL and the corresponding data in Mongo. See https://github.com/openedx/devstack/issues/1073 for the tracking issue.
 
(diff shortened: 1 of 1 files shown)


timmc-edx:
(I've cancelled the provisioning tests for this doc-only PR.)

docs/troubleshoot_general_tips.rst
```diff
@@ -218,7 +218,7 @@ Another error you may get if the code and the image are out of sync is sql or Dj
 
 Demo course is empty in studio
 ------------------------------
-After provisioning and opening Studio, you may see an empty outline for the demo course. This usually means there is a disconnect between the block ids in mySQL and the corresponding data in Mongo.
+After provisioning and opening Studio, you may see an empty outline for the demo course. This usually means there is a disconnect between the block ids in mySQL and the corresponding data in Mongo. See https://github.com/openedx/devstack/issues/1073 for the tracking issue.
```
robrap: 1. I'd move this to a new line after 223 (that details the fix).
2. In 223, consider updating to "To fix locally...", or any other language that notes that this is a workaround fix.
3. Consider adding something to clarify the point of the issue you are point to, for example, if it is the issue for a more permanent fix that would avoid this workaround?

docs/troubleshoot_general_tips.rst
```diff
@@ -243,6 +243,13 @@ See `the github issue`_ to follow the work being done on the resolution.
 .. _the github issue: https://github.com/openedx/devstack/issues/1072
 
 
+Starting From Scratch
+=====================
+
+If you think your devstack is broken beyond repair, you can start from scratch using ``make dev.destroy``, followed by the Getting Started instructions.
```
robrap: Nit: link to Getting Started instructions?

docs/troubleshoot_general_tips.rst
```diff
@@ -243,6 +243,13 @@ See `the github issue`_ to follow the work being done on the resolution.
...
+
+If you want to make absolutely sure that there are no lingering data volumes after the ``dev.destroy`` step, run ``docker volume ls --quiet | grep devstack`` -- if you see surviving devstack volumes that are currently mentioned in docker-compose.yml, there may be a bug.
```
robrap: Can you clarify what "there may be a bug" means, and what someone should do?

robrap (commented):
Thanks.

docs/troubleshoot_general_tips.rst
timmc-edx: I will admit I thought of this but then didn't want to figure out the exact rST syntax required. :upside_down_face:

docs/troubleshoot_general_tips.rst
timmc-edx: Adjusted this section.

docs/troubleshoot_general_tips.rst
timmc-edx: I left it vague on purpose. My main goal was to convey "if you see this, something may be abnormal". I'm reticent to encourage people to submit bug reports that are lacking information, but also don't want to overwhelm the section with anything too specific. Suggestions on phrasing? Maybe something like this:

```suggestion
If you want to make absolutely sure that there are no lingering data volumes after the ``dev.destroy`` step, run ``docker volume ls --quiet | grep devstack`` -- if you see surviving devstack volumes that are currently mentioned in docker-compose.yml, there may be a bug. If you can reproduce the issue reliably, consider `reporting an issue <Reporting New Issues_>`_.
```

(except that links to the next section, so it's actually very confusing to click that link...)

robrap: This seems fine to me.

robrap (approved):
If you could commit the change in the comment, that would be great. Thanks!

PR #1142, opened 2023-08-01
feat: Remove redundant `seed_permissions_roles` call for demo course

timmc-edx:
The `import` management command (from cms contentstore) is called right before the `seed_permissions_roles` command (from lms discussions). However, the import command already performs the seeding.

See https://github.com/openedx/devstack/issues/1129 for archaeological notes.

Confirmation that the call doesn't add anything:

- Provisioned LMS with `DEVSTACK_SKIP_DEMO=true`
- Manually ran first two commands for importing demo course (git clone, `import` management command)
- Took MySQL DB dumps
- Ran final demo course command, `seed_permissions_roles`
- Took new DB dumps
- Diff of SQL files showed only a change in the dump dates

----

  - None


connorhaugh (approved):
LGTM

PR #1148, opened 2023-08-09
docs: Add copy-node-modules.sh note to troubleshooting page

timmc-edx:
This introduces a "past problems" section, which we're looking at adding to in the future. That work will include details to help people do the "update and pull" step mentioned here.

- [ ] Merge first: https://github.com/openedx/edx-platform/pull/32946

----


rgraber (approved):
LGTM

PR #1165, opened 2023-08-31
fix: updated lms and cms mysql dependency to 8.0

zubairshakoorarbisoft:
Updated MySql dependency to 8.0 for LMS and CMS


UsamaSadiq (approved):
I've tested the change locally on my end and have verified that data migrations work successfully and mysql80 works on devstack as expected.

PR #1173, opened 2023-09-12
build: Don't run provisioning tests on docs-only PRs and pushes

timmc-edx:
NB: This approach will only works so long as the provisioning tests are not required.

Also, drop the `branches: **` directive, which looks like a no-op.

- See https://github.com/openedx/devstack/pull/1174 for an example of a docs-only PR with this change in effect.
- https://github.com/openedx/devstack/pull/1175 shows a mix of doc and non-doc changes

----


rgraber (approved):
LGTM

PR #1178, opened 2023-09-13
Upgrade neo4j image in devstack

rgraber:
----

Upgrade neo4j image used for coursegraph to version that supports Apple Silicon. Requires changing the default password since it now has a minimum of 8 chars. Tested locally on an M2. Was able to provision and query.


kenclary (approved):
ok

PR #1181, opened 2023-09-19
docs: Add note about fixed wget error

timmc-edx:
----

### DIFF
--- a/docs/troubleshoot_general_tips.rst
+++ b/docs/troubleshoot_general_tips.rst
@@ -282,7 +282,7 @@ Cannot run ``make upgrade`` in lms shell due to missing wget
     /bin/sh: 1: wget: not found
     make[1]: *** [Makefile:115: requirements/common_constraints.txt] Error 127
 
-This error was `introduced <https://github.com/openedx/edx-platform/pull/33271>`_ and `resolved <https://github.com/openedx/edx-platform/pull/33288>`_ in September 2023.
+This error was `introduced <https://github.com/openedx/edx-platform/pull/33271>`_ and `resolved <https://github.com/openedx/edx-platform/pull/33288>`_ in September 2023. While this can be solved by updating your devstack, you can also run ``apt update; apt install wget`` from lms-shell to resolve this temporarily.
 
 .. _update your repos and pull the latest images:
 
--- a/docs/troubleshoot_general_tips.rst
+++ b/docs/troubleshoot_general_tips.rst
@@ -274,7 +274,7 @@ During ``make requirements`` there would be an error::
 This was resolved in July 2023 with https://github.com/openedx/edx-platform/pull/32732.
 
 Cannot run ``make upgrade`` in lms shell due to missing wget
-----------------------------------------------------------
+------------------------------------------------------------
 
 ``make upgrade`` or ``make compile-requirements`` in lms-shell would produce an error about wget::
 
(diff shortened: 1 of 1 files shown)


docs/troubleshoot_general_tips.rst
```diff
@@ -273,6 +273,17 @@ During ``make requirements`` there would be an error::
...
+Cannot run ``make upgrade`` in lms shell due to missing wget
+------------------------------------------------------------
+
+``make upgrade`` or ``make compile-requirements`` in lms-shell would produce an error about wget::
+
+    wget -O "requirements/common_constraints.txt" https://raw.githubusercontent.com/edx/edx-lint/master/edx_lint/files/common_constraints.txt
+    /bin/sh: 1: wget: not found
+    make[1]: *** [Makefile:115: requirements/common_constraints.txt] Error 127
+
+This error was `introduced <https://github.com/openedx/edx-platform/pull/33271>`_ and `resolved <https://github.com/openedx/edx-platform/pull/33288>`_ in September 2023.
```
robrap: I think it makes sense to leave a brief note on how to resolve. In Slack you had written:
> you'll need to make lms-pull to get the latest image. Or, you can instead run apt update; apt install wget in lms-shell if you want to patch it quickly. :-)

docs/troubleshoot_general_tips.rst
```diff
@@ -273,6 +273,17 @@ During ``make requirements`` there would be an error::
...
+Cannot run ``make upgrade`` in lms shell due to missing wget
+------------------------------------------------------------
+
+``make upgrade`` or ``make compile-requirements`` in lms-shell would produce an error about wget::
+
+    wget -O "requirements/common_constraints.txt" https://raw.githubusercontent.com/edx/edx-lint/master/edx_lint/files/common_constraints.txt
+    /bin/sh: 1: wget: not found
+    make[1]: *** [Makefile:115: requirements/common_constraints.txt] Error 127
+
+This error was `introduced <https://github.com/openedx/edx-platform/pull/33271>`_ and `resolved <https://github.com/openedx/edx-platform/pull/33288>`_ in September 2023. While this can be solved by updating your devstack, you can also run ``apt update; apt install wget`` from lms-shell to resolve this temporarily.
```
robrap, line 285: Some problems will be fixed with `git pull` in the devstack directory. Others, by updating a specific container, like in this case. Is there a reason why you didn't want to stick with `make lms-pull`, and instead went for the more general "updating your devstack" language?

In case it isn't clear, I prefer the more specific instructions. 🙂

timmc-edx: My understanding was that this section was all going to be past problems that all have the same solution: Follow the instructions in the "Updating Devstack" section.

robrap: I guess I just prefer saving people time. I don't think there is a hard and fast rule - so do whatever you think is best.

robrap (approved):

docs/troubleshoot_general_tips.rst
timmc-edx: I think it's tricky to give a solution along those lines that doesn't cause more problems. For example, if this is the _only_ problem someone is encountering, and we tell them to run `make lms-pull`, then they might have new problems due to repo/image mismatch.

PR #1190, opened 2023-09-28
test: Fix provisioning tests by ignoring maybe-missing google-cloud-sdk

timmc-edx:
We're seeing intermittent provisioning failures with this message:

```
No apt package "google-cloud-sdk", but there is a snap with that name.
Try "snap install google-cloud-sdk"

E: Unable to locate package google-cloud-sdk
Error: Process completed with exit code 100.
```

I couldn't figure out why, but all we care about is that it's gone, so I'm adding a wildcard. (It is also unlikely that there are any packages that we *do* want that have that prefix.)

----

### DIFF
--- a/.github/workflows/provisioning-tests.yml
+++ b/.github/workflows/provisioning-tests.yml
@@ -51,7 +51,10 @@ jobs:
           docker compose --version
 
       - name: free up disk space
-        run:  sudo apt remove --purge -y ghc-* azure-cli google-cloud-sdk hhvm llvm-* dotnet-* powershell mono-* php* ruby*
+        # 2023-09-28: google-cloud-sdk removed from this list because it was intermittently
+        #   unavailable as an apt package to remove, and might be migrating to snap. If more
+        #   disk space is needed, see if the snap is installed, and remove that.
+        run:  sudo apt remove --purge -y ghc-* azure-cli hhvm llvm-* dotnet-* powershell mono-* php* ruby*
 
       - name: set up requirements
         run:  make requirements


robrap (approved):

.github/workflows/provisioning-tests.yml
```diff
@@ -51,7 +51,10 @@ jobs:
           docker compose --version
 
       - name: free up disk space
-        run:  sudo apt remove --purge -y ghc-* azure-cli google-cloud-sdk hhvm llvm-* dotnet-* powershell mono-* php* ruby*
+        # 2023-09-28: google-cloud-sdk removed from this list because it was intermittently
+        #   unavailable as an apt package to remove, and might be migrating to snap. If more
+        #   disk space is needed, see if the snap is installed, and remove that.
+        run:  sudo apt remove --purge -y ghc-* azure-cli hhvm llvm-* dotnet-* powershell mono-* php* ruby*
```
robrap, line 57: @timmc-edx: Is there an alternative that breaks it out on to its own line and swallows errors?

timmc-edx: Yes, but I'm not sure it's worth it unless we see out-of-space issues again. And in that case, I'd first try removing the snap. But I don't have a lot of experience with snap, so this would take more research.

Interestingly, in February it sounded like GitHub might be planning on [not pre-installing this apt package anymore](https://github.com/actions/runner-images/issues/7101#issuecomment-1431174447) partly due to the age of the package; maybe Ubuntu itself is removing the apt package as well. If that's the case, GitHub _probably_ isn't replacing it with the snap package either.

feanil (approved):

timmc-edx:
Still waiting on the CLA check -- GitHub must be behind on webhook-sending. I'm using my repo admin powers to temporarily allow admin-bypassing of that check.

PR #1194, opened 2023-10-03
docs: update troubleshoot_general_tips.rst

Guilleves:
Add fix for blank login/signup pages on localhost:18000

----

### DIFF
--- a/docs/troubleshoot_general_tips.rst
+++ b/docs/troubleshoot_general_tips.rst
@@ -236,6 +236,15 @@ The fix is to get a new auth session. You can do any of the following:
 4. Log in
 5. Navigate back to the MFE
 
+Blank login/signup screens
+--------------------------
+After starting LMS/CMS (and compiling static assets for both), it's possible you run into a blank page when trying to sign in via http://localhost:18000/login. 
+
+This is most likely a cache issue and can be easily fixed in 2 different ways:
+
+1. Hard reloading the page, or
+2. Rebuilding the frontend and giving it another try
+
 Missing vendor file node_modules/backbone.paginator/lib/backbone.paginator.js
 -----------------------------------------------------------------------------
 This message sometimes appears when provisioning. The root cause of this is as yet unknown but the most effective workaround seems to be


docs/troubleshoot_general_tips.rst
```diff
@@ -236,6 +236,15 @@ The fix is to get a new auth session. You can do any of the following:
 4. Log in
 5. Navigate back to the MFE
 
+Blank login/signup screens
+--------------------------
+After starting LMS/CMS (and compiling static assets for both), it's possible you run into a blank page when trying to sign in via http://localhost:18000/login. 
+
+This is most likely a cache issue and can be easily fixed in 2 different ways:
```
robrap, line 243: Feel free to choose your own language, but what do you think of the sentiments I am trying to add here?
```suggestion
If you think that everything is up-to-date, this is most likely a cache issue and might be easily fixed in 2 different ways:
```

docs/troubleshoot_general_tips.rst
```diff
@@ -236,6 +236,15 @@ The fix is to get a new auth session. You can do any of the following:
 4. Log in
 5. Navigate back to the MFE
 
+Blank login/signup screens
+--------------------------
+After starting LMS/CMS (and compiling static assets for both), it's possible you run into a blank page when trying to sign in via http://localhost:18000/login.
```
robrap, line 241: [question] Does this mean this is for the legacy login experience, and **not** using an MFE? Would it help to be explicit, or is the port good enough? I'm not sure if the MFE has a different port?

docs/troubleshoot_general_tips.rst
```diff
@@ -236,6 +236,15 @@ The fix is to get a new auth session. You can do any of the following:
...
 5. Navigate back to the MFE
 
+Blank login/signup screens
+--------------------------
+After starting LMS/CMS (and compiling static assets for both), it's possible you run into a blank page when trying to sign in via http://localhost:18000/login. 
+
+This is most likely a cache issue and can be easily fixed in 2 different ways:
+
+1. Hard reloading the page, or
+2. Rebuilding the frontend and giving it another try
```
robrap, line 246: Can you clarify what "rebuilding the frontend" means?

robrap (commented):
Great. Mostly just some clarifying questions. Thanks.

docs/troubleshoot_general_tips.rst
Guilleves: that is some great wording, thanks!

docs/troubleshoot_general_tips.rst
Guilleves: After discussing this with Marcos (who pointed me to the fix), looks like this is not relevant to this use case. It was intended to say something along the lines of "try recompiling your assets", but it doesn't have anything to do with the actual fix. Will update

PR #1197, opened 2023-10-04
fix: Add mysql80 to list of completeable service names

timmc-edx:
----


timmc-edx:
I've cancelled the provisioning checks, to save work, since they do not exercise this pathway.

johnnagro (approved):

PR #1198, opened 2023-10-04
feat: Add util for waiting for container startup; use in dbcopyall8

timmc-edx:
`make dev.dbcopyall8` doesn't work unless the mysql containers have been started and are ready to run. This is not the only place we need to wait for a functional check on a container, so in this commit I've gone ahead and started a utility for doing so.

This first pass only supports MySQL, but we should at least add Mongo as well, as that's also needed in one of the provisioning scripts.

I've moved the _db_copy8_targets from being an order-only prerequisite to being a recursive make call. I'm not sure why the order-only syntax was in use here, but in any case I needed to add in a script call before those targets were called, so it's likely moot.

Also:

- Drop unneeded `bash -c` wrappers from db copy docker exec calls

----


johnnagro (approved):

timmc-edx:
xqueue provisioning test passed, so I'm cancelling the others.

PR #1199, opened 2023-10-04
feat: Unify wait-ready with check.sh; implement mongo wait

timmc-edx:
- Move MySQL readiness checks from wait-ready.sh to existing check.sh, and call check.sh for any service we want to wait for
- Add Mongo readiness check to check.sh, and use it from provisioning
- Give cms a separate check from lms
- Limit docker log output to 30 lines when check fails

----

### DIFF
--- a/check.sh
+++ b/check.sh
@@ -46,7 +46,7 @@ run_check() {
     if $cmd; then  # Run the command itself and check if it succeeded.
         succeeded="$succeeded $check_name"
     else
-        docker compose logs -n30 "$service"
+        docker compose logs --tail 30 "$service"  # Just show recent logs, not all history
         failed="$failed $check_name"
     fi
     set -e  # Re-enable exit-on-error
--- a/check.sh
+++ b/check.sh
@@ -46,12 +46,35 @@ run_check() {
     if $cmd; then  # Run the command itself and check if it succeeded.
         succeeded="$succeeded $check_name"
     else
-        docker compose logs "$service"
+        docker compose logs -n30 "$service"
         failed="$failed $check_name"
     fi
     set -e  # Re-enable exit-on-error
     echo  # Newline
 }
+
+mysql_run_check() {
+    container_name="$1"
+    mysql_probe="SELECT EXISTS(SELECT 1 FROM mysql.user WHERE user = 'root')"
+    run_check "${container_name}_query" "$container_name" \
+        "docker compose exec -T $(printf %q "$container_name") mysql -uroot -se $(printf %q "$mysql_probe")"
+}
+
+if should_check mysql57; then
+    echo "Checking MySQL 5.7 query endpoint:"
+    mysql_run_check mysql57
+fi
+
+if should_check mysql80; then
+    echo "Checking MySQL 8.0 query endpoint:"
+    mysql_run_check mysql80
+fi
+
+if should_check mongo; then
+    echo "Checking MongoDB status:"
+    run_check mongo_status mongo \
+        "docker compose exec -T mongo mongo --eval \"db.serverStatus()\""
+fi
 
 if should_check registrar; then
     echo "Checking Registrar heartbeat:"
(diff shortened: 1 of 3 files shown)


check.sh
```diff
@@ -46,13 +46,36 @@ run_check() {
     if $cmd; then  # Run the command itself and check if it succeeded.
         succeeded="$succeeded $check_name"
     else
-        docker compose logs "$service"
+        docker compose logs -n30 "$service"
```
robrap: Could you add a code comment that this is limiting to 30 lines and maybe explain why?

provision.sh
```diff
@@ -160,11 +160,7 @@ docker compose exec -T mysql80 bash -e -c "mysql -uroot mysql" < provision-mysql
 if needs_mongo "$to_provision_ordered"; then
 	echo -e "${GREEN}Waiting for MongoDB...${NC}"
 	# mongo container and mongo process/shell inside the container
-	until docker compose exec -T mongo mongo --eval "db.serverStatus()" &> /dev/null
-	do
-	  printf "."
-	  sleep 1
-	done
+        ./wait-ready.sh mongo
```
robrap: What is the indent for?

robrap (commented):
I think it looks good. I'm not certain about how some of the old code works, but since you are just using that code, it seems fine. :)

check.sh
timmc-edx: How about using the long name instead? `--tail 30`

provision.sh
timmc-edx: Hah. So, most of this block is indented with tabs instead of spaces, which I want to clean up in a separate PR, but it turns out my change ended up indented with spaces. (8 spaces, matching Emacs' display of a single tab. But GitHub shows tabs as an indentation of 4.) I'll make sure to change it to a tab for now.

robrap (approved):

timmc-edx:
Hmm, provisioning tests are failing at the wait-ready step... :-/

timmc-edx:
Figured out what was wrong. See commit message for details.

robrap (commented):
I'm not sure what is meant by this part of the commit comment?:
> (The correct way to build up a command and run it is instead with arrays,
such as `"${cmd[@]}"`.)

Is this something that should go into a code comment instead, and would it become more clear?

timmc-edx:
Executing an array would be the "clean" way of doing it, and it's what I would normally do. It has the advantage that it can only represent an executable (the first array element) and its arguments (the rest of the array), and doesn't support redirection, pipelines, etc. -- the reduced functionality means it can be easier to reason about in some cases.

But... passing an array to a function, which is what we would need to do here, is more complicated and annoying, so just passing a precomposed string to a bash subprocess is the more expedient option.

I think it will be clear to a reader what's going on (possibly more clear than before, even) so I don't think it needs a comment about alternative options.

timmc-edx:
Credentials dev.check failed with `ModuleNotFoundError: No module named 'edx_credentials_themes'` which is very weird because you can see that getting installed earlier on in the logs. Rerunning and hoping that's just a weird glitch... :-/

robrap:
> I think it will be clear to a reader what's going on (possibly more clear than before, even) so I don't think it needs a comment about alternative options.

Ok. I'm not sure if you also plan to drop it from the squashed commit message, and if it was just for the review, but it feels disconnected as-written. So, if you don't think it makes sense in a code comment, it probably doesn't make sense in the commit comment either.

timmc-edx:
Yeah, I'm not planning on including it in the squash commit.

Still need to figure out what's going with credentials. Seems like it should be unrelated, but it's failing in the check phase.

timmc-edx:
OK, it passed that time? So I suspect that this is just an existing race condition with `dev.check` that... happened to come up just as I was working with check.sh? A little suspicious, but I _think_ it has come up before.

I'll fix that race condition in a separate PR.

PR #1202, opened 2023-10-06
chore: ignore django_migrations table while taking dump

mumarkhan999:
Ignore `django_migrations` while taking dump as it's causing `operational errors`


mumarkhan999:
Closing this PR as this solution will cause some other issues.

PR #1204, opened 2023-10-09
chore: update dbcopyall8 command

mumarkhan999:
The previous `dbcoopyall8` command causing `operational errors`

The occurrence of the `operational errors: table already exists issue` can be traced to the following:
- During the `provisioning` process, the `mysql80 container` is populated with the `most recent data`, (new migrations/tables).
- When `data is transferred` from `mysql57` using `dbcopyall8` command, the `newly created tables persist` in the mysql80 container. But, the `django_migrations` table becomes `outdated` in this process.
- Consequently, the `records` for any new migration `are removed` from the `django_migrations` table. This leads to a failure when attempting to execute dev.migrate.

This PR is to fix the above-mentioned issue.
It will wipe out all the data of `mysql8` before actually dumping `mysql57` data inside `mysql80` container.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -469,8 +469,8 @@ dev.dbcopyall8: ## Clean mysql80 container and copy data from old mysql 5.7 cont
 	$(MAKE) dev.remove-containers.mysql80
 	docker volume rm devstack_mysql80_data
 	$(MAKE) dev.up.mysql57+mysql80
-	sleep 10
 	$(MAKE) dev.wait-for.mysql57+mysql80
+	sleep 10
 	docker compose exec -T mysql80 mysql -uroot mysql < provision-mysql80.sql
 	$(MAKE) $(_db_copy8_targets)
 	$(MAKE) stop
--- a/Makefile
+++ b/Makefile
@@ -462,21 +462,16 @@ dev.shell.%: ## Run a shell on the specified service's container.
 dev.dbshell:
 	docker compose exec mysql80 bash -c "mysql"
 
-dev.dbsetupmysql8:
-	docker compose exec -T mysql80 bash -e -c "mysql -uroot mysql" < provision-mysql80.sql
-
-dev.dbcleanmysql8:
-	docker container rm edx.devstack.mysql80
-	docker volume rm devstack_mysql80_data
-
 DB_NAMES_LIST = credentials discovery ecommerce notes registrar xqueue edxapp edxapp_csmh dashboard analytics-api reports reports_v1
 _db_copy8_targets = $(addprefix dev.dbcopy8.,$(DB_NAMES_LIST))
 dev.dbcopyall8: ## Clean mysql80 container and copy data from old mysql 5.7 containers into new mysql8 dbs
 	$(MAKE) stop
-	$(MAKE) dev.dbcleanmysql8
+	$(MAKE) dev.remove-containers.mysql80
+	docker volume rm devstack_mysql80_data
 	$(MAKE) dev.up.mysql57+mysql80
+	sleep 10
 	$(MAKE) dev.wait-for.mysql57+mysql80
-	$(MAKE) dev.dbsetupmysql8
+	docker compose exec -T mysql80 mysql -uroot mysql < provision-mysql80.sql
 	$(MAKE) $(_db_copy8_targets)
 	$(MAKE) stop
 
--- a/Makefile
+++ b/Makefile
@@ -462,12 +462,23 @@ dev.shell.%: ## Run a shell on the specified service's container.
 dev.dbshell:
 	docker compose exec mysql80 bash -c "mysql"
 
+dev.dbsetupmysql8:
+	docker compose exec -T mysql80 bash -e -c "mysql -uroot mysql" < provision-mysql80.sql
+
+dev.dbcleanmysql8:
+	docker container rm edx.devstack.mysql80
+	docker volume rm devstack_mysql80_data
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -462,12 +462,23 @@ dev.shell.%: ## Run a shell on the specified service's container.
 dev.dbshell:
 	docker compose exec mysql80 bash -c "mysql"
 
+dev.dbsetupmysql8:
```
timmc-edx: It might make more sense to include these commands directly in `dev.dbcopyall8`. These targets wouldn't be called directly or from other recipes, right?

Makefile
```diff
@@ -462,12 +462,23 @@ dev.shell.%: ## Run a shell on the specified service's container.
...
+	docker compose exec -T mysql80 bash -e -c "mysql -uroot mysql" < provision-mysql80.sql
```
timmc-edx: This should work fine without the `bash -c` indirection:
```suggestion
	docker compose exec -T mysql80 mysql -uroot mysql < provision-mysql80.sql
```

Makefile
```diff
@@ -462,12 +462,23 @@ dev.shell.%: ## Run a shell on the specified service's container.
...
+
+dev.dbcleanmysql8:
+	docker container rm edx.devstack.mysql80
```
timmc-edx: If we use docker compose, we can avoid the prefix:
```suggestion
	docker compose rm mysql80
```

Makefile
mumarkhan999: Yes, these commands wouldn't be called directly or from other recipes. 
Would add them under dbcopy command.

Makefile
timmc-edx: Or I guess we could just use `make mysql80-remove-containers`.

Makefile
```diff
@@ -464,10 +464,16 @@ dev.dbshell:
 
 DB_NAMES_LIST = credentials discovery ecommerce notes registrar xqueue edxapp edxapp_csmh dashboard analytics-api reports reports_v1
 _db_copy8_targets = $(addprefix dev.dbcopy8.,$(DB_NAMES_LIST))
-dev.dbcopyall8: ## Copy data from old mysql 5.7 containers into new mysql8 dbs
+dev.dbcopyall8: ## Clean mysql80 container and copy data from old mysql 5.7 containers into new mysql8 dbs
+	$(MAKE) stop
+	$(MAKE) dev.remove-containers.mysql80
+	docker volume rm devstack_mysql80_data
 	$(MAKE) dev.up.mysql57+mysql80
+	sleep 10
```
timmc-edx: This sleep call shouldn't be necessary given the call to `dev.wait-for` below it.

docs/manual_upgrades.rst
```diff
@@ -12,20 +12,17 @@ Please add new instructions to the top, include a date, and make a post in the `
...
 
     make dev.pull
 
-4. Run provisioning command ::
+3. Run provisioning command ::
 
     make dev.provision
 
-5. [Optional] Additionally, there is a database copy command to help you transfer data from MySQL 5.7 to 8.0. After provisioning use the following command ::
+4. [Optional] Additionally, there is a database copy command to help you transfer data from MySQL 5.7 to 8.0. After provisioning use the ``dev.dbcopyall8`` command. This command will stop all of your services. Will clean your ``mysql80`` container and copy all of your databases from ``mysql57`` to ``mysql80``. ::
```
timmc-edx: [nit] Grammar
```suggestion
4. [Optional] Additionally, there is a database copy command to help you transfer data from MySQL 5.7 to 8.0. After provisioning use the ``dev.dbcopyall8`` command. This command will stop all of your services, clean your ``mysql80`` container, and copy all of your databases from ``mysql57`` to ``mysql80``. ::
```

Makefile
```diff
@@ -464,10 +464,16 @@ dev.dbshell:
...
 	$(MAKE) dev.wait-for.mysql57+mysql80
+	docker compose exec -T mysql80 mysql -uroot mysql < provision-mysql80.sql
 	$(MAKE) $(_db_copy8_targets)
```
timmc-edx, line 475: The copying is now failing for me, with the following error:

```
docker compose exec mysql57 mysqldump "credentials" > .dev/credentials.sql
docker compose exec -T mysql80 mysql "credentials" < .dev/credentials.sql
ERROR 2013 (HY000) at line 621: Lost connection to MySQL server during query
make[1]: *** [Makefile:481: dev.dbcopy8.credentials] Error 1
```

If I dump the logs just before this statement runs, I get a line about "Temporary server started." This temporary server is started and then stopped before the regular mysql server runs. I don't see this in the mysql57 logs. This might mean that mysql80 needs a different availability check.

timmc-edx: If I move the `sleep 10` to below the wait-for line, it works. Something is still very strange here, because I can run various checks against MySQL immediately between the provision-mysql80.sql and _db_copy8_targets lines and they always pass, even when run repeatedly in short succession.

timmc-edx (approved):

Makefile
timmc-edx: OK, I think I've got it! The problem is that the mysql command is talking to the temporary server over a UNIX socket. The solution seems to be to tell mysql to use TCP. I'll submit a separate PR for that, though.

mumarkhan999:
Closing this to re-run `openedx/cla`. It's stuck for an unknown reason.

PR #1219, opened 2023-10-31
chore: Added tox<4.0.0 to constraints.txt

thezaeemaanwar:
https://github.com/openedx/public-engineering/issues/217


feanil (approved):

feanil:
@thezaeemaanwar you should also run make upgrade so that we can confirm that the new constraint is being read correctly

PR #1223, opened 2023-11-02
build: Delete add-to-project workflow

timmc-edx:
This never worked (we would have needed a different secret, from 2U) and we managed to get it working with a workflow we could set up in the GH UI instead.

----


rgraber (approved):
Red diff!

PR #1252, opened 2024-01-09
fix: remove periods from project name when release version is present

rgraber:
See https://github.com/openedx/devstack/issues/1248
----

### DIFF
--- a/options.mk
+++ b/options.mk
@@ -30,7 +30,7 @@ DEVSTACK_WORKSPACE ?= $(shell pwd)/..
 # so changing it will give you a separate set of databases.
 # See https://docs.docker.com/compose/reference/envvars/#compose_project_name
 # If OPENEDX_RELAESE is defined, defaults to `devstack-${OPENEDX_RELEASE}`;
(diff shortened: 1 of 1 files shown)


options.mk
```diff
@@ -30,10 +30,10 @@ DEVSTACK_WORKSPACE ?= $(shell pwd)/..
 # so changing it will give you a separate set of databases.
 # See https://docs.docker.com/compose/reference/envvars/#compose_project_name
 # If OPENEDX_RELAESE is defined, defaults to `devstack-${OPENEDX_RELEASE}`;
-# otherwise, it defaults to `devstack`.
+# otherwise, it defaults to `devstack`. Any periods will be replaced with hyphens to comply with docker project naming rules.
```
robrap: Nit:
1. Consider providing an example with a real openedx release so it stands out more.
2. Do we have docs around using devstack for openedx releases?

robrap (approved):

options.mk
rgraber: 1. Will fix
2. Yes. I am going to make a ticket to remove them, just haven't done that yet

rgraber: 2 is now https://github.com/openedx/devstack/issues/1253

PR #1272, opened 2024-02-07
fix: Update for new version of demo course

timmc-edx:
In https://github.com/openedx/openedx-demo-course/pull/47 the demo course was replaced. This includes a new course key (new name and org) and the course content being moved a few directories deeper. (Ticket: https://github.com/openedx/devstack/issues/1273)

Changes:

- `sed -i 's/course-v1:edX+DemoX+Demo_Course/course-v1:Axim+DemoX+demo_course/g'` on all files turned up by `grep -rl Demo_Course`
- Updated a few other course info locations turned up with `grep -nrI DemoX | grep -x demo_course` (all in programs/discovery.py)
- Updated git clone URL and checkout location in provision-lms.sh to use new repo name (not strictly needed).

----


timmc-edx:
Only failing for `ecommerce+lms`. There are still conflicting demo course references somewhere. I had hoped to fix this by changing mentions of `edX` as a partner in provision-ecommerce.sh along with changes in https://github.com/openedx/ecommerce/compare/master...timmc-edx:ecommerce:timmc/new-demo-course but no luck -- there are now duplicate partner code entries. Maybe requires reverting the SQL changes in this PR and re-running `update-dbs-init-sql-scripts.sh` with just the right changes to partner codes in devstack and ecommerce?

PR #1283, opened 2024-02-27
fix: update .readthedocs.yaml

robrap:
Attempt to fix the build. It is unclear why this is failing, so simply attempting to match the following as closely as possible:
https://docs.readthedocs.io/en/stable/config-file/v2.html

The failure no longer has an error (which used to be around missing os), and the raw output is missing the call:
> cat .readthedocs.yaml

It's unclear what is breaking. I'm attempting to clean-up the redundant python version reference to see what happens.

**Additional Context:**

* Here is where it went from passing to failing: https://readthedocs.org/projects/open-edx-devstack/builds/?page=3
* Last passing build details: https://readthedocs.org/projects/open-edx-devstack/builds/22194482/
  * View raw output for additional details.
* Latest failing build details: https://readthedocs.org/projects/open-edx-devstack/builds/23559660/
  * Raw output doesn't seem to provide much. It just dies.
* There was a rename from `.readthedocs.yml` to `readthedocs.yaml` not long before the failures, but the build was passing after this rename.

----

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -15,6 +15,5 @@ sphinx:
   configuration: docs/conf.py
 
 python:
-  version: 3.8
   install:
     - requirements: requirements/doc.txt


.github/workflows/provisioning-tests.yml
```diff
@@ -7,9 +7,11 @@ on:
     branches: [master]
     paths-ignore:
       - '**.rst'
+      - '**.readthedocs.yaml'
```
robrap: Note: The following did not work:
```
- '.readthedocs.yaml'
```

.github/workflows/provisioning-tests.yml
```diff
@@ -7,9 +7,11 @@ on:
     branches: [master]
     paths-ignore:
       - '**.rst'
+      - '**/.readthedocs.yaml'
   pull_request:
     paths-ignore:
       - '**.rst'
+      - '**/.readthedocs.yaml'
```
robrap: **Note to reviewer:** I can't get this to work. Provisioning tests still ran. Any thoughts?

timmc-edx: I wonder if it would work if you removed the `**/`. The docs say that `**` [matches any character](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#patterns-to-match-file-paths), which would require `/.readthedocs.yaml` to be a match. But there's no leading slash in the target path you're trying to match, right? So unless they're quietly treating leading-slash and no-leading-slash as the same, that would prevent the match you want.

(On the other hand they give the example of `**/README.md` meaning "A README.md file anywhere in the repository." but I wonder if it's actually true.)

robrap: * Yes. I based it off that README comment.
* I had also tried `- '**.readthedocs.yaml'` and - `'.readthedocs.yaml'`
* I wasn't sure if it was using my changes, or just the changes on master? It didn't make sense why none of this worked.

timmc-edx: I _think_ it uses the version on the branch.

robrap: Agreed. That was my assumption as well, but I was just grasping at straws since I don't understand why none of this worked. Should I just back it out?

timmc-edx: Yeah. We can cancel unwanted workflows manually for now.

robrap: Just saw your comment in Slack as well. I dropped that commit and I cancelled that workflow. Any other blockers?

timmc-edx (approved):
I don't think this will fix the problem we're seeing on builds for master, but I don't think it will make anything worse, and maybe RTD was just having a transient problem.

robrap:
> I don't think this will fix the problem we're seeing on builds for master, but I don't think it will make anything worse, and maybe RTD was just having a transient problem.

I didn't think so either, but I wanted to rule this out, and it looks like it worked. I'll create a support ticket with readthedocs, but I'm not sure who gets those emails yet.

